PR #2, opened 2018-05-15
List sources explicitly in CMake

jrood-nrel:
- This brings Nalu-Wind closer to CMake best practices
- This allows for CMake to know when files have been modified so the configure will be run again if necessary
- This also allows for conditional omission of source files to get rid of warnings when TPL-specific source files don't have symbols

@sayerhs This is how I remember listing sources explicitly in CMake in previous projects, but there might be better ways to do this now. I tried to read about using `target_sources` but was not able to understand its advantages or how to use it. I'm considering this a WIP because I'm open to any improvements that can be made on it, but it's ready to review as is.


CMakeLists.txt
```diff
@@ -174,19 +172,25 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/NaluVersionInfo.h.in"
...
+add_subdirectory(unit_tests)
+get_property(NALU_SOURCES GLOBAL PROPERTY GlobalSourceList)
+get_property(NALU_HEADERS GLOBAL PROPERTY GlobalHeaderList)
+get_property(NALU_UNIT_SOURCES GLOBAL PROPERTY GlobalUnitSourceList)
+get_property(NALU_UNIT_HEADERS GLOBAL PROPERTY GlobalUnitHeaderList)
 
 include_directories(${CMAKE_SOURCE_DIR}/include)
 include_directories(${CMAKE_BINARY_DIR}/include)
-add_library(nalu ${SOURCE} ${HEADER})
+add_library(nalu ${NALU_SOURCES} ${NALU_HEADERS})
```
sayerhs: Headers aren't necessary in `add_library` rule.

cmake/AddSourcesFunctions.cmake
```diff
@@ -0,0 +1,43 @@
+function(add_sources)
+   foreach(_source IN ITEMS ${ARGN})
+       if (IS_ABSOLUTE "${_source}")
+           set(source_abs "${_source}")
```
sayerhs: Did you mean to say `_source_abs`? Otherwise a file with absolute path doesn't get appended to `GlobalSourceList`

include/CMakeLists.txt
```diff
@@ -0,0 +1,217 @@
+add_sources(
```
sayerhs: This should be `add_headers`?

include/element_promotion/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+add_sources(
```
sayerhs: `add_headers`?

src/CMakeLists.txt
```diff
@@ -0,0 +1,213 @@
...
+   AssembleScalarElemOpenSolverAlgorithm.C
+   AssembleScalarElemSolverAlgorithm.C
+   AssembleScalarFluxBCSolverAlgorithm.C
+   AssembleScalarNonConformalSolverAlgorithm.C
+   AssembleTemperatureNormalGradientBCSolverAlgorithm.C
+   AssembleWallHeatTransferAlgorithmDriver.C
+   AuxFunctionAlgorithm.C
+   AveragingInfo.C
+   BoundaryConditions.C
+   CMakeLists.txt
```
sayerhs: Remove `CMakeLists.txt`

include/overset/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+add_sources(
+   AssembleOversetSolverConstraintAlgorithm.h
+   CMakeLists.txt
```
sayerhs: Remove `CMakeLists.txt`

sayerhs:
`target_sources` will eliminate the custom functions `add_sources` and `add_utest_sources` you have defined currently. You can do 

```
#########################################
# nalu-wind/CMakeLists.txt
add_library(nalu)

include(src/CMakeLists.txt)

#########################################
# nalu-wind/src/CMakeLists.txt
target_sources(nalu 
    Realm.C
    # Other sources
)

#########################################
# unit_tests/CMakeLists.txt
target_sources(${utest_ex_name}
    UnitTestBasicKokkos.C
)
```

This would be preferable since it is a builtin CMake function and not something custom that we have to maintain.

jrood-nrel:
Thanks @sayerhs, I hopefully fixed all my copy and paste mistakes and tested the build process again. Next I will try to implement `target_sources`.

sayerhs:
@jrood-nrel One downside with `target_sources` is that you'll have to explicitly prefix `${CMAKE_CURRENT_LIST_DIR}` to every file if it is not in the same directory as the `add_library` definition. Your current solution might be cleaner that way.

jrood-nrel:
I see. Clearly not difficult, but requires more text. Might be a wash if we want to have the list of headers and will still need one or two of the current functions around for that anyway.

One thing I feel like I should be able to do is template the 4 functions into one that will still handle all 4 source lists.

Another thing I was wondering was if it makes sense to move `nalu.C` and `unit_tests.C` under their respective directories as well here?

sayerhs:
So I would change the functions to be something like this (where it accepts a target and populates the appropriate property that way. 

```
function(add_sources target)
   foreach(_source IN ITEMS ${ARGN})
       if (IS_ABSOLUTE "${_source}")
           set(_source_abs "${_source}")
       else()
           get_filename_component(_source_abs "${_source}" ABSOLUTE)
       endif()
       set_property(GLOBAL APPEND PROPERTY "${target}" "${_source_abs}")
   endforeach()
endfunction(add_sources)
```

Any reason why we declare a `property` instead of just doing `list(APPEND ${NALU_SOURCES} "${_source_abs}")`?

That way there can be one function that handles headers, sources, for different targets.

jrood-nrel:
So I just got this function from stackoverflow and the CMake mailing list. I think since we have nested directories, we need a global property for this to work though? As in the list would just have local scope in that case? I think I agree with the single function with an extra argument though.

PR #6, opened 2018-06-01
ABL Statistics post-processing utility

sayerhs:
Implement a boundary layer post-processing utility to generate spatially
and temporally averaged quantities of interest:

  - Velocity
  - Resolved stress
  - SFS stress
  - Temperature
  - Temperature flux
  - Temperature variance
  - uTau

This utility builds on top of TurbulenceAveragingPostProcessing utility, the
temporal averages are computed in that class.

PR #7, opened 2018-06-01
Optimize ABLForcingAlgorithm for precursor simulations

sayerhs:
- Move ABLForcingAlgorithm to `wind_energy` directory

- Convert ABLForcingAlgorithm to use the ABL Statistics object that already
  contains the spatially averaged velocities and temperatures as a function of
  height to determine driving pressure gradient

- Update ablNeutralEdge regression test to use the new ABL forcing class.
  Change in order of operations caused diffs of O(1e-10), updated gold files
  accepting diffs. Test runtime changed from 56s to 33s (on NREL Peregrine
  system)

- Tests on the OWEZ (9x9x1 km) precursor run indicates that the wall-clock time
  per Nalu timestep reduced to approximately 4s/time-step from 9s/time-step with
  the original implementation.


sayerhs:
This pull request is dependent on #6 and should not be processed before that.

tonyinme:
This looks good.
I approve.

jrood-nrel:
@sayerhs Can you resolve the conflicts here?

PR #8, opened 2018-06-06
Beginning to compile nalu for Nvidia/cuda.

alanw0:
Need to change a few things to make explicit that current kernels
which involve non-gpu stk-mesh etc., can only run on a 'host'
execution space, not a 'device' execution space. Cuda is the only
current platform for which the 'host' and 'device' are actually
different.

As we add actual gpu unit-tests and code changes, we'll use 'device'
variants of team-policy, etc., which will still run fine on open-mp and
other configurations.

TpetraLinearSystem.C still doesn't build for cuda yet, but this commit 
gets most of the code building.


alanw0:
@sayerhs thanks for approving. I was going to merge but noticed there is also the 'Fix failing ablNeutralEdge' commit here now... I'm not sure what that means. Should I still go ahead and process the pull-request? Would rebase-and-merge be the right thing to do?

sayerhs:
@alanw0 Yes, rebase and merge is the right thing to do here.

PR #13, opened 2018-06-15
Add a small class to create a class instance on device.

overfelt:
And a unit test for it.  It is not just creating a device
object but a virtual class instance on device and then
invoking a virtual function through the base class on
device.

This is meant to be a simple example and the start of the
discussion on how to structure the code for future
development of GPU enabled Kokkos kernels.

### DIFF
--- a/include/utils/CreateDeviceExpression.h
+++ b/include/utils/CreateDeviceExpression.h
@@ -1,0 +1,27 @@
+#ifndef CREATEDEVICEEXPRESSION_H
+#define CREATEDEVICEEXPRESSION_H
+
+#include <type_traits>
+
+#include <KokkosInterface.h>
+
+namespace sierra {
+namespace nalu {
+
+template <typename T>
+inline 
+T* create_device_expression(const T & rhs)
+{
+  const std::string debuggingName(typeid(T).name());
+  T* t = kokkos_malloc_on_device<T>(debuggingName);
+  // Bring rhs into local scope for capture to device.
+  const T RHS(rhs);
+  kokkos_parallel_for(debuggingName, 1, [&] (const int i) {
+    new (t) T(RHS); 
+  });
+  return t;
+}
+} // namespace nalu
+} // namespace sierra
+
+#endif /* CREATEDEVICEEXPRESSION_H */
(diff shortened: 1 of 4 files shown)


include/utils/CreateDeviceExpression.h
```diff
@@ -0,0 +1,27 @@
...
+
+template <typename T>
+inline 
+T* create_device_expression(const T & rhs)
+{
+  const std::string debuggingName(typeid(T).name());
+  T* t = kokkos_malloc_on_device<T>(debuggingName);
+  // Bring rhs into local scope for capture to device.
+  const T RHS(rhs);
+  kokkos_parallel_for(debuggingName, 1, [&] (const int i) {
```
alanw0: if rhs was passed in by value instead of reference, then would the local RHS be un-necessary?

alanw0: Overall looks great, I like the unit test.

overfelt: Yes, passing by value would remove the need for a local RHS and I agree it would make the code a bit cleaner.

overfelt: Wait, darn, should have read my own comment.  I don't think the cuda compiler will correctly bring a passed parameter into local scope and mirror it to the device.  It should but we have had trouble with this.  We could pass by value and then set a reference to it to bring it into local scope.  Could also be that newer versions of the compiler fix this.

alanw0: That's surprising. It seems like passing something by value through the argument list should be the same as constructing the local from a reference. Both necessarily involve copying the class data from the host to the device. In any case, this isn't a big deal. Perhaps we could query Christian Trott just to hear his insight on this issue.

unit_tests/UnitTestCreateOnDevice.C
```diff
@@ -0,0 +1,91 @@
...
+#include <string>
+#include <iostream>
+#include <vector>
+
+#include <KokkosInterface.h>
+#include "utils/CreateDeviceExpression.h"
+
+class deviceable {
+protected :
+  deviceable *DeviceCopy;
```
CamelliaDPG: Apologies if this is picking nits, but shouldn't class names start with uppercase letters, and member variables start with lowercase letters, and end in an underscore?

Overall, though, I quite agree with @alanw0 — this looks great.  Thanks for doing this, @overfelt!

PR #15, opened 2018-07-02
Add compile guards in EigenDecomposition for non-SIMD architectures

sayerhs:
On non-SIMD architectures, stk::simd::DoubleType is the same as double.
This causes compilation errors because of redefinition of functions.

### DIFF
--- a/include/EigenDecomposition.h
+++ b/include/EigenDecomposition.h
@@ -24,6 +24,7 @@ namespace EigenDecomposition
   void matrix_matrix_multiply(const double (&D)[2][2], const double (&Q)[2][2], double (&A)[2][2]);
   void matrix_matrix_multiply(const double (&D)[3][3], const double (&Q)[3][3], double (&A)[3][3]);
 
+#ifndef STK_HAVE_NO_SIMD
   // Simd solvers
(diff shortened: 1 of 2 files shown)


sayerhs:
@jamelvin I don't think the lines 146-288 need the guards, because `stk::simd::DoubleType` will just be `double`. The STK functions should still work.

jamelvin:
Will these functions still resolve correctly?

for (unsigned is = 0; is < stk::simd::ndoubles; is++) {
        EXPECT_NEAR(stk::simd::get_data(b_[i][j], is), stk::simd::get_data(A2d_rand_simd[i][j], is), tol);
}

Also, this kind of data access, which I used for initialization or the matrices in the tests? 

A2d_rand_simd[0][0]._data[j] = a11*(j + 1);

jrood-nrel:
I can try this PR out when Mira is back.

alanw0:
Looks good. It would also work to just delete the 'double' versions. Unless you actually need both when simd is on.

jamelvin:
@sayerhs @alanw0 I think we still need both at this point as the only live function that uses it, AssembleScalarEigenEdgeSolverAlgorithm has not been updated to use SIMD datatypes yet.  Also, we may want to leave the double form in there until we have finished adding our new hybrid models, as its not clear exactly where those will need to be used as part of them yet.

PR #16, opened 2018-07-03
Replace deprecated MPIEXEC with MPIEXEC_EXECUTABLE

sayerhs:
See https://cmake.org/cmake/help/latest/module/FindMPI.html for
deprecation details.


jrood-nrel:
I'm fine with this. I just can't find what version of CMake deprecated it. Do you know?

sayerhs:
@jrood-nrel Ah, good point. Looking back at versions it seems like a recent change v3.10.x and above. So I'll close this for now.

PR #19, opened 2018-07-10
Add tensor product kernels

rcknaus:
* To activate the option, add the line `tensor_product_cvfem: yes` to the solution options block.  You can't mix and match between the old style and the new style at least for now.  Only hex27 and promoted hex8 meshes are usable.

* Heat conduction, momentum, pressure, and scalar transport kernels implemented.

* Boundary conditions still use the old formulation.  There hasn't seemed to be an issue with this, but it's something to note.

* There are some dropped terms in the jacobian for momentum with variable viscosity.  Since this is an intermediate stage to using a Jacobian-free solver and it doesn't seem to be an issue for the couple test cases I ran, I'm keeping the lhs assembly as is for this pull request.

 * Added a consolidated option for the projected nodal gradient eqsys.  `use_consolidated_png: yes` in solution options.  Only the new type of kernel is supported  right now.

Some order-of timing data for a serial taylor green vortex breakdown (100 steps) test case with a fixed 32^3 nodes, on an Intel E5-2650 v3 (Haswell). Times are those reported by the timer overview in the log file

Assembly times (in seconds):

| Case  | Momentum | Continuity | PNG
| ------------- | ------------- | ------------- | ------------- |
| P1 baseline |  54 | 21 | NA
| P2, previous | 206 | 64 | 99
| P4, previous | 3018 | 1582 | 841
| P2, new | 45 | 9 | 36
| P4, new | 113 | 18 | 96   

The tallpoles for assembly timing are in forming the jacobians and summing into the global left-hand side.


rcknaus:
The new assembleelemsolverho etc are a bit different in that there's now a mapping into a locally structured array and back again before summing into the linear system.   That and they lack a master element.  The functions/classes didn't really need to be duplicated but combining them ended up complicating things a good amount and would cause conflicts.

Anyway, sorry about the pull request size.

spdomin:
Overall, great speed-ups. Lots of code that I am working through. It may be into next week and beyond before I get to the final review. Feel free to proceed as the review team suggests.

alanw0:
I am also free next week.

sayerhs:
@rcknaus @spdomin There as a _formal write-up_ of this approach? Can we please see that?

PR #21, opened 2018-07-19
Created ElemDataRequestsNGP class.

alanw0:
The purpose of this class is to copy the contents of
ElemDataRequests to a GPU device for use by ScratchViews etc.

Data-enums and fields are stored in kokkos views instead of
std sets and maps.

Unit test in UnitTestElemDataRequests.C shows how this class
could be used to copy the contents of a regular ElemDataRequests
class to a gpu device for access within a parallel_for loop.

This class is not yet used by ScratchViews and SharedMemData, but
will be soon as they are next up for GPU overhaul.

Still need to address the master-element pointers, which need
to be allocated on device in order to be used on device...

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -96,6 +96,7 @@ add_sources(GlobalHeaderList
    EffectiveSSTDiffFluxCoeffAlgorithm.h
    EigenDecomposition.h
    ElemDataRequests.h
+   ElemDataRequestsNGP.h
    EnthalpyABLSrcNodeSuppAlg.h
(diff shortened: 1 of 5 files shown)


CamelliaDPG:
This looks good to me.  I'm glad to see the unit test.  Are we at a point where the unit test can actually run on CUDA, at least with the upcoming updates to stk?  Have you tried that, @alanw0?

alanw0:
@CamelliaDPG haven't gotten a bigger unit test to run yet, as scratchviews still need to be updated. To finish updating scratchviews, we will need the kokkos fix for alignment of shared-mem-views.

PR #29, opened 2018-08-03
Extrapolate pressure gradient to opposing face for zero grad option

rcknaus:
* Instead of working on an assembled quantity, just use the opposing dual cell face value.  This will avoid the need to deal with corners.

* Tested it on the full OWEZ case for 530s of simulation time


src/AssembleNodalGradPOpenBoundaryAlgorithm.C
```diff
@@ -0,0 +1,178 @@
...
+        // evaluate pressure at opposing face.  If zeroGrad_ option is used, then
+        // we'll copy this value to the exposed face.  Otherwise, it's unused
+        double pOpp = 0.0;
+        const int opp_offset = meSCS.opposingFace(faceOrdinal, ip) * nodesPerElement;
+        const double* oppShapeFns = &p_elem_shape_function[opp_offset];
+        for (int ic = 0; ic < nodesPerElement; ++ic) {
+          pOpp += oppShapeFns[ic] * p_elemPressure[ic];
+        }
+
+        double press_div_vol = (zeroGrad_ ? pOpp : pIp) / volNN;
```
sayerhs: @rcknaus Can you explain reason to evaluate both and use a ternary to select the appropriate value? Is this in preparation for NGP?

rcknaus: Not really.  It's more following the style of masking out these options, like we do for NOC or divU etc

spdomin:
We actually enforce zero gradient on a number of boundary faces (for the dofs). Although this formulation is only valid for a zero-pressure gradient flow, it would be beneficial for all open bcs for all dofs. We may like to generalize this. I think that we talked about this when we realized that what we do for projected nodal gradients might be inconsistent wrt to the diffusive fluxes.

sayerhs:
@rcknaus Are you planning on generalizing this as per @spdomin's suggestion? It would be great if we could merge this feature for use in the ABL+ALM simulations.

PR #35, opened 2018-09-13
Implement Momentum UVW system solver with HYPRE

sayerhs:
Use HYPRE to solve both coupled UVW and segregated UVW systems using HYPRE.

Other changes/enhancements to HYPRE include

- Implement the feature to allow different linear solver tolerances for
  intermediate and final outer iterations of the linear system.

- Add CO-GMRES solver if available in the HYPRE version.

- Implement `reuse_preconditioner` option for HYPRE also. Previously, this
  option was not used in HYPRE solvers.

- Update regression test to use UVW momentum solver and update gold files


sthomas61:
I have been testing the segregated extensively on the V27 mesh without any encountering any issues . The memory leak previously reported has been fixed.

PR #43, opened 2018-10-02
Unit test of ngp::Mesh, demonstrating the most basic usage.

alanw0:
This commit also has several fixes for cuda compile errors, but
to see the unit test just read file UnitTestNgpMesh1.C.

### DIFF
--- a/unit_tests/UnitTestShmemAlignment.C
+++ b/unit_tests/UnitTestShmemAlignment.C
@@ -8,7 +8,7 @@ bool is_aligned(const void *pointer, size_t byte_count)
     return (uintptr_t)pointer % byte_count == 0;
 }
 
-TEST(Shmem, align)
+void do_the_test()
 {
   unsigned N = 1;
   unsigned numScalars = 3;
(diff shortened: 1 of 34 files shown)


unit_tests/UnitTestShmemAlignment.C
```diff
@@ -39,3 +39,8 @@ TEST(Shmem, align)
   EXPECT_EQ(3u, numCorrectlyAlignedViews);
 }
 
+TEST(Shmem, align)
```
sayerhs: @alanw0 What is the reason for this extra level of indirection?

alanw0: @sayerhs By indirection do you mean pulling the body of the test out into a separate function? cuda complained that a device lambda can't be launched from within a class method that is private or protected. gtest TESTs are protected members of a class. So pulling the body of the test into a free-standing function allowed it to compile.

alanw0:
Guys, I added a second commit to this which makes the rest of the unit-tests at least compile for CUDA, and the one in UnitTestNgpMesh1.C actually runs and works correctly on GPU now.

alanw0:
ok guys, I just pushed one last commit to the pull request, I won't do any more until this is approved and in. This last commit got the test in UnitTestShmemAlignment.C working on GPU. There is a weirdness with the result though, as you can see at the bottom of the 'do_the_test' function... I may consult with some kokkos/gpu experts about that, but I think this pull-request is at a good stopping point now.

unit_tests/UnitTestShmemAlignment.C
```diff
@@ -8,34 +8,54 @@ bool is_aligned(const void *pointer, size_t byte_count)
...
-  }, numCorrectlyAlignedViews);
+  });
+
+  Kokkos::deep_copy(hostResults, ngpResults);
+
+#ifdef KOKKOS_HAVE_CUDA
+//We're expecting the result to be 3.
+//On GPU the result is 96... Is it 96 because there are 32 threads
+//per warp and you never have less than 1 warp ???
+//Not sure I understand or like this...
```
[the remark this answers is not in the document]
alanw0: Thanks @overfelt I'll try to adjust the test. Generally I think Kokkos::AUTO is the way to go but for a special-case unit-test like this, it probably makes more sense to just set the team size to 1.

PR #45, opened 2018-10-05
Refactor Actuator Classes 

psakievich:
I've pulled the redundant code between ActuatorLinePointDrag and ActuatorLineFAST to simplify the interface as a precursor step to creating an actuator disk.

### DIFF
--- a/include/Actuator.h
+++ b/include/Actuator.h
@@ -87,7 +87,8 @@ public:
     const YAML::Node &node):realm_(realm),
     searchMethod_(stk::search::KDTREE),
     needToGhostCount_(0){}
-  virtual ~Actuator() {}
+
+  virtual ~Actuator();
 
   // load all of the options
   virtual void load(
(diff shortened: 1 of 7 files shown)


include/Actuator.h
```diff
@@ -66,6 +99,105 @@ class Actuator
...
+
+  // save off product of search
+  std::vector<std::pair<theKey, theKey> > searchKeyPair_;
+  // bounding box data types for stk_search */
+  std::vector<boundingSphere> boundingSphereVec_;
+  std::vector<boundingElementBox> boundingElementBoxVec_;
+  // target names for set of bounding boxes
+  std::vector<std::string> searchTargetNames_;
+
+  std::vector<ActuatorInfo*> actuatorInfo_;
```
sayerhs: Would there be an issue using `std::unique_ptr` here? That will eliminate the code in destructor.

include/Actuator.h
```diff
@@ -66,6 +99,105 @@ class Actuator
...
+  std::map<size_t, ActuatorPointInfo*> actuatorPointInfoMap_;
```
sayerhs: Also `std::unique_ptr` here.

include/Actuator.h
psakievich: Yes definitely.  I noticed this yesterday when I started working on the actuator disk.  I will make the change and push an update.

PR #46, opened 2018-10-10
Adding new unit test for stk ngp mesh fields

jvo1012:


unit_tests/UnitTestNgpMesh1.C
```diff
@@ -67,6 +67,84 @@ TEST_F(Hex8MeshWithNSOFields, ngpMesh1)
...
+  const double flowRate = 4.0;
+
+  Kokkos::parallel_for(team_exec, KOKKOS_LAMBDA(const sierra::nalu::DeviceTeamHandleType& team)
+  {
+    const ngp::Mesh::BucketType& b = ngpMesh.get_bucket(stk::topology::ELEM_RANK, team.league_rank());
+
+    const size_t bucketLen   = b.size();
+    const size_t simdBucketLen = sierra::nalu::get_num_simd_groups(bucketLen);
+ 
+    Kokkos::parallel_for(Kokkos::TeamThreadRange(team, simdBucketLen), [&](const size_t& bktIndex)
```
sayerhs: So a question: when to use `KOKKOS_LAMBDA` (as you did for team `parallel_for`) vs. `[&](...)`?

alanw0: My understanding is that it's good to use KOKKOS_LAMBDA for the outer parallel_for. KOKKOS_LAMBDA captures using '=' which makes the captured things const. (Because things from the host can't be modified in the device kernel.) But the inner loop can use non-const capture because it's being launched from already on device. Maybe that breaks down in some cases but that's my understanding at this point.

sayerhs: @alanw0 I didn't realize there was a `const ([=])` or `non-const ([&])` implication to these two. I thought `[=]` implied a copy (pass by value?) instead of pass by reference. Did you mean the call site essentially looks like a const because it is pass by value?

alanw0: @sayerhs I think you are precisely correct. I was thinking about constness with respect to the calling scope, but to be precise the difference is pass by value vs reference.

PR #48, opened 2018-10-12
Add line of site sampling along a Spinner Lidar rosette pattern

rcknaus:
* Added an exodus output option to data probes.  It doesn't seem like it scales well with a large number of data probes from some local tests


include/wind_energy/SyntheticLidar.h
```diff
@@ -0,0 +1,95 @@
...
+};
+
+class LidarLineOfSite
+{
+public:
+  LidarLineOfSite() = default;
+  std::unique_ptr<DataProbeSpecInfo> determine_line_of_site_info(const YAML::Node& node);
+private:
+  void load(const YAML::Node& node);
+  void write_netcdf();
```
sayerhs: This method is not needed?

include/wind_energy/SyntheticLidar.h
```diff
@@ -0,0 +1,95 @@
...
+  void setup_transfer();
```
sayerhs: Transfers are handled by DataProbePostProcessing, so this is not required?

include/wind_energy/SyntheticLidar.h
```diff
@@ -0,0 +1,95 @@
...
+  PrismParameters innerPrism_;
+  PrismParameters outerPrism_;
+  double beamLength_{1.0};
+
+  std::array<double, 3> lidarCenter_{{0,0,0}};
+  std::array<double, 3> laserAxis_{{1,0,0}};
+  std::array<double, 3> groundNormal_{{0,0,1}};
+};
+
+class LidarLineOfSite
```
sayerhs: Misspelling. Did you mean "line of sight"?

src/user_functions/GaussianWakeVelocityAuxFunction.C
```diff
@@ -0,0 +1,76 @@
+/*------------------------------------------------------------------------*/
```
sayerhs: I don't see the usage of this class anywhere. Is a commit missing, or is this code no longer needed?

include/wind_energy/SyntheticLidar.h
rcknaus: No.  Both spellings are valid and the existing capability is called "line of site" in Nalu currently.

PR #51, opened 2018-10-23
Bug fix in Courant calculation.

jamelvin:
Minor bug fix in the AssembleCourant where it was intending to average over the left and right node but used the left node twice.  

It is possible this will cause a minor change in timesteps and thus results in the tests.  Should I be running it through the tests first?


jamelvin:
Hey @sayerhs, where should I run the tests with GCC?  I assume I need to use GCC 6.4 and the correct stack specified here: https://nalu-wind.readthedocs.io/en/latest/source/developer/testing.html.  My local machine times out for a lot of the tests (I only have 12 processors available)...  Should I try building this on Cori?  Or is this already setup on Peregrine?  Thanks!

PR #58, opened 2018-11-29
Updating the MPI execution configuration used during testing

jrood-nrel:
This adds an `--oversubscribe` flag to `mpiexec` when CMake sees it's on a Mac. I can't figure out a good way to know whether or not the MPI implementation isn't OpenMPI to not use the flag, which means if someone was using MPICH on a Mac, it would probably fail on the flag being injected. Not sure what else to do though since this flag is now required when testing with a new version of OpenMPI on a Mac.

Next, since Trilinos forced a minimum CMake 3.10, we can now complete the recommended MPI execution model by removing the deprecated `MPIEXEC`.

### DIFF
--- a/reg_tests/CMakeLists.txt
+++ b/reg_tests/CMakeLists.txt
@@ -51,5 +51,12 @@ execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink
                 ${CMAKE_CURRENT_SOURCE_DIR}/xml
                 ${CMAKE_CURRENT_BINARY_DIR}/xml)
 
+# Allow oversubscribing OpenMPI on a Mac during testing
+# TODO: Don't inject the oversubscribe flag if not OpenMPI, e.g. MPICH
+if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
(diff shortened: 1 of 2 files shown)


jrood-nrel:
Actually I just realized this isn't specific to the Mac, just that our laptops tend to not have enough cores, so maybe `--oversubscribe` needs to be turned on regardless of OS.

sayerhs:
I agree with deprecating the `MPIEXEC` which we discussed in #16. However, wouldn't it be better to pass 

```
        -DMPIEXEC_PREFLAGS:STRING='"--use-hwthread-cpus --oversubscribe"' 
```

argument to CMake command line, put that in the example config scripts and update the documentation? This way, the user can choose what they want to do hyperthreading etc.

sayerhs:
I wonder if it would be better to do 

```
-DMPIEXEC_PREFLAGS="--use-hwthread-cpus;--oversubscribe"
```

jrood-nrel:
Ah, yes, that sounds like the right solution. I will update this PR with an implementation of that.

jrood-nrel:
After looking through the scripts in `build` and the documentation, it appears many things are now out-of-date, so I think that will have to be overhauled at a later date. I'm just going to start by implementing the flags during testing.

PR #59, opened 2018-11-30
Remove build directory

jrood-nrel:
This removes the `build` directory and out-of-date build scripts therein. I also moved the (probably out-of-date) Trilinos configuration to the documentation.

The documentation is due for an overhaul, but I think it's time for the `build` directory to go.


docs/source/user/build_manually.rst
```diff
@@ -274,6 +274,105 @@ Prepare:
     git clone https://github.com/trilinos/Trilinos.git
     cd $nalu_build_dir/packages/Trilinos
     mkdir build
+    cd build
+
+Now create a ``do-configTrilnos`` script with the following recommended options:
```
sayerhs: typo `trilinos` is missing the second `i`

docs/source/user/build_manually.rst
```diff
@@ -274,6 +274,105 @@ Prepare:
...
+
+.. code-block:: bash
```
sayerhs: Recommend we change this to `.. literalinclude::` and then also provide a link for the user to download the file.

docs/source/user/build_manually.rst
```diff
@@ -412,6 +496,10 @@ top of the file to the root build directory path.
 Nalu-Wind
 ~~~~~~~~~
 
+.. note::
+
+     The scripts referred to in the following section are available in the Nalu-Wind git history but have since been removed.
+
```
sayerhs: Should we add a link to [do-config-nalu-wind-peregrine.sh](https://github.com/Exawind/build-test/blob/master/configs/machines/peregrine/do-config-nalu-wind-peregrine.sh) as a template that users should use until you clean up this section of docs?

PR #61, opened 2018-12-07
Added metric tensor functionality

marchdf:
This implements the metric tensor that will be used as part of the
UT-A hybrid models. EigenDecomposition was templated. Unit tests were
added for the metric tensor. All other unit tests pass.


src/ComputeMetricTensorElemAlgorithm.C
```diff
@@ -0,0 +1,144 @@
...
+  coordinates_ = meta_data.get_field<VectorFieldType>(
+      stk::topology::NODE_RANK, realm_.get_coordinates_name());
+  Mij_ = meta_data.get_field<GenericFieldType>(stk::topology::ELEMENT_RANK,
+                                               "metric_tensor");
+}
+
+//--------------------------------------------------------------------------
+//-------- execute ---------------------------------------------------------
+//--------------------------------------------------------------------------
+void ComputeMetricTensorElemAlgorithm::execute() {
```
rcknaus: Is it necessary to do any regularization/smoothing on mij for the tensor eddy viscosity model when the mesh has poor regularity?  This would apply to gij too, I guess.

marchdf:
@sayerhs and @rcknaus , thanks for the suggestions. I incorporated them as best I could:

Changes:
- added SIMD unit test
- removed Fortran code
- added a 2D generic Mij capability
- cleaned EigenDecomposition calls
- removed otype
- cleaned left over comments

src/ComputeMetricTensorElemAlgorithm.C
marchdf: That is unclear right now and one of the things that will need investigating when we do tests in Nalu. UT-A is not reporting that this is necessary so far.

PR #63, opened 2018-12-11
ScratchViews prototype for execution on GPUs

sayerhs:
This pull request implements a unit test that demonstrates the use of `ngp::Mesh`, `ngp::Field` in constructing and executing a TestKernel with ScratchViews on GPU. It introduces a new class `ElemDataRequestsGPU` to ease the transition of host-only kernels to NGP kernels. The `*GPU` classes will be merged with the original classes `ScratchViews` and `ElemDataRequests` once the transition of the core `Kernel` classes are complete.


sayerhs:
@jrood-nrel This pull request introduces two changes to `CMakeLists.txt`, 

- Introduce a new variable `ENABLE_CUDA` to track compilation on GPU. This is necessary to turn off SIMD when building with GPUs
- We are currently using `Trilinos_CXX_FLAGS` to ensure that Kokkos introduced flags are consistently used when building Nalu for GPUs. I am not happy with that hack and will be working to have a better solution for long-term, but for now this is the best way to ensure that everything compiles properly. 

Long term solution to CUDA would be to use `project(Nalu-Wind CXX CUDA C Fortran)` or 

```
if (ENABLE_CUDA)
  enable_language(CUDA)
endif()
```

but we are not there yet.

jrood-nrel:
Would it make sense to only use `Trilinos_CXX_COMPILER_FLAGS` if we enable CUDA for now?

sayerhs:
@alanw0 I opted to not squash to preserve authorship, but I can do a "squash and merge" if you and @overfelt are OK with it.

@jrood-nrel I've introduced a check to add Trilinos CXX flags only if we are targeting GPUs

alanw0:
@sayerhs that's a good consideration but in this case I'm in favor of squashing because the individual commits probably aren't good on their own.

overfelt:
I prefer to lose authorship and have it squashed instead.   I'm going through an example right now of merging the trunk into my branch where someone had a single push with lots of commits and I have to remove all the conflicts from each commit sequentially.  Would have been easier if they had squashed first.

jro

PR #67, opened 2018-12-18
Fix unused code warnings in ScratchViewsNGP

jrood-nrel:
@sayerhs This ok? I'm just trying to narrow down some of the errors and warnings currently on the dashboard.


alanw0:
Yes that's ok, we'll get that code turned back on as soon as we get some gpu testing of master elements.

PR #71, opened 2019-01-15
Added total-num-fields to ElemDataRequestsNGP.

alanw0:
This is a small pre-requisite step towards bringing the
NGP classes ScratchViewsNGP and ElemDataRequestsGPU into
the mainstream.

Also cleaned up a minor error in KokkosInterface.h related
to 4D and 5D scratch-views. Kokkos doesn't let us use views
of that dimension, but still seemed reasonable to correct
the error in our API.


alanw0:
@sayerhs Kokkos does allow 4D and 5D views in general, but it seems to fail to compile if I try to use shared-scratch views of that dimension. I was working on a unit-test for copy-and-interleave and ran into that. I'll reconfirm and let you know exactly what it says.

PR #79, opened 2019-01-21
another step towards integrating ScratchViews and ScratchViewsNGP.

alanw0:
They now both use the new MultiDimViews class.

Next step will be to make them both have the same template parameters.

### DIFF
--- a/include/MultiDimViews.h
+++ b/include/MultiDimViews.h
@@ -20,6 +20,19 @@ struct NumNeededViews {
   unsigned num4DViews;
 };
 
+KOKKOS_FUNCTION
+inline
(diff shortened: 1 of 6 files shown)


include/MultiDimViews.h
```diff
@@ -20,6 +20,19 @@ struct NumNeededViews {
...
+size_t adjust_up_to_alignment_boundary(size_t input, size_t alignment)
+{
+  size_t remainder = input % alignment;
+  if (remainder > 0) {
+    return input + alignment - remainder;
+  }
+  return input;
+}
+
+constexpr unsigned maxViewsPerDim = 25;
```
sayerhs: Can this be made a class member in `MultiDimViews`

PR #80, opened 2019-01-22
Jroverf/gpu master element unit test

overfelt:
Initial unit test for new approach to master elements on the GPU.  There are still four hex master element functions that should be tested by UnitTestHexMasterElementsNgp.C commented out.  These should be easy since the pattern is established.  

It is going to be an effort to convert all the master elements over to this approach where all data is const data on the derived class and the Element base class has no data.


overfelt:
I notice this is a bunch of commits and I would like to compress to just one.  Working on that now...

jro

sayerhs:
@overfelt We could squash and commit the entire pull request as one commit. No need to reopen a new pull request for that.

overfelt:
The squash had a bunch of conflicts but I thought I could apply a patch and not have github know.  Instead I got a new branch with the same name....

PR #81, opened 2019-01-22
New unit test for new GPU enabled Hex8 master element.

overfelt:
This shows how to modify and use the master element
functions on a GPU. We would like to avoid use of
std::vector and virtual function calls, the master
element class needs to be changed somewhat.

### DIFF
--- a/include/AlgTraits.h
+++ b/include/AlgTraits.h
@@ -14,6 +14,9 @@
 namespace sierra {
 namespace nalu {
 
+class HexSCS;
+class HexSCV;
+
 // limited supported now (P=1 3D elements)
 struct AlgTraitsHex8 {
   static constexpr int nDim_ = 3;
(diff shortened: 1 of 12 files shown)


overfelt:
OK, I messed up the squash and ended up with a new branch of the same name.  Same changes though....

include/AlgTraits.h
```diff
@@ -22,6 +25,8 @@ struct AlgTraitsHex8 {
   static constexpr int numScvIp_ = 8;
   static constexpr int numGp_ = 8; // for FEM
   static constexpr stk::topology::topology_t topo_ = stk::topology::HEX_8;
+  using            masterElementScs_ = HexSCS;
+  using            masterElementScv_ = HexSCV;
```
sayerhs: For types we have been using a a captialized name in Nalu-Wind, should we adopt that approach here also? And should we drop the trailing underscore as this is a typedef and not a member of the struct? Would like to hear what would be the best practice for this.

overfelt:
So the suggestion is:

   using            MasterElementScs = HexSCS;
   using            MasterElementScv = HexSCV;


jro

PR #82, opened 2019-01-22
Extract ghosting utilities from NonConformalManager

sayerhs:
The ghosting add/remove logic does not depend on NonConformalManager
class and could be utilized in other places where ghosting changes as a
function of timestep (e.g., overset, actuator line modeling, etc.). This
commit extracts the common pieces and consolidates those functions as
part of StkHelpers so that they can be used across different parts of
Nalu-Wind.


sayerhs:
@alanw0 I would like to replace `elems` in the functions `keep_elems_*` and `keep_only_elems*` to `entities`, as I feel it could be also applicable for other entities. Would you agree?

src/utils/StkHelpers.C
```diff
@@ -0,0 +1,151 @@
...
+            entityKeys.push_back(key);
+          }
+        }
+      }
+    }
+  }
+}
+
+void
+keep_elems_not_already_ghosted(
```
sayerhs: @alanw0 this function seems like it doesn't do any downward relations

alanw0: Looks like you're right. The keep_only_elems function removes entities of lower rank, then the other functions figure out what needs to be added and removed, then the stk call to change_ghosting adds back in the lower-rank entities.

PR #86, opened 2019-01-24
NGP: Add ngp::Mesh instance to Realm

sayerhs:
Modify ngpMesh unittest to access the ngpMesh object from Realm.


alanw0:
Currently it should be a non-issue for CPU runs, should only be an issue when we actually run on GPU. Because for CPU runs, no cuda, ngp::Mesh is a thin adaptor around stk::mesh and no actual data copying is done.

sayerhs:
Yes, I agree this is not an issue with CPU runs, just wanted to make a note for future GPU performance studies.

PR #87, opened 2019-01-28
Clear the master element database to avoid memory leak.

overfelt:


jrood-nrel:
This doesn't seem to fix the memory leak on the unit test. I started looking at this Friday, but didn't get to finish it. I was looking at the `new` statements here assuming they need `delete`s: https://github.com/Exawind/nalu-wind/commit/bb4120660229f0ab2b730b801fd441bbcc4fb4b3#diff-0e94ce6b52348af9b2f663b9fa121ed8R32

overfelt:
The kokkos_free_on_device should have freed the corresponding kokkos_malloc_on_device. It is a bit tricky since these are stored in static arrays that will not be deleted unless explicitly asked for and making them unique_ptr() would not call the correct kokkos free memory function.

I'll try valgrind and see what it says....

PR #90, opened 2019-01-28
Jroverf/gpu master element unit test

overfelt:
The memory for the Hex 8 master element on the GPU is allocated
    through kokkos allocate memory and kokkos delete memory. But the
    actual class instance is formatted by a placement new executed
    on the GPU. So I think that it is necessary to call ~MasterElement()
    on the GPU to free the std::vectors that are in this base class and
    avoid the memory leak. Ran valgrind on the CPU and it is now clean.


jrood-nrel:
Thanks for helping to clear warnings too. If we get down to few enough, I will turn on `Wextra` for the tests and we can get closer to using `Werror` in the future.

PR #99, opened 2019-01-30
Fix bug in Hex8 element introduced last week.

overfelt:
The new hex8 still needs to push the const values
defined in this class to the non-const values that
are used in the base MasterElement class. In the
long run we would like for the base class to be
interface only with no data.

Alan & John,
Could this be the base cause of the bug you are seeing?  Maybe it was not Alan's fault.
jro


sayerhs:
@overfelt Should we prioritize and remove all data from the MasterElement class?

overfelt:
Shreyas,

I think you are the one who eventually sets the priorities and the order of the tasks.

There is one design decision to decide first.  When doing the Hex27 there is a Kokkos::View with certain data that I did not convert to a fixed array of POD.  Even though it is not POD it does seem to play well with the GPU and I think does not prevent you from creating an instance on the GPU and using it.  Using these Views in the derived class might make moving the std::vectors out of the MasterElement class easier.

jro

PR #102, opened 2019-01-31
Change formatting of output in pass_fail.sh

sayerhs:
Use 'printf' to provide a user-friendly output on max-diff for failing
tests. Also make the output more awk/sed/grep friendly.


sayerhs:
Example output

```
airfoilRANSElem                     PASSED:     3.7586 s
airfoilRANSEdge                     FAILED:     3.2964 s max-diff: 1.608900e-05
```

PR #105, opened 2019-02-04
Removing indirect realm dependency from mesh motion framework

ashesh2512:
Instead of passing the realm to the mesh motion framework, I now pass the bulk_data to remove any kind of dependency the mesh motion framework might have on the realm.


src/HeatCondEquationSystem.C
```diff
@@ -729,7 +729,7 @@ HeatCondEquationSystem::register_wall_bc(
     stk::mesh::put_field_on_mesh(*alphaField, *part, nullptr);
   
     // aux algs
-    AuxFunctionAlgorithm *alphaAuxAlg;
+    AuxFunctionAlgorithm *alphaAuxAlg = NULL;
```
gantech: I suppose this doesn't have anything to do with `mesh_motion`?

ashesh2512: no. this is a compiler warning fix from a previous commit. I actually forgot to rebase resulting in a merge commit which in turn resulted in a previous commit showing up in this pull request, although the actual file changes only correspond to the last commit in the pull request.

sayerhs:
Is it possible to rebase and have only one commit for your please?

PR #109, opened 2019-02-07
Update ghosting implementation for TIOGA overset connectivity

sayerhs:
Previous implementation of TIOGA overset ghost would destroy ghosting at each timestep, this commit adds necessary logic to determine the set of elements to be added and removed while retaining the already ghosted elements from previous domain connectivity step.

The new logic mirrors the setup found in `NonConformalManager` and utilizes the helper functions defined in `StkHelpers.h`

### DIFF
--- a/src/overset/TiogaSTKIface.C
+++ b/src/overset/TiogaSTKIface.C
@@ -19,7 +19,7 @@
 #include "master_element/MasterElement.h"
 #include "stk_util/parallel/ParallelReduce.hpp"
 #include "stk_mesh/base/FieldParallel.hpp"
-#include <stk_mesh/base/SkinBoundary.hpp>
+#include "stk_mesh/base/SkinBoundary.hpp"
 
 
 #include <iostream>
@@ -75,12 +75,14 @@ void TiogaSTKIface::setup(stk::mesh::PartVector& bcPartVec)
     tb->setup(bcPartVec);
   }
 
-  // Initialize the inactive part
-  oversetManager_.inActivePart_ = &meta_.declare_part(
-    inactivePartName_, stk::topology::ELEM_RANK);
+  if (populateInactivePart_) {
+    // Initialize the inactive part
+    oversetManager_.inActivePart_ = &meta_.declare_part(
+      inactivePartName_, stk::topology::ELEM_RANK);
 
-  oversetManager_.backgroundSurfacePart_ = &meta_.declare_part(
-    "nalu_overset_fringe_boundary", stk::topology::ELEM_RANK);
+    oversetManager_.backgroundSurfacePart_ = &meta_.declare_part(
+      "nalu_overset_fringe_boundary", stk::topology::ELEM_RANK);
+  }
 }
 
 void TiogaSTKIface::initialize()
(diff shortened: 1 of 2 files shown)


src/overset/TiogaSTKIface.C
```diff
@@ -195,24 +193,46 @@ void TiogaSTKIface::reset_data_structures()
...
     bulk_.modification_end();
 
-    sierra::nalu::populate_ghost_comm_procs(bulk_, *oversetManager_.oversetGhosting_, oversetManager_.ghostCommProcs_);
+    sierra::nalu::populate_ghost_comm_procs(
+      bulk_, *(oversetManager_.oversetGhosting_), oversetManager_.ghostCommProcs_);
 
 #if 1
     sierra::nalu::NaluEnv::self().naluOutputP0()
-      << "TIOGA: Overset algorithm will ghost " << g_ghostCount << " entities"
+      << "TIOGA: Overset algorithm will ghost " << global[0] << " elements"
```
alanw0: Might be nice to also print out how many recv-ghosts are being removed here. (global[1]).

sayerhs: @alanw0 I had that initially, but it is misleading. `global[0]` only has the count of elements that are being ghosted, but `global[1]` is the count of all entities including downward relations, which causes confusion (which was once reported by Lawson for his sliding mesh runs). For example, from one of my runs with overset after a few timesteps. 

```
TIOGA: Overset algorithm will ghost 797719 entities and remove 0 entities
TIOGA: Overset algorithm will ghost 466783 entities and remove 6029750 entities
TIOGA: Overset algorithm will ghost 500647 entities and remove 6149094 entities
TIOGA: Overset algorithm will ghost 547298 entities and remove 7041817 entities
TIOGA: Overset algorithm will ghost 595677 entities and remove 8131086 entities
TIOGA: Overset algorithm will ghost 643980 entities and remove 9301716 entities
TIOGA: Overset algorithm will ghost 689529 entities and remove 10415239 entities
TIOGA: Overset algorithm will ghost 731881 entities and remove 11483473 entities
```

sayerhs:
@alanw0 From my understanding, `inactivePart` was introduced to transfer entities from the actual mesh part to another part, it could have been named `holePart` or some other name. This change occurred before we replaced the `meta.universal_part()` in TpetraLinearSystem and this was the only way to eliminate nodes from the linear system. Now we use explicit part inclusion logic, so a lot of the `inactivepart` and `get_inactive_selector()` logic that was previously in Nalu are no longer necessary. 

For the TIOGA based overset, I am planning to remove this concept of inactivating nodes altogether so that we can reuse linear systems over multiple timesteps. So future commits will remove this bit of code altogether.

src/overset/TiogaSTKIface.C
sayerhs: @alanw0 Let me know if you still think adding _removed_ entities is useful, and I can change the code appropriately.

alanw0: That's a good point, I had forgotten that the two numbers were basically different things. I agree that adding 'removed' entities is not useful.

PR #110, opened 2019-02-07
WIP: Discuss design for transitioning field pointers to field ordinals in Kernels

sayerhs:
This pull request is to start discussions around possible designs to transition the field pointer references within Kernels to field ordinals and unit test them on CPU. Testing on GPU would also require `MasterElement` and `ipNodeMap_` data to be handled correctly on GPU. 

Currently, there are two commits. The first is just introducing some utility functions, the second modifies `MomentumMassElemKernel` for a working example of the transition from field pointer to ordinals. The unit test passes on CPU. 

```
$ ./unittestX --gtest_filter=MomentumKernelHex8Mesh.momentum_time_*
Note: Google Test filter = MomentumKernelHex8Mesh.momentum_time_*
[==========] Running 2 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 2 tests from MomentumKernelHex8Mesh
[ RUN      ] MomentumKernelHex8Mesh.momentum_time_derivative
[       OK ] MomentumKernelHex8Mesh.momentum_time_derivative (5 ms)
[ RUN      ] MomentumKernelHex8Mesh.momentum_time_derivative_lumped
[       OK ] MomentumKernelHex8Mesh.momentum_time_derivative_lumped (2 ms)
[----------] 2 tests from MomentumKernelHex8Mesh (7 ms total)

[----------] Global test environment tear-down
[==========] 2 tests from 1 test case ran. (7 ms total)
[  PASSED  ] 2 tests.
```

### DIFF
--- a/src/kernel/MomentumMassElemKernel.C
+++ b/src/kernel/MomentumMassElemKernel.C
@@ -32,8 +32,7 @@ MomentumMassElemKernel<AlgTraits>::MomentumMassElemKernel(
   ElemDataRequests& dataPreReqs,
   const bool lumpedMass)
   : Kernel(),
-    lumpedMass_(lumpedMass),
-    ipNodeMap_(sierra::nalu::MasterElementRepo::get_volume_master_element(AlgTraits::topo_)->ipNodeMap())
+    lumpedMass_(lumpedMass)
 {
 
   const stk::mesh::MetaData& metaData = bulkData.mesh_meta_data();
@@ -61,6 +60,8 @@ MomentumMassElemKernel<AlgTraits>::MomentumMassElemKernel(
 
   MasterElement* meSCV = sierra::nalu::MasterElementRepo::get_volume_master_element(AlgTraits::topo_);
 
+  // Initialize ipNodeMap_
+  get_scv_ipnodemap<AlgTraits>(meSCV->ipNodeMap(), ipNodeMap_);
   // compute shape function
   if ( lumpedMass_ )
     get_scv_shape_fn_data<AlgTraits>([&](double* ptr){meSCV->shifted_shape_fcn(ptr);}, v_shape_function_);
--- a/src/kernel/MomentumMassElemKernel.C
+++ b/src/kernel/MomentumMassElemKernel.C
@@ -14,6 +14,7 @@
 // template and scratch space
 #include "BuildTemplates.h"
 #include "ScratchViews.h"
+#include "utils/StkHelpers.h"
 
 // stk_mesh/base/fem
 #include <stk_mesh/base/Entity.hpp>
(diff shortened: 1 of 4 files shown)


src/kernel/MomentumMassElemKernel.C
```diff
@@ -71,14 +71,15 @@ MomentumMassElemKernel<AlgTraits>::MomentumMassElemKernel(
...
   // fields and data
-  dataPreReqs.add_coordinates_field(*coordinates_, AlgTraits::nDim_, CURRENT_COORDINATES);
-  dataPreReqs.add_gathered_nodal_field(*densityNm1_, 1);
-  dataPreReqs.add_gathered_nodal_field(*densityN_, 1);
-  dataPreReqs.add_gathered_nodal_field(*densityNp1_, 1);
-  dataPreReqs.add_gathered_nodal_field(*velocityNm1_, AlgTraits::nDim_);
-  dataPreReqs.add_gathered_nodal_field(*velocityN_, AlgTraits::nDim_);
-  dataPreReqs.add_gathered_nodal_field(*velocityNp1_, AlgTraits::nDim_);
-  dataPreReqs.add_gathered_nodal_field(*Gjp_, AlgTraits::nDim_);
+  const auto& fields = metaData.get_fields();
```
sayerhs: @alanw0 I am not crazy about this. Ideally it would be good if `ElemDataRequests` just could take field ordinals. However, it does quite a bit of stuff internally querying about `entity_rank` etc., so I am not sure it would be straightforward. Also doing this keeps the commit smaller and focused on just the conversion of the pointer.

src/kernel/MomentumMassElemKernel.C
```diff
@@ -61,6 +60,8 @@ MomentumMassElemKernel<AlgTraits>::MomentumMassElemKernel(
 
   MasterElement* meSCV = sierra::nalu::MasterElementRepo::get_volume_master_element(AlgTraits::topo_);
 
+  // Initialize ipNodeMap_
+  get_scv_ipnodemap<AlgTraits>(meSCV->ipNodeMap(), ipNodeMap_);
```
sayerhs: @alanw0 here is my proposed design for transitioning `ipNodeMap_` from a pointer to Kokkos::View, and follows the logic that was already there for shape functions.

overfelt: I thought the idea was that the ipNodeMap would be a static POD array or a Kokkos view in the master element so that it would be available on the GPU without copying it to a View and then deepcopying the view.

src/kernel/MomentumMassElemKernel.C
alanw0: Yes we could make ElemDataRequests take field ordinals, if we made ElemDataRequests hold a stk MetaData (it could be a constructor argument to ElemDataRequests.)

src/kernel/MomentumMassElemKernel.C
```diff
@@ -41,23 +42,22 @@ MomentumMassElemKernel<AlgTraits>::MomentumMassElemKernel(
   ScalarFieldType *density = metaData.get_field<ScalarFieldType>(
     stk::topology::NODE_RANK, "density");
 
-  velocityN_ = &(velocity->field_of_state(stk::mesh::StateN));
-  velocityNp1_ = &(velocity->field_of_state(stk::mesh::StateNP1));
+  velocityN_ = get_field_ordinal(metaData, "velocity", stk::mesh::StateN);
```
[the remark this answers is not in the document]
sayerhs: @alanw0 Yes, I was hoping to also get rid of the `velocity` variable, otherwise I'll switch to what you suggest.

src/kernel/MomentumMassElemKernel.C
rcknaus: I don't think there's any issue with that.  AlgTraits to MasterElement types aren't exactly one-to-one, but we need to write separate kernels for different discretizations anyway so we'd know if we're working with a Hex8FEM or Hex8CVFEM master element.

sayerhs: @overfelt I see what you are saying, `execute()` is entered for each element, so I didn't want to ask for ME instance everytime. However, it makes no difference since you have them already stored in a repo and you're returning the same instance even on GPU, so I think we can remove `ipNodeMap_` altogether. I assume this is where you were going?

overfelt: Yes,
meSCV=sierra::nalu::MasterElementRepo::get_volume_master_element<AlgTraits>();
will return a master element pointer for use on the GPU and ipNodeMap() can be called to return the array pointer that will also be on the GPU.  Since the master element is stored as a pointer on the class it does not have to be allocated every time execute() is called.

sayerhs: @overfelt OK. Got it. I assume we will want the same thing for 

```
  /// Shape functions
  AlignedViewType<DoubleType[AlgTraits::numScvIp_][AlgTraits::nodesPerElement_]> v_shape_function_ {"view_shape_func"};
```

overfelt: Yes, pass v_shape_function_ to:
template<typename ViewType> KOKKOS_FUNCTION void HexSCV::shape_fcn(ViewType &shpfc)
and it will fill it in just as the similar call does on the CPU.

overfelt: Note that for some reason the master elements do not keep these shape function evaluations around even though they are constant.  Maybe we should change that?

sayerhs: I was just going to say! Not sure why we compute it every time

rcknaus: For the low order elements, the cost of computing the shape functions is in the noise, so we didn't switch to saving them off.  But with the GPU, it makes sense just to save them.   The bigger one is the shape function derivatives which can also be saved (though also not performance critical).

sayerhs: It appears that in the kernel design, the shape functions were saved off, just not in the MasterElement instance but in the kernel instance.

PR #111, opened 2019-02-08
Add a Dirichlet inflow kernel for momentum

rcknaus:
Add a kernel method for Dirichlet inflow

* Generalized my high order MMS for use in regular kernels and the edge based scheme.  We didn't have a 3D pure fluids mms case, so I thought it was useful.  Also tested with the analytical onebytwobyten channel solution, which might be a good test to use for regression testing.

* I'm not sure how we'd handle different boundary condition implementations with the face/elem style kernels ATM.  It hasn't come up yet, but here I could have done just a pure advective flux inflow boundary and it would be nice to specify boundary condition options like we do in the elem kernel code

* Some clean-up for the node map construction.  Lets you use the tensor-product evaluation for regular p=1 elements from exodus without turning on promotion (if the USER_POLY_ORDER is 1).  I also switched the default user_poly_order to 1 since that's what I always use.

* Still need to do scalars.  But I thought this was a good chunk.


src/LowMachEquationSystem.C
```diff
@@ -423,6 +426,25 @@ LowMachEquationSystem::register_interior_algorithm(
...
+  stk::mesh::Part *part,
+  const stk::topology &theTopo,
+  const InflowBoundaryConditionData& /*inflowBCData*/)
+{
+  stk::mesh::MetaData &metaData = realm_.meta_data();
+  if ( !realm_.solutionOptions_->activateOpenMdotCorrection_ ) {
+    VectorFieldType *theBcField = &(metaData.declare_field<VectorFieldType>(stk::topology::NODE_RANK, "cont_velocity_bc"));
+    stk::mesh::put_field_on_mesh(*theBcField, *part, metaData.spatial_dimension(), nullptr);
+  }
+  // mdot at open bc; register field
```
sayerhs: Is this comment accurate?

src/LowMachEquationSystem.C
```diff
@@ -1569,6 +1598,9 @@ MomentumEquationSystem::register_inflow_bc(
     else if ( fcnName == "kovasznay") {
       theAuxFunc = new KovasznayVelocityAuxFunction(0,nDim);
     }
+    else if ( fcnName == "OneTwoTenVelocity" ) {
+      theAuxFunc = new OneTwoTenVelocityAuxFunction(0,nDim);
```
sayerhs: Can you please add a doxygen-style doc in `OneTwoTenVelocityAuxFunction.h` explaining what it is?

include/user_functions/TGMMSElemKernel.h
```diff
@@ -0,0 +1,51 @@
...
+#include <memory>
+
+namespace sierra{
+namespace nalu{
+
+class Realm;
+class ElemDataRequests;
+
+template <class AlgTraits>
+class TGMMSElemKernel final : public Kernel
```
sayerhs: Can you add a comment explaining TGMMS? Should there be an addition to verification section of the nalu docs?

unit_tests/kernels/UnitTestFaceElemBasic.C
```diff
@@ -223,3 +225,30 @@ TEST_F(Hex8ElementWithBCFields, faceElemScalarOpen)
 
   faceElemAlg.execute();
 }
+
+TEST_F(Hex8ElementWithBCFields, momentumInflowElemKernel)
```
sayerhs: Move this to `MomentumKernelHex8Mesh` for easy filtering.

include/user_functions/TGMMSElemKernel.h
rcknaus: Added a comment.  It's a simplified version of the Boussinesq MMS case.  It's nice for development IMO but it's just verifying a subset of physics that's already verified by a number of other verification tests, so I wasn't planning on doing separate documentation for it.

unit_tests/kernels/UnitTestFaceElemBasic.C
rcknaus: This probably should stay here.  It's just checking that the inflowelem kernel doesn't segfault essentially.  There's still some work we need to do on getting good unit tests for all these boundary condition kernels

sayerhs:
@rcknaus Please take a look at this and let us know how to proceed on this. Do you just want to fix the conflicts and merge, or do you want to do some changes in light of #427

sayerhs:
Closing due to lack of activity. Please reopen when ready.

PR #113, opened 2019-02-08
Mesh motion unit test addition and minor clean up

ashesh2512:
Added unit tests to test composite frame and motion capability specific to mesh motion framework. Also moved all mesh motion unit tests to a new folder under the unit tests directory. 

Further modified certain function calls to have const arguments for clean code, as determined while creating the above unit tests.

### DIFF
--- a/src/mesh_motion/MotionRotation.C
+++ b/src/mesh_motion/MotionRotation.C
@@ -40,13 +40,16 @@ void MotionRotation::load(const YAML::Node& node)
 }
 
 void MotionRotation::build_transformation(
-  const double time,
+  double time,
   const double*  /* xyz */)
 {
   double eps = std::numeric_limits<double>::epsilon();
 
-  if( (time >= (startTime_-eps)) && (time <= (endTime_+eps)) )
+  if(time >= (startTime_-eps))
   {
+    if(time >= (endTime_+eps))
+      time = endTime_;
+
     // determine current angle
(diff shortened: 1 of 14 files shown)


src/mesh_motion/MotionRotation.C
```diff
@@ -40,13 +40,16 @@ void MotionRotation::load(const YAML::Node& node)
...
+  double time,
   const double*  /* xyz */)
 {
   double eps = std::numeric_limits<double>::epsilon();
 
-  if( (time >= (startTime_-eps)) && (time <= (endTime_+eps)) )
+  if(time >= (startTime_-eps))
   {
+    if(time >= (endTime_+eps))
+      time = endTime_;
```
sayerhs: You are modifying a variable passed by value here. 1. What is the use case for modifying a variable like `time` within this method? and 2. pass by value variables won't have an effect outside of this method.

ashesh2512: 1) The use case is - If end time for a particular motion is less than the total simulation time. In such a case the mesh configuration for time > end_time should correspond to when time = end_time.
 
2) Following 1), the variable is meant to reflect the change only within this function.

ashesh2512: @sayerhs I have addressed the coding style issues in b05fabab6b4cbc4bcea938a616419fa8b34dd792

PR #117, opened 2019-02-11
Allow some usage of face elem kernels with promoted hex elements

rcknaus:
* Just hooks up face_grad_op for hexs
* This hooks up the ability to use face-elem kernels with the tensor-product CVFEM code.  The tensor-product stuff will need its own BCs eventually, but they can use the usual promoted element versions for the time being.  
* Some misc clean up

### DIFF
--- a/include/element_promotion/PromotedPartHelper.h
+++ b/include/element_promotion/PromotedPartHelper.h
@@ -33,8 +33,8 @@ namespace nalu {
 
   std::string super_element_part_name(std::string base_name);
 
-  std::string super_subset_part_name(const std::string& base_name, int numElemNodes, int numSideNodes);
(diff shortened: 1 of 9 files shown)


include/element_promotion/PromotedPartHelper.h
```diff
@@ -43,7 +43,7 @@ namespace nalu {
 
   void transform_to_super_elem_part_vector(stk::mesh::PartVector& parts);
 
-  stk::mesh::PartVector super_elem_part_vector(const stk::mesh::PartVector& parts);
+  stk::mesh::PartVector super_elem_part_vector(stk::mesh::PartVector parts);
```
psakievich: Why did you change this to pass by value?

PR #121, opened 2019-02-13
Jroverf/add alg traits to all master elements

overfelt:


src/master_element/MasterElementHO.C
```diff
@@ -997,7 +997,7 @@ HigherOrderQuad3DSCS::HigherOrderQuad3DSCS(
   quadrature_(std::move(quadrature))
 {
   surfaceDimension_ = 2;
-  nDim_ = 3;
+  ndim(3);
```
alanw0: Shouldn't this be AlgTraits::nDim_ instead of '3'?

src/master_element/MasterElementHO.C
```diff
@@ -1776,7 +1776,7 @@ HigherOrderEdge2DSCS::HigherOrderEdge2DSCS(
   basis_(basis),
   quadrature_(quadrature)
 {
-  nDim_ = 2;
+  ndim(2);
```
alanw0: AlgTraits::nDim_ instead of '2'?

overfelt: The higher order elements do not have an AlgTraits associated with it.  No fixed integration order for instance.

Issue #124, opened 2019-02-20
Dealing with geometric conservation law in presence of specified mesh motion


ashesh2512:
Since #77 we now have the capability of moving/deforming the mesh such that the cell volumes can change. This is important for when we might interface with a structural solver for FSI as well as for testing basic geometric conservation law with overset. However currently the only way to do so is through the `MeshDisplacement` class, some side effects of which are an extra equation system solve and numerical instability in my experience. It would be cleaner to have a class/function wherein the divergence of the velocity is computed for use by the `MomentumGclSrcNodeSuppAlg` and `ContinuityGclNodeSuppAlg`.

PR #129, opened 2019-03-01
Computation of divergence of mesh velocity for GCL

ashesh2512:
This pull request addresses #124 

The goal of this pull request is to create a standalone function for computation of the divergence of the vector field given a part vector. The motivation stems from the need to compute the divergence of mesh velocity for the GCL source term which was not possible prior to this pull request without invoking the`MeshDisplacementEquationSystem`which would in turn result in unnecessary solve of an extra system of equations. 

The pull request also includes a unit test that checks this functionality.  

The pull request also provides a pathway for computing divergence of the mesh velocity within the mesh motion framework.


gantech:
Looks good to me... Let's squash before commit.

include/mesh_motion/FrameBase.h
```diff
@@ -45,6 +39,8 @@ class FrameBase
   bool is_inertial() const {
     return isInertial_; }
 
+  virtual void post_work() {};
```
sayerhs: Remove semi-colon, will generate a warning with pedantic flags on.

src/mesh_motion/MotionPulsatingSphere.C
```diff
@@ -30,8 +40,9 @@ void MotionPulsatingSphere::load(const YAML::Node& node)
 
   get_if_present(node, "frequency", frequency_, frequency_);
 
-  origin_ = node["origin"].as<ThreeDVecType>();
-  assert(origin_.size() == threeDVecSize);
+  // get origin based on if it was defined
+  if( node["centroid"] )
```
sayerhs: Is this name change from `origin` to `centroid` in the input file intentional? Is this consistent with other mesh motion types?

ashesh2512: This was an error on my end when I committed it. It has always been `centroid` since #77. Here I made it consistent with the other motions.

include/mesh_motion/FrameBase.h
ashesh2512: This has been fixed in ba40395

PR #142, opened 2019-03-11
Modify Quad43D to avoid standard vector.

overfelt:


overfelt:
Yes, std::array does work on the GPU.  
I'm not sure it is guaranteed but it does work fine.  I speculate that since std::array is supposed to be as efficient as a plain old array, all of the member functions inline so do not have to be marked for the GPU.

PR #144, opened 2019-03-13
Working unit test of Kernel on a GPU

sayerhs:
Demonstrate a working prototype of the virtual Kernel::execute method executed
on GPU by invoking it through AssembleElemSolverAlgorithm::execute

- Introduced an intermediate subclass NGPKernel that inherits from Kernel but
  will be the base class for Kernels that are capable of running on GPUs.
  NGPKernel provides a concrete implementation of `create_on_device` that will
  create an instance of the Kernel on the device and move all data from host to
  device.

- Created a test Kernel that inherits from NGPKernel so that it can be executed
  on the GPU

- Extended AssembleElemSolverAlgorithm::execute that can transfer the kernels to
  the device, create a view of the device pointers, and call the execute method
  on the device.

- Added the necessary KOKKOS_FUNCTION decorators to get everything working.

- Introduced three unit tests: test create_on_device to move the Kernel to the
  GPU, test execution of a virtual method using run_algorithm, and test
  AssembleElemSolverAlgorithm::execute method.


sayerhs:
@alanw0 No, I don't think we need this on `MasterElement`, @overfelt has already created that infrastructure in `MasterElementRepo`. With appropriate `KOKKOS_FUNCTION` decorations to all the MasterElement constructors, we can register a pointer to the device instance to `ElemDataRequests::add_cvfem_surface_me` type calls.

include/kernel/Kernel.h
```diff
@@ -129,8 +132,67 @@ class Kernel
...
+ *
+ */
+struct NGPKernelInfo
+{
+  KOKKOS_FUNCTION
+  NGPKernelInfo() = default;
+
+  NGPKernelInfo(Kernel& kernel)
+  {
+    kernel_ = kernel.create_on_device();
```
overfelt: kernel.create_on_devce() always returns "this".  Why not just kernel_(&kernel).

sayerhs: @overfelt `NGPKernel::create_on_device` returns the device pointer and not `this`. I have a dummy function in the base `Kernel` class for migration purposes so that the code will compile for non-NGP ready kernels. `Kernel::create_on_device` should be a pure virtual function eventually. 

Although you are correct, I can do in the initializer list. 

```
kernel_(kernel.create_on_device())
```

PR #152, opened 2019-03-19
Add KOKKOS_FUNCTION to master element destructors.

alanw0:
Still need to do constructors, but this is step 0...


include/master_element/Edge32DCVFEM.h
```diff
@@ -38,6 +38,7 @@ class Edge32DSCS : public MasterElement
   using AlgTraits = AlgTraitsEdge3_2D;
 
   Edge32DSCS();
+  KOKKOS_FUNCTION
   virtual ~Edge32DSCS() {}
```
sayerhs: While we are doing this cleanup can we convert the empty bodies to

```
virtual ~Edge32DSCS() = default;
``` 

and move destructor bodies in the C file to headers?

include/master_element/Hex8CVFEM.h
```diff
@@ -51,6 +51,7 @@ class HexSCV : public MasterElement
   using AlgTraits = AlgTraitsHex8;
 
   HexSCV();
+  KOKKOS_FUNCTION
   virtual ~HexSCV();
```
sayerhs: If this is an empty body implementation in the C file we should probably move this to the header with 

```
virtual ~HexSCV() = default;
```

alanw0: Sure, I'll do that along with marking the constructors.

PR #153, opened 2019-03-19
Add KOKKOS_FUNCTION to MasterElement constructor and derived-class

alanw0:
constructors.
Also turned derived-class empty destructors into '= default'.
Can't do that for base-class destructor, it causes derived-class
destructors to issue compile errors.


include/master_element/MasterElement.h
```diff
@@ -43,13 +43,13 @@ struct ElementDescription;
...
 
-
 class MasterElement
 {
 public:
+  KOKKOS_FUNCTION
   MasterElement();
   KOKKOS_FUNCTION
-  virtual ~MasterElement();
+  virtual ~MasterElement() {} // = default is apparently not allowed for virtual destructors...
```
sayerhs: @alanw0 Curious why this won't work, but works fine in [Kernel.h](https://github.com/Exawind/nalu-wind/blob/master/include/kernel/Kernel.h#L102). Wondering if this is because of the `KOKKOS_FUNCTION` decorator here, but not in the other one.

alanw0: Yes I'm curious about that... Maybe I'll try it again and send you the compiler output.

PR #154, opened 2019-03-22
GCL term enhancement and related modifications in mesh motion framework

ashesh2512:
This pull request addresses the following points - 

1) The GCL term has been modified to account for the error due to discretization between `dV/dT` and `divV`. This requires dual nodal volume to have multiple states which have been introduced as and where required. This pull request is designed such that dual nodal volumes have multiple states only if the mesh moves.

2) The computation of `divV` in presence of mesh motion has been moved to after call to `compute_geometry`. Consequently `post_work` within the mesh motion framework has been changed to `post_compute_geometry` for an explicit understanding of anyone who reads the code.

3) General cleanup and minor modification of the mesh motion framework was performed including adding a rate-based scaling motion. This motion acts as a guideline for motions that may be added in the future involving deforming meshes.

### DIFF
--- a/src/HeatCondEquationSystem.C
+++ b/src/HeatCondEquationSystem.C
@@ -197,8 +197,9 @@ HeatCondEquationSystem::register_nodal_fields(
   tTmp_ =  &(meta_data.declare_field<ScalarFieldType>(stk::topology::NODE_RANK, "tTmp"));
   stk::mesh::put_field_on_mesh(*tTmp_, *part, nullptr);
 
-  dualNodalVolume_ = &(meta_data.declare_field<ScalarFieldType>(stk::topology::NODE_RANK, "dual_nodal_volume"));
+  dualNodalVolume_ = &(meta_data.declare_field<ScalarFieldType>(stk::topology::NODE_RANK, "dual_nodal_volume", numStates));
   stk::mesh::put_field_on_mesh(*dualNodalVolume_, *part, nullptr);
+  realm_.augment_restart_variable_list("dual_nodal_volume");
 
   coordinates_ =  &(meta_data.declare_field<VectorFieldType>(stk::topology::NODE_RANK, "coordinates"));
   stk::mesh::put_field_on_mesh(*coordinates_, *part, nDim, nullptr);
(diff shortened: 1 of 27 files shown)


src/HeatCondEquationSystem.C
```diff
@@ -230,6 +232,18 @@ HeatCondEquationSystem::register_nodal_fields(
...
+
+    ScalarFieldType &dualNdVolN = dualNodalVolume_->field_of_state(stk::mesh::StateN);
+    ScalarFieldType &dualNdVolNp1 = dualNodalVolume_->field_of_state(stk::mesh::StateNP1);
+
+    CopyFieldAlgorithm *theCopyAlgDlNdVol
+      = new CopyFieldAlgorithm(realm_, part,
+                               &dualNdVolNp1, &dualNdVolN,
+                               0, 1,
+                               stk::topology::NODE_RANK);
+    copyStateAlg_.push_back(theCopyAlgDlNdVol);
```
sayerhs: For BDF2, the number of states will be 3. Should also probably copy to `StateNM1`.

src/LowMachEquationSystem.C
```diff
@@ -311,20 +311,34 @@ LowMachEquationSystem::register_nodal_fields(
...
-    CopyFieldAlgorithm *theCopyAlg
+    CopyFieldAlgorithm *theCopyAlgDens
       = new CopyFieldAlgorithm(realm_, part,
                                &densityNp1, &densityN,
                                0, 1,
                                stk::topology::NODE_RANK);
-    copyStateAlg_.push_back(theCopyAlg);
+    copyStateAlg_.push_back(theCopyAlgDens);
+
+    if ( numVolStates <= 2 ) return;
```
sayerhs: Could this logic be isolated in a function that is called within the  EquationSystems?

ashesh2512: @sayerhs Such that it is inherited by all the derived equation classes instead of adding them separately in every equation system? I can do that.

src/HeatCondEquationSystem.C
ashesh2512: @sayerhs Throughout the code we copy fields only from NP1 to N and never N to NM1 for a restart. I was under the impression that was because the restart files save states only at NP1. Is that not correct?

src/LowMachEquationSystem.C
sayerhs: Sorry, I missed the fact that this was within a `restart` if-conditional.

sayerhs:
@ashesh2512 Please add the states logic to the other equation systems that _declare_field_ for `dual_nodal_volume`

PR #156, opened 2019-03-23
Use global node IDs to minimize fringe/field mismatch

sayerhs:
Pass global Node IDs to TIOGA API for use in determining unique nodes
across processor interface and minimize the fringe/field mismatches.


sayerhs:
@jrood-nrel No, this is more for the full turbine ABL simulation targeting the ECP FY19-Q2 milestone. @ashesh2512 is still looking into the `oversetRotCylinder`, we are unable to reproduce the indeterminate nature on Peregrine or Eagle, so that is still an open issue.

PR #172, opened 2019-04-01
Change output behavior of Nalu-Wind

sayerhs:
Currently, Nalu-Wind requires user to explicitly list the variables in the
output section. This commit changes the default behavior so that certain
variables are automatically added to the output list without the user explicitly
listing them in the `output_variables` list. The new list
`disabled_output_variables` is provided to explicitly disable output of these
variables.

### DIFF
--- a/include/OutputInfo.h
+++ b/include/OutputInfo.h
@@ -71,6 +71,8 @@ public:
   Ioss::PropertyManager *restartPropertyManager_;
 
   std::set<std::string> outputFieldNameSet_;
+  //! List of fields explicity disabled by user for output
+  std::set<std::string> disabledOutputSet_;
   std::set<std::string> restartFieldNameSet_;
 
 };
--- a/src/EnthalpyEquationSystem.C
+++ b/src/EnthalpyEquationSystem.C
@@ -279,6 +279,7 @@ EnthalpyEquationSystem::register_nodal_fields(
   // temperature required in restart
   temperature_ =  &(meta_data.declare_field<ScalarFieldType>(stk::topology::NODE_RANK, "temperature"));
   stk::mesh::put_field_on_mesh(*temperature_, *part, nullptr);
+  realm_.augment_output_variable_list("temperature");
   realm_.augment_restart_variable_list("temperature");
 
   dhdx_ =  &(meta_data.declare_field<VectorFieldType>(stk::topology::NODE_RANK, "dhdx"));
--- a/src/LowMachEquationSystem.C
+++ b/src/LowMachEquationSystem.C
@@ -1043,6 +1043,7 @@ MomentumEquationSystem::register_nodal_fields(
   // register dof; set it as a restart variable
   velocity_ =  &(meta_data.declare_field<VectorFieldType>(stk::topology::NODE_RANK, "velocity", numStates));
   stk::mesh::put_field_on_mesh(*velocity_, *part, nDim, nullptr);
+  realm_.augment_output_variable_list("velocity");
   realm_.augment_restart_variable_list("velocity");
 
   dudx_ =  &(meta_data.declare_field<GenericFieldType>(stk::topology::NODE_RANK, "dudx"));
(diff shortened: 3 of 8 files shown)


psakievich:
I feel the same as @rcknaus.  Could we instead have a flag for "output_default_variables" that will add in the common variables?  This accomplishes the desired result as well, but doesn't change historical behavior.

sayerhs:
@psakievich With your proposed change the user still has to remember to add `output_default_variables: true` in the input file right? Because if we default that to `true` then we won't retain historical behavior. 

The current setup is not end-user friendly, and it is easy not set up everything right and have to rerun a case, a good example is `mesh_displacement` which is not obvious to new users. The new behavior ends up adding a few extra variables that doesn't cause problems for users who prefer the old behavior because they can just ignore these fields. Are you concerned about disk usage? 

I also find the current behavior inconsistent because somehow it decides `pressureForce`, `yplus`, and `tauwall` are mandatory output variables even if the user didn't explicitly request them. Is the recommendation then that we remove them and have the user explicitly list them?

I agree with you on the inconsistency with `enthalpy` vs. `temperature`, my rationale was that we specify `temperature` for our initial and boundary conditions and so it stands to reason that we want to look at `temperature` fields, at least for wind energy applications.

rcknaus:
I guess I agree with adding mesh_displacement since that's a field people don't know they need.  Or  q-criterion, since people already specify they want that field in another place in the code.

For things like turbulent viscosity, I just think users will still need to know what fields they want to output, and this just complicates things for what they need to list or what they want to disable.

psakievich:
@sayerhs yes the user would need to include this line.  In my opinion it is a minimal impact/burden though because the end user will likely copy previous input decks and not create them from scratch.  

I agree that the current form is not new user friendly, mainly because new users don't know what fields to include and what fields are available to them. This new feature will help definitely help decrease the learning curve. I'm not too concerned about disk usage. I just think it is good practice to default to the more verbose syntax so the input decks are clear and behavior doesn't change in time (backwards compatibility, etc.) 

For example, if we were to wrap this in a gui, then I would want the gui to include every verbose detail in the generated input deck.  The end user would only see this when they transition to a power user, but it is helpful for developers/ user-support. I would be fine removing the inconsistency you pointed out as well.  

However, if the majority feel strongly that the default behavior should be inverted then I won't stand in the way.

sayerhs:
So just to make sure, this pull request doesn't remove `output_variables` entry in the input file. The user can continue explicitly list the variables (as in the current input files in regression tests). The user provided variable list is _merged_ with the default outputs. So in that sense, I would disagree that the old behavior is no longer available. The difference is that the user might see additional fields in the output files than those they explicitly listed. 

I can see the benefit in `disable_default_output_variables` (defaulted to `true`), that a power user can always set to `no` and take full control. 

I also think there is probably some discussion worth having about what fields should be defaulted. Given Nalu-Wind is wind energy focused I feel these are quantities that are most useful, others are welcome to add any that I missed. I am of the opinion that outputting more and letting user reduce it is better than having them state everything.

Fields for wind applications:

- `velocity`
- `pressure`
- `turbulent_ke`
- `turbulent_viscosity`
- `specific_dissipation_rate` (or whatever fields UT Austin model will use)
- `mesh_displacement`
- surface forces: `pressure_force`, `viscous_force` (if Surface post processing is added)
- `q-criterion`, `vorticity` (if Turbulent PP is added)
- `iblank` (if using TIOGA overset)

gantech:
@rcknaus  and @psakievich Thanks for your comments. I suppose this is a difference in philosophy on the input file design. My thinking is that the input file should be designed in such a way that for 99% of the use cases, the user should have to specify no inputs at all unless absolutely necessary. This prevents accidentally running a long simulation where they forgot to ask for one output that they wanted. The power users / developers can choose to specifically modify the input file for the remaining 1% of the use cases. For instance, I have not read any papers in the wind-energy field trying to plot enthalpy to analyze ABL flow-fields. However, if a user wants to do that, they can do so with the suggested approach. 

If you think that we should use a more verbose input file by default, then I suggest that we create an intermediate layer or a preprocessing step that generates the verbose input file from a much more simplified basic input file.

Nalu-wind is finally getting to a state where we are planning to do several production runs rather than one-off debugging runs. And I believe we should continue to trim the input file to require only the absolute minimum inputs and default all other inputs. The input file should also allow any power user/developer to change any/all inputs if necessary.

marchdf:
@sayerhs I like that default list as these are the fields I always want on hand to inspect and use for results plots. I might add `temperature` and `tau_wall` (if surface pp is turned on) to that list.

tonyinme:
Hey @sayerhs, this is useful. With fewer options the users have to specify in the input files, running the code becomes smoother.

PR #178, opened 2019-04-05
Fix array layout issue with NgpField in ScratchView gather 

sayerhs:
The `gather_elem_node_field_3D` type functions grabbed the address of of `field.get(ngpMesh, elemNodes[i], 0)` and then tried to access data sequentially using pointer manipulation. This is only true on Host memory-layout and not on device memory layout. This commit fixes the issue by calling `NgpField::get` method appropriately to copy data into `SharedMemView` arrays.

### DIFF
--- a/src/ScratchViews.C
+++ b/src/ScratchViews.C
@@ -55,12 +55,11 @@ void gather_elem_tensor_field(const NGPDoubleFieldType& field,
                               int tensorDim2,
                               ViewType& shmemView)
 {
-  const double* dataPtr = static_cast<const double*>(&field.get(elem, 0));
   unsigned counter = 0;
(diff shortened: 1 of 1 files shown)


sayerhs:
@alanw0 Question: Does [the element data gather](https://github.com/Exawind/nalu-wind/pull/178/files#diff-2663bc27b6f0ca1d3cf0e1ff1d0639fbL59) also need to be modified? Or is that data access sequential? I will hold off on merging this pull-request until we clarify this.

alanw0:
@sayerhs great catch, using the get function is clearly the only safe way to access the coefficients given underlying layout differences between host and device. I would say this does also apply to the element data gather. Even if the element data is currently contiguous, it could potentially be non-contiguous due to future changes.

PR #182, opened 2019-04-08
Convert ContinuityMassElemKernel to run on GPU

sayerhs:
- Remove non-POD types from class variables
- Switch `meSCV` to device instance
- Access shape functions via `ScratchViews`
- Enable unit-test of the ContinuityMassElemKernel (and related code)
- Currently only checks if the Kernel runs on GPU, actual values are not checked for since `apply_coeff` is still disabled within `AssembleElemSolverAlgorithm`


sayerhs:
@alanw0 @overfelt This pull request demonstrates the first actual Nalu-Wind Kernel executing on GPU, while retaining its behavior on host. This provides a template for converting other Kernels.

sayerhs:
@alanw0 Currently, there would be a performance penalty with the new way of doing things because the view will be updated everytime `SharedMemData` is created. The long term solution would be to have shape_functions in MasterElement and just grabbing a pointer (via ScratchViews or ME instance) within Kernels.

alanw0:
@sayerhs let me know when we're ready to start converting the rest of the kernels. This may be a good point to bring the other stk guys back into the ngp-conversion work.

sayerhs:
@alanw0 I also added ContinuityAdvElemKernel to provide coverage to both volume and surface master element. Can you take a look at the new commits? 

Let's discuss tomorrow on a couple of other nagging issues before we proceed with converting the remaining Kernels.

sayerhs:
@alanw0 the tests already have the test against gold values on CPU. However, the helper object's test LinearSystem inherits from LinearSystem which pulls in Realm, Equation System etc. Moving those to GPU will be quite an endeavor, which is why I didn't take that on at this point. I thought about a skinny test LinearSystem that doesn't inherit, but then we will have to modify AssembleElemSolver algorithm to behave differently on GPUs. `linsys_` is in Algorithm, so it affects more code.

Issue #185, opened 2019-04-10
Redesign ScratchViews/SharedMemData to allow execution on both host and device with CUDA builds


sayerhs:
The current ScratchViews design only allows `Kokkos::parallel_for` execution on the device execution space. Explore an alternate design that allows the user to perform `Kokkos::parallel_for` actions with `ScratchViews` on both host and device with CUDA builds.

sayerhs:
Closing this issue as it is of lower priority at the moment. Will revisit later if necessary

PR #188, opened 2019-04-10
Transition ContinuityMassElemKernel for execution on GPUs

sayerhs:
- Add ability to register and access ME shape functions via ScratchViews interface
- Add missing `shape_fcn` and `shifted_shape_fcn` implementations for Hex8 and Quad42D element types
- Create specializations of templated `get_surface` and `get_volume` master element in MasterElementRepo class for HO and non-HO element types. On GPUs HO element types will return `nullptr` and error out. 
- Remove all non-POD data types from ContinuityMassElemKernel class so that it can be copied to GPU
- Convert `meSCV` instance to be created on device
- Enable unit test for ContinuityMassElemKernel on GPUs


sayerhs:
This pull-request is a cleaned up version of #182, see that request for discussions.

sayerhs:
@alanw0 Tracking the ability to do parallel_for on host *and* device in #185

PR #195, opened 2019-04-11
Transition Momentum Mass and AdvDiff Kernels to GPU

sayerhs:
Follows the same formula used to convert Continuity interior kernels for two Momentum Kernels.

### DIFF
--- a/include/kernel/MomentumAdvDiffElemKernel.h
+++ b/include/kernel/MomentumAdvDiffElemKernel.h
@@ -26,7 +26,7 @@ class ElemDataRequests;
 /** Advection diffusion term for momentum equation (velocity DOF)
  */
 template<typename AlgTraits>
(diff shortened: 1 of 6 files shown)


sayerhs:
@alanw0 Good catch! With the new design we are delegating the responsibility of `shape_fcn` views to ScratchViews and getting a reference to the `SharedMemView` from within `execute` function. We need to know whether to request `shape_fcn` or `shifted_shape_fcn` for `v_adv_shape_function` within `execute` and that requires we carry around `skewSymmetric_` (`bool`) as a data member – see [this line in execute method](https://github.com/Exawind/nalu-wind/pull/195/files#diff-64e66e3baec324468ca1b246fa0edd18R84). The  _recipe_ right now is that we remove ME specific data from class members and then access them via ScratchViews within `execute`.

PR #203, opened 2019-04-23
Convert ScalarMassElemKernel for GPU operation.

alanw0:


include/kernel/ScalarMassElemKernel.h
```diff
@@ -73,8 +75,7 @@ class ScalarMassElemKernel: public Kernel
   /// Integration point to node mapping
   const int* ipNodeMap_;
```
sayerhs: Suggest removing the unused pointer

unit_tests/kernels/UnitTestKernelUtils.h
```diff
@@ -536,6 +540,7 @@ class MixtureFractionKernelHex8Mesh : public TestKernelHex8Mesh
   }
 
   ScalarFieldType* mixFraction_{nullptr};
+  ScalarFieldType* scalarQ_{nullptr};
```
sayerhs: `mixFraction_` is the `scalarQ_` that we should be using. No need to introduce another field here.

unit_tests/kernels/UnitTestScalarMassElem.C
```diff
@@ -0,0 +1,102 @@
...
+
+  // Populate LHS and RHS
+  helperObjs.assembleElemSolverAlg->execute();
+
+#ifndef KOKKOS_ENABLE_CUDA
+  EXPECT_EQ(helperObjs.linsys->lhs_.extent(0), 8u);
+  EXPECT_EQ(helperObjs.linsys->lhs_.extent(1), 8u);
+  EXPECT_EQ(helperObjs.linsys->rhs_.extent(0), 8u);
+
+//  std::cerr.precision(14);
```
sayerhs: Do we need to keep around this commented section? Might be useful to add a helper method in `HelperObjects` that could print out the lhs/rhs on demand (could be used elsewhere).

include/kernel/ScalarMassElemKernel.h
alanw0: Dang, how did I miss that?

unit_tests/kernels/UnitTestKernelUtils.h
alanw0: I needed to make it 2 states, so I made another field to make sure I wouldn't break anything else. But I can certainly make mixFraction be 2 states instead.

unit_tests/kernels/UnitTestScalarMassElem.C
alanw0: Yes I'll clean that up. I meant to do that already, and forgot.

alanw0:
@sayerhs I've now pushed a commit to address your comments.

PR #206, opened 2019-04-26
Convert std::vec to kk::view for High Order

psakievich:


src/master_element/MasterElementHO.C
```diff
@@ -260,7 +257,7 @@ void
 HigherOrderQuad3DSCS::area_vector(
   const double* POINTER_RESTRICT elemNodalCoords,
   const double* POINTER_RESTRICT shapeDeriv,
-  std::array<double,3>& areaVector) const
+  Kokkos::View<double[3]> areaVector) const
```
rcknaus: Kokkos::Array<double,3>& .  Kokkos::View may heap allocate.

psakievich: Ah, yeah I'll move back to std::array in my next pr

PR #209, opened 2019-04-29
Convert TurbKineticEnergySSTSrcElem to NGP

jrood-nrel:
I'm trying to follow the recipe for converting kernels and this is my first attempt. However, I'm currently trying to understand why it's segfaulting the unit test.

### DIFF
--- a/include/kernel/TurbKineticEnergySSTSrcElemKernel.h
+++ b/include/kernel/TurbKineticEnergySSTSrcElemKernel.h
@@ -23,7 +23,7 @@ class MasterElement;
 class ElemDataRequests;
 
 template <typename AlgTraits>
-class TurbKineticEnergySSTSrcElemKernel : public Kernel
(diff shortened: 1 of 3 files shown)


alanw0:
Hi Jon, I don't see any changes to the unit test in your commit. (UnitTestSSTSrcElem.C)

Currently the unit-test is pretty much all ifdef'd out for KOKKOS_ENABLE_CUDA. You should change it so that the unit-test is mostly enabled, but not the lines near the bottom of each test which compare the linsys gold values. The gold values are on the host and the linsys values are on the GPU so they can't be compared for now until we re-work something there to either get the gold values to the GPU or bring the linsys values back to the host.

jrood-nrel:
Thanks @alanw0 . I was able to test this today on the GPU and I fixed up some things I missed. I will probably merge it sometime today.

PR #213, opened 2019-05-01
Converting TurbKineticEnergyKsgsSrcElemKernel to NGP

jrood-nrel:
This is still a WIP.

This one was a bit more involved than the others I did. I had to add a unit test. It appears everything works on the CPU. I have two small questions regarding my implementation:
 1. I am curious if I implemented `KsgsKernelHex8Mesh` correctly?
 2. Why did I need `1.0e-14` `tol` when I used `print_lhs_and_rhs()` and used the values it emitted, when the other SST kernels I did used the `1.0e-15` default?

Also, what makes this a WIP is that at the moment it doesn't build with CUDA because when I put `calc_dual_nodal_volume` in `UnitTestKernelUtils.C` outside of the CUDA ifdefs (so it is live when CUDA is enabled) I can't get it to link with CUDA enabled because of:
```
unit_tests/kernels/UnitTestKernelUtils.C(599): error: no instance of function template "sierra::nalu::fill_pre_req_data" matches the argument list
            argument types are: (sierra::nalu::ElemDataRequestsGPU, ngp::Mesh, stk::topology::rank_t, stk::mesh::Entity, sierra::nalu::ScratchViews<double, sierra::nalu::TeamHandleType, sierra::nalu::HostShmem>)
```
Is this a known problem since `calc_dual_nodal_volume` is currently ifdef'd out? I'm guessing I'm just missing something.

### DIFF
--- a/include/kernel/TurbKineticEnergyKsgsSrcElemKernel.h
+++ b/include/kernel/TurbKineticEnergyKsgsSrcElemKernel.h
@@ -26,7 +26,7 @@ class ElemDataRequests;
 /** Add Ksgs source term for kernel-based algorithm approach
  */
 template<typename AlgTraits>
-class TurbKineticEnergyKsgsSrcElemKernel: public Kernel
+class TurbKineticEnergyKsgsSrcElemKernel: public NGPKernel<TurbKineticEnergyKsgsSrcElemKernel<AlgTraits>>
 {
 public:
   TurbKineticEnergyKsgsSrcElemKernel(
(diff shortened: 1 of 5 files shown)


jrood-nrel:
Ok, thanks. I did try increasing the precision in the `print_lhs_and_rhs` function, but during runtime it didn't help. Also when I used the explicit `exactValues` and `calcValues` for my `lhs` and `rhs`, it was still failing on `1.0e-15`. Anyway, I think I'm just missing something simple there.

I will look at `UnitTestScratchViews.C`.

sayerhs:
@jrood if you notice I reimplemented the `calc_mass_flow_rate` without using Kokkos version to avoid the TeamHandleType issue. You could use one that works with Kokkos type but you'd need to make sure that the field is copied back from device to host again so that ScratchViews can copy it back to device properly. I just felt it was unnecessary. If you do opt for doing the dual volume logic in device, then I'd recommend that we do this in a manner that can be used for all the functions currently ifdef'ed out in that file.

jrood-nrel:
Thanks @sayerhs . Makes sense. It looks like `calc_dual_nodal_volume` isn't that large, so I will just try to convert it and then we can look at all of the functions later to decide if they should be implemented on device in a consistent manner.

alanw0:
Shreyas' solution sounds better. It would be a pain to have to bring the values back to the host just so that ngp::Field can bring them back to the device where ScratchViews can gather them...

sayerhs:
@jrood-nrel also since this is a single Hex8 element case, I believe you could also do `stk::mesh::field_fill(*dualNodalVolume_, 0.125)` to initialize the field.

jrood-nrel:
@sayerhs Interesting. I think I will still try to convert `calc_dual_nodal_volume`, but if in place of `calc_dual_nodal_volume` I can use `field_fill`, maybe I will do that as plan B.

jrood-nrel:
I decided to use `field_fill` because I lack enough understanding of dual nodal volume, and the STK methods to replace the current Kokkos version of `calc_dual_nodal_volume` with a serial version. I tested this unit test on the CPU and GPU and it appears to run as expected.

PR #217, opened 2019-05-03
Convert MomentumBuoyancyBoussinesqSrcElemKernel to NGP

jrood-nrel:
I feel like I should have to replace `AlignedViewType<DoubleType[AlgTraits::nDim_]> gravity_{ "v_gravity"};` with something else here like we do with the shape functions, but this seems to run on the GPU. Let me know if that looks like a problem.

### DIFF
--- a/include/kernel/MomentumBuoyancyBoussinesqSrcElemKernel.h
+++ b/include/kernel/MomentumBuoyancyBoussinesqSrcElemKernel.h
@@ -56,7 +56,7 @@ private:
   double rhoRef_;
   double tRef_;
   double beta_;
-  AlignedViewType<DoubleType[AlgTraits::nDim_]> gravity_{ "v_gravity"};
+  NALU_ALIGNED DoubleType gravity_[3];
 
   MasterElement* meSCV_{nullptr};
 };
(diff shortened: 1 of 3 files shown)


sayerhs:
@jrood-nrel I have only looked at the code and not had a chance to compile/execute this on GPU build, so I am not sure how this is working. `AlignedViewType` is a `Kokkos::View` whose memory space defaults to the device, so the assignment to `gravity_` in the [constructor](https://github.com/Exawind/nalu-wind/pull/217/files#diff-82798366e6c59f0920fea4ab8cc41325L44) seems suspect. (@alanw0 @overfelt Does Kokkos allow assignment to a device View from host? I was under the impression that we had to use `HostMirror` followed by a `Kokkos::deep_copy`). Since we don't check the `lhs` and `rhs` when executing on GPUs, I would recommend that we `printf` the `gravity_` array while executing on device to make sure that it has the right data before we merge this commit (and #218). 

Is there a reason to not do 

```
NALU_ALIGNED DoubleType[3] gravity_;
```

instead of the `Kokkos::View` for this array? This would allow assignment on host within the constructor and get copied to device when `create_device_expression` is called.

overfelt:
Kokkos does not copy from host to device or back without an explicit request through deep_copy or some sync call.  That assignment to a device view on host should segfault.  But as @sayerhs says, fixed length views are prime candidates for replacement by arrays.

jrood-nrel:
Do these changes look better? Or can I not use:
```
const std::vector<double>& solnOptsGravity = solnOpts.get_gravity_vector(AlgTraits::nDim_);
```
because of the `std::vector` on the GPU? Eagle is down today so I haven't been able to test this on the GPU.

sayerhs:
@jrood-nrel The `solnOptsGravity` line is fine as the constructor is executed on host when the Kernel instance is created.

PR #225, opened 2019-05-10
WIP: Prototype NGP implementation of Edge algorithm

sayerhs:
- Create an NGP version of AssembleEdgeSolverAlgorithm
- Port ContinuityEdgeSolveAlgorithm to run on NGP version

### DIFF
--- a/include/AssembleEdgeSolverAlgorithm.h
+++ b/include/AssembleEdgeSolverAlgorithm.h
@@ -1,0 +1,109 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2019 National Renewable Energy Laboratory.                  */
+/*  This software is released under the license detailed                  */
+/*  in the file, LICENSE, which is located in the top-level Nalu          */
+/*  directory structure                                                   */
+/*------------------------------------------------------------------------*/
+
+#ifndef ASSEMBLEEDGESOLVERALGORITHM_H
+#define ASSEMBLEEDGESOLVERALGORITHM_H
+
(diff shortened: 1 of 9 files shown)


sayerhs:
@alanw0 @overfelt Here is my attempt at porting the Edge algorithm to NGP/GPU. After some back and forth, I ended up doing a hybrid of what we had discussed. I created `AssembleEdgeSolverAlgorithm` along the same lines as `AssembleElemSolverAlgorithm` with several simplifications possible for edge. Then the entire logic is implemented as a `lambda` in the body of the `execute()` method of the subclass (see `ContinuityEdgeSolverAlg` for an example). 

I ended up using `ScratchViews` because it gave nice way to access the field data for left/right states instead of the verbose `field.get(edge, nodeL, iDim)` and `field.get(edge, nodeR, iDim)` everywhere. Overall, this makes the edge algorithm look quite similar to the elem algorithm I feel. Of course, we will have to revert this if this has a performance penalty. 

Cc: @marchdf @jamelvin You might find this discussion useful for your edge implementation of TAMS algorithm.

sayerhs:
@alanw0 Yes, I think all SIMD logic can be encapsulated in `AssembleEdgeSolverAlgorithm` (and `SharedMemData_Edge` can just become `SharedMemData`) and the *edge kernels* themselves will be not need changes if we implement SIMD in the current implementation.

sayerhs:
Will pursue #226 design instead of the one proposed in this pull request.

PR #226, opened 2019-05-12
WIP: Prototype NGP Edge algorithm accessing ngp::Field directly

sayerhs:
Modify the NGP edge algorithm implemented in #225 to eliminate the intermediate ScratchViews and directly access the `ngp::Field` within the `lambda` of the algorithm. On CPU executions, with regression tests, this recovers the performance of the original implementation whereas #225 approach incurs >15% overhead in my preliminary tests. 

- I am unsure if the performance penalty of #225 might be overcome with SIMD, considering it would require copying overhead + interleave/de-interleave overhead. I feel this design implemented in this pull request will be difficult to implement with SIMD.
- I haven't explored if copying the vector data into local arrays will have any performance benefits.
- While the `VectorFieldType` accesses are a bit more verbose compared to the non-NGP and ScratchViews implementations, I feel it is not a huge barrier now that I look at it. 
- All the bucket looping and `sumInto` logic is encapsulated within `AssembleEdgeAlgorithm::run_algorithm` and the _edge kernels_ themselves still look very similar to the _element kernels_.

### DIFF
--- a/include/AssembleEdgeSolverAlgorithm.h
+++ b/include/AssembleEdgeSolverAlgorithm.h
@@ -1,0 +1,109 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2019 National Renewable Energy Laboratory.                  */
+/*  This software is released under the license detailed                  */
+/*  in the file, LICENSE, which is located in the top-level Nalu          */
+/*  directory structure                                                   */
+/*------------------------------------------------------------------------*/
+
+#ifndef ASSEMBLEEDGESOLVERALGORITHM_H
+#define ASSEMBLEEDGESOLVERALGORITHM_H
+
+#include "SolverAlgorithm.h"
+#include "ElemDataRequests.h"
+#include "ElemDataRequestsGPU.h"
+#include "Realm.h"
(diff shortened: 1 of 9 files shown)


sayerhs:
@overfelt I still need to run this on a GPU to make sure there aren't other issues that the design needs to address, so far I have been focusing on CPU tests to make sure that we are not losing performance with the modifications. If the team agrees to proceed with one of the two designs proposed, then I think we are ready to proceed with conversion of other algorithms. I haven't been able to come up with a third alternative so far. If you guys any other ideas you think we should pursue, we can look at those too before proceeding with the conversion. 

@alanw0 It would be interesting if `ngp::Field` could support SIMD datatypes. But this wouldn't be restricted to just `ngp::Field` correct? `ngp::Mesh` also needs to be aware of this to generate the buckets, connectedNodes, etc. differently depending on if the user is requesting SIMD or not. `ngp::Mesh` will also need to provide some way to interleave/de-interleave then?

sayerhs:
Closing this pull request. Will submit a new pull request that rebases the commits in this request and adds a unit test that will execute on GPUs.

PR #227, opened 2019-05-12
WIP: GPU-friendly coeff-applier

alanw0:
LinearSystem.h has a default, host-only CoeffApplier.
UnitTestLinearSystem has a GPU-capable CoeffApplier implementation.
UnitTestScalarAdvDiffElem.C is modified to check the lhs and rhs

Adding a CoeffApplier for TpetraLinearSystem will be less trivial,
but not too bad, I think.


unit_tests/UnitTestLinearSystem.h
```diff
@@ -10,16 +10,98 @@
...
+  {}
 
- TestLinearSystem( sierra::nalu::Realm &realm, const unsigned numDof, sierra::nalu::EquationSystem *eqSys)
+  KOKKOS_FUNCTION
+  TestCoeffApplier(const TestCoeffApplier&) = default;
+
+  KOKKOS_FUNCTION
+  ~TestCoeffApplier()
+  {
+    if (this != devicePointer_) {
```
sayerhs: In `NGPKernel` I moved this out of the destructor to avoid warnings regarding `kokkos_free_on_device`, how did you avoid those warnings here? I would recommend providing a `free_on_device` method that can be called from `TestLinearSystem`.

unit_tests/kernels/UnitTestScalarAdvDiffElem.C
```diff
@@ -66,14 +62,15 @@ TEST_F(MixtureFractionKernelHex8Mesh, NGP_advection_diffusion)
   // Populate LHS and RHS
   helperObjs.execute();
 
-#ifndef KOKKOS_ENABLE_CUDA
-  EXPECT_EQ(helperObjs.linsys->lhs_.extent(0), 8u);
-  EXPECT_EQ(helperObjs.linsys->lhs_.extent(1), 8u);
-  EXPECT_EQ(helperObjs.linsys->rhs_.extent(0), 8u);
+  Kokkos::deep_copy(helperObjs.linsys->hostlhs_, helperObjs.linsys->lhs_);
```
sayerhs: Since we know the sizes of `lhs_` and `rhs_`, would it make sense to consolidate the code from 65-70 within the body of `helperObjs.execute()` and reduce the verbosity in the `TEST_F` body?

unit_tests/kernels/UnitTestKernelUtils.h
```diff
@@ -172,23 +172,23 @@ void calc_projected_nodal_gradient(
   GenericFieldType& tensorField);
 
 void expect_all_near(
-  const Kokkos::View<double*>& calcValue,
+  const Kokkos::View<double*,Kokkos::LayoutRight,sierra::nalu::MemSpace>& calcValue,
```
sayerhs: I never noticed it until now, but `MemSpace` is not a good name to indicate `HostMemSpace`, should we rename this to be consistent with `DeviceMemSpace`, `DeviceShmem`, `HostShmem` etc.?

sayerhs:
Looks good! At least now we can sanity check the kernels that have been converted to GPUs.

unit_tests/UnitTestLinearSystem.h
alanw0: Yes I think your approach is better, I'll clean that up. I do have warnings about that.

unit_tests/kernels/UnitTestScalarAdvDiffElem.C
alanw0: I agree, I'll do that.

unit_tests/kernels/UnitTestKernelUtils.h
alanw0: Yes I'm confused about that memspace argument to expect_all_near. Renaming it would probably help. I probably don't have that correct in the TestLinearSystem views. I'm getting a compile error where on gpu I have to comment out one of the expect_all_near functions to avoid a 'ambiguous' error.

sayerhs: Is it the templated one that's coming up ambiguous?

alanw0: Here is the compiler output:
`
/scratch/william/source/nalu-wind-alanw0/unit_tests/kernels/UnitTestScalarAdvDiffElem.C(73): error: more than one instance of overloaded function "unit_test_kernel_utils::expect_all_near" matches the argument list:
            function "unit_test_kernel_utils::expect_all_near(const Kokkos::View<double *> &, const double *, double)"
            function "unit_test_kernel_utils::expect_all_near(const Kokkos::View<double **> &, const double *, double)"
            argument types are: (Kokkos::View<double *, Kokkos::Cuda::array_layout, Kokkos::Device<Kokkos::HostSpace::execution_space, Kokkos::Cuda::memory_space>>, const double [8]
`
Strangely, it seems to be confusing the 1D and 2D kokkos views... It gives the same message with and without the MemSpace template parameter.

alanw0: In KokkosInterface.h it looks like we have MemSpace defined as CudaUVMSpace so that's almost certainly not what we want...

sayerhs: Should the argument to `expect_all_near` be of type

```
Kokkos::View<double*,Kokkos::LayoutRight, sierra::nalu::MemSpace>::HostMirror
Kokkos::View<double**,Kokkos::LayoutRight, sierra::nalu::MemSpace>::HostMirror
```

sayerhs: @alanw0 How are you interpreting this 

```
Kokkos::View<double *, Kokkos::Cuda::array_layout,Kokkos::Device<Kokkos::HostSpace::execution_space, Kokkos::Cuda::memory_space>>
```

It seems to me that execution is HostSpace, but memory is CudaSpace?

alanw0: Yes I noticed that, it doesn't seem right. I wonder if it's related to MemSpace == CudaUVMSpace in KokkosInterface.h. Do you think we should set MemSpace to just be Kokkos::CudaSpace?

sayerhs: @alanw0 the `Kokkos::View` you are sending to the function is the hostview right? you are not comparing the gold values on device. So the memory space shouldn't be CUDA at all?

alanw0: Yes, sending the hostview, specifically the 'RHSView::HostMirror'. I'm going to try making MemSpace be CudaSpace, and then HostMirror should presumably have nothing to do with cuda. I'll let you know what that does.

alanw0: No combination I could think worked. I'm kind of stuck on this. I've resorted to renaming one of the expect_all_near functions to expect_all_near_2d. I'll push another commit to this pull-request and see what you think.

overfelt: Since it is just a unit test utility, seems a minor thing to have both expect_all_near and expect_all_near_2d.

sayerhs: I agree with @overfelt, renaming the function will clarify things.

alanw0:
@sayerhs I was going to merge this but I see I have a commit from Jon mixed in with my changes. I'm not sure what I need to do. Is that a problem? Do I need to create a new branch, get it up to date and move my changes to that branch before merging? Or will a  squash-and-merge do the right thing here?

sayerhs:
Squash and merge will do the right thing

PR #237, opened 2019-05-24
Reduce redundancy between examples and regression tests.

jrood-nrel:
I don't know exactly how the examples work so I don't know how to test this in depth. However, I think I'm mostly there. This PR should make a new target `make examples` that should build Nalu-Wind and DISCON. It copies the source tree of the examples to the build directory for running the examples there, and it symlinks `nrel5mw.fst` and `nrel5mw2.fst` for `turbine_uniform_inflow` to the one in the `nrel5MWactuatorLine` test, even when the tests aren't enabled. So this basically hooks the `turbine_uniform_inflow` example onto the setup for the `nrel5MWactuatorLine` test, eliminating redundancy between the two.

I haven't updated any documentation partly because I don't know if this will be the final design, and I think I will leave that to @tonyinme because it also seems out-of-date regarding Peregrine disappearing in July.


tonyinme:
Hey @jrood-nrel, thanks for addressing the issue. 
This will ease the compilation of the controller.

It might be a bit dangerous to have the openfast input files in the examples as symlinks because one could modify the regression test files intending to modify the examples.
Is there an option to copy the file instead of creating a symlink?

Thanks,
Tony

jrood-nrel:
@tonyinme I think that's a valid point, but I'm inclined to leave it so that it is more clear to the user that `nrel5mw.fst` in the examples begins as equivalent to the one in the test. It should be clear in their terminal if it has colors that it's a symlink and they can overwrite it after the cmake configure if they wish.

PR #239, opened 2019-05-28
Remove aggregation of header files.

jrood-nrel:
When I removed the globbing of source files in CMake, I also added the header listings, because I thought they might be useful for something, but they are just noise, so this removes them. I also changed some messages to `STATUS` messages.


sayerhs:
Since we do 

```

# and 

install(DIRECTORY include/ DESTINATION include)
```

A list of headers was not really used. The only use-case where I'd see a custom-list being useful is if we wanted to install a subset of the headers (nalu-wind public API), but right now we don't have anything like that. 

Looking at the `install` directive, I am noticing that the auto-generated headers from build directory are not transferred, just for future reference.

PR #241, opened 2019-05-29
WallDistEdgeSolver GPU conversion

jrood-nrel:
I'm not really sure if I'm close to done with this, but everything compiles and runs, except it's not clear to me why I get this error:
```
77: unknown file: Failure
77: C++ exception with description "Requirement( (field != nullptr) ) FAILED
77: Error occured at: my-nalu-wind/include/utils/StkHelpers.h:101
77: Error: Requested field does not exist: edge_area_vector
77: " thrown in the test body.
77: [  FAILED  ] WallDistKernelHex8Mesh.NGP_wall_dist_edge (2 ms)
```

### DIFF
--- a/include/edge_kernels/WallDistEdgeSolverAlg.h
+++ b/include/edge_kernels/WallDistEdgeSolverAlg.h
@@ -1,0 +1,36 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2019 National Renewable Energy Laboratory.                  */
+/*  This software is released under the license detailed                  */
+/*  in the file, LICENSE, which is located in the top-level Nalu          */
+/*  directory structure                                                   */
+/*------------------------------------------------------------------------*/
+
+#ifndef WALLDISTEDGESOLVERALG_H
+#define WALLDISTEDGESOLVERALG_H
+
+#include "AssembleEdgeSolverAlgorithm.h"
+
+namespace sierra {
+namespace nalu {
+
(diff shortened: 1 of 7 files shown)


sayerhs:
@jrood-nrel This was my mistake. I put the `edgeAreaVec` field under [LowMachKernelHex8Mesh](https://github.com/Exawind/nalu-wind/blob/master/unit_tests/kernels/UnitTestKernelUtils.h#L329) instead of in `TestKernelHex8Mesh`. `edgeAreaVec` will be needed for all equation systems, and it might be better if the logic associated with `edgeAreaVec` was moved into the base class so that we don't have to re-declare this in each equation system test class. This would mean that we will need [fill_mesh_and_init_fields](https://github.com/Exawind/nalu-wind/blob/master/unit_tests/kernels/UnitTestKernelUtils.h#L361) even in the base class, which the subclasses call before initializing their own fields. Do you want to take a stab at the redesign?

jrood-nrel:
I see. Thanks. I think I will try implementing `edgeAreaVector` in `WallDistKernelHex8Mesh` first and see if I can understand the commonalities between them enough to attempt the redesign.

jrood-nrel:
Ok this works on the CPU now. I need to test it on the GPU, and then I will see if I can push `edgeAreaVec` up level of classes.

jrood-nrel:
@sayerhs Does this look right?

PR #242, opened 2019-05-30
WIP: Preliminary design for NGP Nodal Kernels

sayerhs:
@alanw0 Here is a first attempt at NGP-design for the conversion of `*NodeSuppAlg`s to `Kernel` based design. Went down this design based on the following:

- We will have multiple source terms + *lumped mass time-derivative term* as nodal terms, so having a list of *kernels* for nodal terms that will be looped within an `AssembleNodeSolverAlg` seems to make sense to me. 
- Based on the limited experience with timings on edge algorithm (and CPU), it seemed that we should probably not use `ScratchViews` but access the `ngp::Field` directly. 
- Didn't see a good way to store a vector of `lamdba`s that could be transferred to CUDA easily. It seems we can't use `std::function` and will require [nvstd::function](https://devblogs.nvidia.com/new-compiler-features-cuda-8/). However, this was only a cursory research and I probably missed some other option. 

Regarding this first attempt:
- I opted to create a new base class `NodeKernel` instead of subclassing off of `Kernel` because it has a lot of element specific methods that we don't need. I felt this was cleaner as there is no need for these to be referred from the same base class pointer. 
- There is some duplication of `NGPNodeKernel` and `NGPKernel` logic, which would be nice to eliminate. 
- Regarding the kernel itself, I am not 100% happy about how the fields are accessed in the example kernel – [MomentumBodyForce](https://github.com/sayerhs/nalu-wind/blob/44a222744aafa6953f02a174c9fc53382764254b/src/node_kernels/MomentumBodyForceNodeKernel.C) (compare this to [NodeSuppAlg](https://github.com/Exawind/nalu-wind/blob/master/src/MomentumBodyForceSrcNodeSuppAlg.C)). 
  - Here I am extracting the field ID from bulk within the constructor, and accessing the actual instance of `ngp::Field` within the `setup()` method. We can't extract `ngp::Field` instance directly in the constructor because in cases of mesh motion this instance needs to be updated. We end up carrying the field ID as well as an `ngp::Field` instance. 
  - Is the assignment to `ngp::Field` and subsequent copy to GPU going to be cheap, or is this a costly operation? 
  - Is there a better way to do handle the fields?
  - Right now I am using `SharedMemView` for `lhs` and `rhs` although these could be static arrays, is there a preference?


sayerhs:
Thanks @alanw0, I'll continue working on this and have a working prototype for GPU.

PR #243, opened 2019-05-31
NGP ready version for Nodal source algorithms 

sayerhs:
Implement an NGP-version of `AssembleNodeSolverAlgorithm` and convert `MomentumBodyForceSrcSuppAlg` into a nodal kernel that can be executed on the GPU.


sayerhs:
@alanw0 Here is a proposed design for the nodal source `SupplementalAlgorithm`s that are needed for a full demo simulation with edge algorithm. 

- We discussed the `NodeKernel` design in #242, I didn't change that.
- I added an `AssembleNGPNodeSolverAlgorithm` as the old one was already called `AssembleNodeSolverSolverAlgorithm`. 
- I opted to keep the `SharedMemData_Node` implementation in the `.C` file for `AssembleNGPNodeSolverAlgorithm` as I didn't want the `SharedMemData.h` header to pull in ScratchViews etc when compiling this file. 
- I opted to use `unique_ptr` to store the `kernel` pointers in the vector, using an `add_kernel` method so that we don't have to deal with the deletion logic like we do in `Kernel` (and its unit tests).
- Updated the `TestEdgeLinearSystem` for handling the `3x3` momentum matrices per degree of freedom

PR #244, opened 2019-06-03
Fix "too many fields" warning message (#230)

### LINKED ISSUE #230: Too many fields warning message
In a case I am running in a development branch of mine, the following warning message

https://github.com/Exawind/nalu-wind/blob/b7f6a9756ababa5fd0a59d8de1f66d6d879c1196/include/MultiDimViews.h#L54

prints out an excessive number of times.  I can disable the warning, but @sayerhs wanted me to bring this up here to figure out the proper way to address this situation for NGP.

sayerhs:
Fix logic for handling views in MultiDimViews. Remove the `printf` statements
and replace by `NGP_ThrowRequire`

- Increase `maxFieldOrdinal` to 256

- Throw immediately if the number of requested views for any dimension is
  greater than maxViewsPerDim. Current sizes seem to be sufficient for the
  regression tests and other wind-energy related cases

- Add checks within each `add_*_view` method to check if an ordinal greater than
  the maxFieldOrdinal is actually being used, if not there is no need to abort

Closes #230.

### DIFF
--- a/include/MultiDimViews.h
+++ b/include/MultiDimViews.h
@@ -43,26 +43,17 @@ public:
 
   KOKKOS_FUNCTION
   MultiDimViews(const TEAMHANDLETYPE& /*team*/,
-                unsigned maxOrdinal,
+                unsigned /* maxOrdinal */,
                 const NumNeededViews& numNeededViews)
   : indices(),
     views_1D(), views_2D(), views_3D(), views_4D(), 
     views_1D_size(0), views_2D_size(0), views_3D_size(0), views_4D_size(0)
   {
-    if (maxOrdinal > maxFieldOrdinals)
-    {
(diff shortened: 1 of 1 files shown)


sayerhs:
@alanw0 I set the limit to 256 instead of the 128 we discussed because I see the following message in one of the regression tests

```
Too many fields (154), current limit in internal arrays is 64. Code will crash...
```

sayerhs:
@alanw0 An audit of field registration would be a good idea. Right now fields most fields are only registered if they are required (i.e., certain equation systems are activated, edge vs. elem fields are activated etc.). Also not all fields are registered on all parts, some are only registered on sidesets. Investigating the regression test that generated this warning, however, would be a good idea.

PR #247, opened 2019-06-11
Convert ScalarMassBDFNodeKernel to NGP

jrood-nrel:
I think I just need to put together a unit test for this.

### DIFF
--- a/unit_tests/node_kernels/UnitTestScalarMassBDFNodeKernel.C
+++ b/unit_tests/node_kernels/UnitTestScalarMassBDFNodeKernel.C
@@ -9,7 +9,7 @@
 #include "UnitTestUtils.h"
 #include "UnitTestHelperObjects.h"
 
-#include "node_kernels/ScalarMassBDF2NodeKernel.h"
+#include "node_kernels/ScalarMassBDFNodeKernel.h"
 
 #ifndef KOKKOS_ENABLE_CUDA
 namespace {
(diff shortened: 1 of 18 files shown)


unit_tests/node_kernels/UnitTestScalarMassBDFNodeKernel.C
```diff
@@ -0,0 +1,72 @@
...
+
+#include "node_kernels/ScalarMassBDFNodeKernel.h"
+
+#ifndef KOKKOS_ENABLE_CUDA
+namespace {
+namespace bdf2_golds {
+namespace scalar_mass {
+
+static constexpr double lhs[8][8] = {
+{0.10943331816113, 0, 0, 0, 0, 0, 0, 0, },
```
sayerhs: I am pretty certain these gold values are incorrect, the numbers along the diagonal should be 

```
gamma1 * rhoNp1 * dualVolume * dt = 1.0 * 1.0 * 0.125 / 0.1 = 1.25
```

I think we have a bug somewhere that we need to track down.

sayerhs: Mixture fraction deals with `density` field differently than other equation systems. So the previous comment is incorrect.

sayerhs:
@jrood-nrel @alanw0 I added the `lumped` version of `ScalarMassElemKernel` in commit 1ff76ff3, this matches well with the gold values you have for this NodeKernel (after accounting for node-ordering). So I feel confident that this NodeKernel has been implemented correctly.

jrood-nrel:
Thanks for verifying this. I will merge this one then. I can fix the momentum mass next week with the diff you provided.

PR #250, opened 2019-06-12
Support selection of input and output mesh format in YAML configuration file

fbudin69500:
Now that the SEACAS package of Trilinos has a factory for the ADIOS2 BP format, nalu-wind can read inputs and write outputs in that format. In this pull-request, these inputs and outputs are recognized based on the extension of the file. To test that this new format saves all the necessary information, the tests have been updated to use this format.

### DIFF
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -171,6 +171,7 @@ namespace nalu{
   : realms_(realms),
     name_("na"),
     type_("multi_physics"),
+    meshDBType_("exodus"),
     inputDBName_("input_unknown"),
     spatialDimension_(3u),  // for convenience; can always get it from meta data
     realmUsesEdges_(false),
@@ -667,6 +668,7 @@ Realm::load(const YAML::Node & node)
   name_ = node["name"].as<std::string>() ;
   inputDBName_ = node["mesh"].as<std::string>() ;
   get_if_present(node, "type", type_, type_);
+  get_if_present(node, "mesh_db_type", meshDBType_, meshDBType_);
 
   // provide a high level banner
   NaluEnv::self().naluOutputP0() << std::endl;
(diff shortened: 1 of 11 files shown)


fbudin69500:
@chuckatkins

reg_tests/CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ else(NOT ${TEST_TOLERANCE} STREQUAL "")
     if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
       set(TOLERANCE 0.0001)
     elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-      set(TOLERANCE 0.000000000000001)
+      set(TOLERANCE 0.00001)
```
sayerhs: Please revert this change.

src/Realm.C
```diff
@@ -2085,10 +2097,15 @@ Realm::create_mesh()
   }
 
   // Initialize meta data (from exodus file); can possibly be a restart file..
-  inputMeshIdx_ = ioBroker_->add_mesh_database( 
-   inputDBName_, restarted_simulation() ? stk::io::READ_RESTART : stk::io::READ_MESH );
+  if(ends_with(inputDBName_, "bp")) {
```
sayerhs: I would recommend that instead of depending on some extension type, we allow users to provide a `mesh_db_type: adios` and use that to read files, and set the default `mesh_db_type: exodus`

sayerhs:
One more question: I assume that this will require us to have trilinos builds with ADIOS dependencies enabled? If so, I would further recommend that this be made optional via a CMake option (like Catalyst interface for example) so that it doesn't require users to incur this overhead if they don't need it.

fbudin69500:
Thanks for the great feedback. I can remove the commit that lowers the tolerance. This is something I need to do on my local computer for the tests to pass, both with this new file format but also with the current file format.

I will address the two other concerns and follow the good suggestions, and update this PR.

sayerhs:
@fbudin69500 Regarding the test tolerance for your local machine, you can pass `-DTEST_TOLERANCE=<desired_value>` during the CMake configure phase to set a custom tolerance on a per-build basis. This will override the default tolerance logic that only kicks in if the user hasn't passed a `TEST_TOLERANCE` variable [see here](https://github.com/Exawind/nalu-wind/blob/master/reg_tests/CMakeLists.txt#L16).

sayerhs:
Closing due to inactivity. Please reopen when ready

PR #254, opened 2019-06-16
Jroverf/momentum actuator node kernel

overfelt:


include/node_kernels/MomentumActuatorNodeKernel.h
```diff
@@ -0,0 +1,54 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2019 National Renewable Energy Laboratory.                  */
```
sayerhs: @overfelt You might want to assert Sandia copyright assertion here.

include/node_kernels/MomentumActuatorNodeKernel.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include "node_kernels/NodeKernel.h"
+
+#include "stk_mesh/base/BulkData.hpp"
+#include "stk_ngp/Ngp.hpp"
+
+namespace sierra{
+namespace nalu{
+
+class SolutionOptions;
```
sayerhs: Forward declaration not used?

include/node_kernels/MomentumActuatorNodeKernel.h
overfelt: Interesting.  Should we have both NREL and SNL in all files for consistency?  Or can only one of us assert copyright at a time?

include/node_kernels/MomentumActuatorNodeKernel.h
overfelt: Yes, I missed that one.

include/node_kernels/MomentumActuatorNodeKernel.h
sayerhs: @overfelt We will have to wait for Mike and Paul to weigh in on this one. I think they are working on getting a joint copyright assertion for the codebase.

PR #259, opened 2019-06-19
Converting ContinuityMassBDFNodeKernel to NGP

jrood-nrel:


src/node_kernels/ContinuityMassBDFNodeKernel.C
```diff
@@ -0,0 +1,68 @@
...
+  const stk::mesh::FastMeshIndex& node)
+{
+  const NodeKernelTraits::DblType rhoNm1        = densityNm1_.get(node, 0);
+  const NodeKernelTraits::DblType rhoN          = densityN_.get(node, 0);
+  const NodeKernelTraits::DblType rhoNp1        = densityNp1_.get(node, 0);
+  const NodeKernelTraits::DblType dualVolume    = dualNodalVolume_.get(node, 0);
+  const NodeKernelTraits::DblType projTimeScale = dt_/gamma1_;
+
+  rhs(0) -= (gamma1_*rhoNp1 + gamma2_*rhoN + gamma3_*rhoNm1)*dualVolume/dt_/projTimeScale;
+  lhs(0, 0) += 0.0;
```
alanw0: Seems like there's no reason to do '+= 0.0' ...

jrood-nrel: Agreed. It was just in the original. Also, I think `dt` cancels out as well.

PR #258, opened 2019-06-18
TpetraSegregatedLinearSystem: preparing the linear system/linear solver files (see issue #249)

lucbv:
There are a reasonable amount of changes in this PR, here is a quick summary:

1. I am adding the `PT_TPETRA_SEGREGATED` in LinearSolver.h and changing the logic in `TpetraLinearSystem::getType()` to prepare it for my next PR.
2. I am creating `TpetraLinearSystemHelpers` that gathers all the helper functions from `TpetraLinearSystem`, the idea it to then reuse them in `TpetraSegregatedLinearSystem` without code duplication.
3. As the helper functions are instantiated in their `.C` file I am removing unnecessary template arguments and replacing them with `typedef` (see LinearSolverTypes.h) to actually use the concrete type of parameters in function signatures.
4. I removed functions that were not called anywhere in Nalu.

I also strongly debated whether I should remove the following macros:
```
#define LID_(lid, ndof, idof)  ((ndof)*((lid))+(idof))

#define GLOBAL_ENTITY_ID(gid, ndof) ((gid-1)/ndof + 1)
#define GLOBAL_ENTITY_ID_IDOF(gid, ndof) ((gid-1) % ndof)
```
The 1st one does not appear to be used in the code and the 2nd and 3rd ones are used only once which makes them a little superfluous. Any opinion on that?

I already have the start of the next PR for issue #249 with the actual `TpetraSegregatedLinearSystem.{h/C}` but I would rather not introduce them here to compartmentalize changes in different commits/PRs.

PR #260, opened 2019-06-19
Removing macros from TpetraLinearSystemHelpers see comment in PR #258

lucbv:
@sayerhs @alanw0 @rcknaus 
Three macros are removed:
      1) LID_(), since it was actually not used anymore
      2) GLOBAL_ENTITY_ID() and GLOBAL_ENTITY_ID_IDOF() are replaced by direct implementation

This mainly cleans the code and macros always have a small potential for unexpected side effect...

### DIFF
--- a/src/TpetraLinearSystem.C
+++ b/src/TpetraLinearSystem.C
@@ -1649,12 +1649,12 @@ bool TpetraLinearSystem::checkForZeroRow(bool useOwned, bool doThrow, bool doPri
     if (global_row_exists[ii] && bulkData.parallel_rank() == 0 && row_sum < 1.e-10) {
       found = true;
       GlobalOrdinal gid = ii+1;
-      stk::mesh::EntityId nid = GLOBAL_ENTITY_ID(gid, numDof_);
+      stk::mesh::EntityId nid = (gid - 1) / numDof_ + 1;
       stk::mesh::Entity node = bulkData.get_entity(stk::topology::NODE_RANK, nid);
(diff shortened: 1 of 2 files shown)


src/TpetraLinearSystem.C
```diff
@@ -1649,12 +1649,12 @@ bool TpetraLinearSystem::checkForZeroRow(bool useOwned, bool doThrow, bool doPri
     if (global_row_exists[ii] && bulkData.parallel_rank() == 0 && row_sum < 1.e-10) {
       found = true;
       GlobalOrdinal gid = ii+1;
-      stk::mesh::EntityId nid = GLOBAL_ENTITY_ID(gid, numDof_);
+      stk::mesh::EntityId nid = (gid - 1) / numDof_ + 1;
       stk::mesh::Entity node = bulkData.get_entity(stk::topology::NODE_RANK, nid);
```
alanw0: Hmm, I think your change is ok, but I didn't realize we had code like this. That's a weird way of coming up with a global node id.

lucbv: @alanw0 I'm happy to confess that I do not know why this is how the code computes the entityId, but I hope that this is compatible with stk's algorithms!

lucbv:
This is a fairly straight forward PR with minor changes.
No diffs with my latest trilinos/nalu-wind build : )
[ctest.log](https://github.com/Exawind/nalu-wind/files/3307225/ctest.log)

PR #261, opened 2019-06-19
Convert ContinuityMassBackwardEulerNodeKernel to NGP

jrood-nrel:


src/node_kernels/ContinuityMassBackwardEulerNodeKernel.C
```diff
@@ -0,0 +1,55 @@
...
+ContinuityMassBackwardEulerNodeKernel::execute(
+  NodeKernelTraits::LhsType& /*lhs*/,
+  NodeKernelTraits::RhsType& rhs,
+  const stk::mesh::FastMeshIndex& node)
+{
+  const NodeKernelTraits::DblType projTimeScale = dt_;
+  const NodeKernelTraits::DblType rhoN       = densityN_.get(node, 0);
+  const NodeKernelTraits::DblType rhoNp1     = densityNp1_.get(node, 0);
+  const NodeKernelTraits::DblType dualVolume = dualNodalVolume_.get(node, 0);
+  rhs(0) -= (rhoNp1 - rhoN)*dualVolume/dt_/projTimeScale;
```
alanw0: Looks like dt_ cancels out again here. Given how little this kernel is doing, removing those divides might be a significant speedup.

sayerhs:
@jrood-nrel I think we could collapse the backward Euler and BDF2 the same way we collapsed `MomentumMassBDFNodeKernel` and `ContinuityMassElemKernel`.

jrood-nrel:
Ok I will try it. They look really similar.

sayerhs:
#259 already handles this, there is nothing for you to do :)

src/node_kernels/ContinuityMassBackwardEulerNodeKernel.C
jrood-nrel: I was hasty in my math on the cancelling out, but it looks like we can at least get rid of some divides by multiplying by (dt*dt) and it appears to not lose precision when I test it.

jrood-nrel: Scratch that. I meant we can switch a divide for multiply in #259 . Using dt^2 here loses precision.

jrood-nrel:
I don't think I'm following. Is it the case that the `gamma` values get set here such that they reduce to `rhs(0) -= (rhoNp1 - rhoN)*dualVolume/dt_/dt_;`?

sayerhs:
Yes, that's correct. In fact, your regression test is actually testing the `BDF1` algorithm with `gamma1 = 1.0; gamma2 = -1.0` and not BDF2.

jrood-nrel:
Ah I see. I'm glossing over this stuff instead of spending a moment to understand what I'm even doing 🙂

jrood-nrel:
So this PR seems redundant to #259 .

PR #262, opened 2019-06-19
Remove the gtest_shuffle flag since it occasionally runs tests in a

alanw0:
different order on different mpi procs. They need to be in lock-step
across mpi procs.

Also changed some unit-test output to use cerr instead of cout, it seems
like some of the cout output wasn't coming out due to weird buffering or
something.

### DIFF
--- a/reg_tests/CTestList.cmake
+++ b/reg_tests/CTestList.cmake
@@ -46,7 +46,7 @@ endfunction(add_test_r_np)
 
 # Standard unit test
 function(add_test_u testname np)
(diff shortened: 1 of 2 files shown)


alanw0:
@sayerhs that sounds like a good idea. @jrood-nrel is there a way to use conditional logic based on num-procs in this CTestList.cmake file?

jrood-nrel:
Thanks for solving this Alan. I made another PR if we want to keep shuffling on for unitTest1.

alanw0:
Closing this since @jrood-nrel has a better solution.

jrood-nrel:
I didn't mean to counteract the s`td::cerr` changes though 🙂

alanw0:
@jrood-nrel no problem, I'll put those in separately.

PR #263, opened 2019-06-20
Add parameter for unit test function to turn shuffling on or off

jrood-nrel:


jrood-nrel:
Actually let me change this to check if `np>1` and turn off shuffling then. I think Alan's suggestion is better than how I first implemented this.

jrood-nrel:
I'm going to merge this before I leave for vacation today. @sayerhs Feel free to suggest anything here and I can change it when I get back with another PR.

Issue #249, opened 2019-06-12
Tpetra Segregated System


lucbv:
This issue is here to keep track of progress made to implement a segregated linear system with the Tpetra stack in nail-wind.
There is already a working implementation in my fork but rebasing on head and merging afterwards seems quite hard due to many changes in `nalu-wind`.
I will take an incremental approach in which I slowly introduce all the features I need to get to the full segregated linear system. This should make it easy to review my PRs.

1. [x] Switch from `LinSys::Vector` to `LinSys::MultiVector` in `LinearSolver`, `LinearSystem` and `TpetraLinearSystem` classes. (done through PR #251 )
2. [x] Create a new type of linear system: `TpetraSegregatedLinearSystem` this will initially be implemented as a copy of the `TpetraLinearSystem` class (done in PR #258 and #260)
3. [x] Actually implement and interface the specific methods for a segregated system.

alanw0:
Do we really need a new linear system class? That sounds like a lot of code duplication. I don't know how specifically your work is implemented, but I was hoping it would make the momentum equation more similar to the scalar equations. Adding a new linear system class sounds like it would be very different. I'll be interested to hear/see more details.

sayerhs:
@alanw0 While don't need an entirely new class, you'll want to override `sumInto` class so that even though the algorithms are `3x3` we are just populating a scalar matrix. In `HypreLinearSystem` vs. `HypreUVWLinearSystem` the differences are in the logic of handling `numDof_`, `sumInto` and the `loadComplete` steps.

alanw0:
@sayerhs does sumInto need to only keep the 0,0 position of each 3x3 and discard the rest, or does it need to do something fancier than that? If it is just discarding most of the coefficients, maybe we should pass an option to the relevant kernels so that they don't waste time calculating those coefficients.

sayerhs:
@sayerhs In scalar/continuity kernels `lhs` is just `(1, 1)`. For momentum, the time derivative kernel should populate the diagonal `(i, i) i = 1..3` and not just `(0, 0)`. I'll have to look through the other kernels to make sure that they are not populate off-diagonal terms.

lucbv:
@alanw0 @sayerhs I did not plan to touch anything but LinearSolver and LinearSystem, I did try to do something fancier initially but realized quickly that there are a lot of physic kernels that would need to be changed otherwise.
[...]

PR #265, opened 2019-06-24
TpetraSegregatedSystem: adding the new TpetraSegregatedLinearSystem class (see issue #249)

lucbv:
Create the new class `TpetraSegregatedLinearSystem` and adds the necessary hooks in LinearSystem calss and CMakeList.txt
Test performed with dgNonConformalThreeBlade shows convergence of the new system and keeps two digits of accuracy when compared to non segregated system.


lucbv:
@sayerhs @rcknaus @alanw0 @jhux2
Here is the last PR for the TpetraSegregatedLinearSystem.
I did a check with `dgNonConformalThreeBlade`, I am attaching corresponding input, output and norm files for both the current and new approach.
As expected there is a certain amount of difference between the two code path as they solve different linear equations within each Newton iteration.
Let me know if you have an idea on how we want to test/check that the new code path is robust enough

[dgNonConformalThreeBlade_seg.i.txt](https://github.com/Exawind/nalu-wind/files/3321300/dgNonConformalThreeBlade_seg.i.txt)
[dgNonConformalThreeBlade_seg.log.txt](https://github.com/Exawind/nalu-wind/files/3321301/dgNonConformalThreeBlade_seg.log.txt)
[dgNonConformalThreeBlade_seg.norm.txt](https://github.com/Exawind/nalu-wind/files/3321302/dgNonConformalThreeBlade_seg.norm.txt)
[dgNonConformalThreeBlade.i.txt](https://github.com/Exawind/nalu-wind/files/3321307/dgNonConformalThreeBlade.i.txt)
[dgNonConformalThreeBlade.norm.gold.txt](https://github.com/Exawind/nalu-wind/files/3321308/dgNonConformalThreeBlade.norm.gold.txt)

sayerhs:
@lucbv Can you run the following tests

```
BoussinesqNonIso     
cvfemHexHC_P3        
hoVortex_P2          
steadyTaylorVortex_P4
```

These are tests that use norms to ensure consistency and are better when checking that the solver is indeed converging to the same solution even though there might be differences in the iteration residuals.

lucbv:
@alanw0 the difference observed are consistent with the tolerances set for the non-linear solver i.e. 1e-5 and the linear solver is also converging to the proper tolerance. We could ask for tighter tolerance on the non-linear solver and we should get more and more accurate results but not necessarily more digits as we are trying to drive the non-linear residual down to zero.

@sayerhs I will run the tests you are suggesting and report on the convergence behavior.

sayerhs:
@lucbv We should think about ways to reduce the amount of code duplication between `TpetraLinearSystem` and `TpetraSegregatedLinearSystem`, may be there needs to be a `TpetraBaseLinearSystem` that implements the common stuff like which can function as long as the `numDof_` is set correctly. From Hypre experience, the only changes necessary are to the `sumInto` methods as they are the only ones that specifically need to know and handle `3x3` matrix per DOF and handle them differently.

lucbv:
@sayerhs you actually need to keep `numDof_=3` even if you intent to compute a segregated system as the assembly classes query it and will not compute the proper right hand side if you set `numDof_=1`.
I did not feel confident that I could rewrite most of the `AssembleXYZ.h` classes and decided to work around that constraint. Since `numDof_` is not set properly I then need to rewrite minor pieces of code in most of the class to ensure that the graph, matrix and multivectors have the proper dimensions...

sayerhs:
@lucbv If you look at Hypre implementations, you will notice that I didn't have to change any of the `Assemble*Algorithm` classes. The assemble algorithm classes use `numDof()` method to query the `linsys_` object and you can override that specifically to pass the desired value to the algorithm classes ([see here](https://github.com/Exawind/nalu-wind/blob/master/include/HypreUVWLinearSystem.h#L115)). The key difference between `HypreLinearSystem` and `HypreUVWLinearSystem` is the `sumInto` method which handles both the matrix and rhs correctly. So I am confident that it can be done in TpetraLinearSystem also without having to touch any of the `Assemble*` classes.

lucbv:
@sayerhs that's a good point. I will continue the checks with the current version of the code since I already started and I will look at refactoring the class afterwards.
I am looking at the `cvfemHex_P3` test and it is not clear to me that it solves a momentum equation, at least the input file does not mention it explicitly. Do you have any guidance on how to modify the input file to introduce the segregated momentum solver in it?

sayerhs:
@lucbv Ach! My mistake, I should have removed that test from the list. It is a heat conduction equation and doesn't have momentum. So ignore that test.

lucbv:
@sayerhs 
Consistency tests results:
1. BoussinesqNonIso
    The two resolutions run fine and the following values are computed for check by `norms.py` (values need to be less than -1.95 for success)
    `{'velocity[2]': -2.035980357922834, 'temperature[0]': -2.321845238521564, 'velocity[1]': -2.0653046753468347, 'velocity[0]': -2.0547198407487035}`

2. hoVortex_P2
    The two resolutions run fine and the following values are computed for check by `norms.py` (values need to be less than -1.0 resp. -2.4 for gradients resp. velocities, for success)
    `{'dpdx[0]': -1.472554604045194, 'dpdx[1]': -1.4056280282601394, 'velocity[1]': -2.4489785522100886, 'velocity[0]': -2.6363175235957317}`
    `{'dpdx[0]': -1.2780348853785448, 'dpdx[1]': -1.2700472472943607, 'velocity[1]': -3.0336486571373444, 'velocity[0]': -3.1448001881562955}`

2. steadyTayloVortex_P4
    The two resolutions run fine and the following values are computed for check by `norms.py` (values need to be less than -3.0 resp. -5.0 for gradients resp. velocities, for success)
    `{'dpdx[0]': -4.159774393179704, 'dpdx[1]': -4.342380906738279, 'velocity[1]': -6.240633275889772, 'velocity[0]': -6.036965172019971}`
    `{'dpdx[0]': -3.614282298322727, 'dpdx[1]': -3.791167746003428, 'velocity[1]': -5.820325211352537, 'velocity[0]': -5.782587726859437}`

rcknaus:
The tests look good.  This isn't part of the pull request, but do we know what the effect of doing this on nonlinear convergence for an LES is?

Otherwise, the code looks fine near term.  Like Shreyas mentioned, we should probably look at splitting up TpetraLinearSystem.  I like the idea of having a mesh handling class that converts between the stk ids and the linear system ids, having a separate class for dealing with the graph, and completely splitting out the linear solve out from the linear system.

src/TpetraSegregatedLinearSystem.C
```diff
@@ -0,0 +1,1829 @@
...
+  }
+
+  copy_tpetra_to_stk(sln_, linearSolutionField);
+  sync_field(linearSolutionField);
+
+  // computeL2 norm
+  Teuchos::Array<double> mv_norm(1);
+  ownedRhs_->norm2(mv_norm());
+  const double norm2 = mv_norm[0];
+
```
rcknaus: Does this need an all-reduce?

sayerhs:
@rcknaus Regarding the _non-linear convergence of LES_, you bring up a good point. Can you share your thoughts on how you think we should test that?

rcknaus:
Set a nonlinear tolerance rather just running a fixed number of fixed point iterations and see if there's a difference between the segregated vs block momentum solve in terms of nonlinear iterations required--maybe on an atmospheric boundary layer or a channel flow case. I'm guessing you'd need a pretty tight nonlinear tolerance to see a difference between the two, but I don't know for sure.

sayerhs:
@rcknaus Ah ok. I had done this for the ABL precursor simulation with the hypre stack and observed the `NLinear Res` was match pretty well for both approaches. Would be good to do this with the Tpetra stack also.

src/TpetraSegregatedLinearSystem.C
lucbv: Yes, `Tpetra::MultiVector` computes the global norm of each vector.
By the way this is already being used in `TpetraLinearSystem` since I switched it from using `Tpetra::Vector` to `Tpetra::MultiVector`.

lucbv:
@sayerhs would you mind sending me some explanations on how to perform the ABL precursor simulation? I'm worried I don't know exactly what we are talking about there...
Also if you still have the results you obtained while comparing Hypre segregated vs. Hypre coupled it would be a good point of comparison for me.

sayerhs:
@lucbv The setup for ABL precursor simulation is the `ablNeutralEdge` regression test. For the Hypre system, I just made sure that the non-linear residuals remained the same with monolithic and segregated approaches.

lucbv:
@sayerhs before I start doing the wrong thing, one more question, after looking at the input file for `ablNeutralEdge` I notice that I can set a convergence tolerance more tightly for the LowMach system:
```
        - LowMachEOM:
            name: myLowMach
            max_iterations: 1                           <-- this can be increased to achieve convergence tolerance first
            convergence_tolerance: 1.0e-5   <-- this can be set more tightly
```
But there is also an outer loop specified here:
```
    equation_systems:
      name: theEqSys
      max_iterations: 4
```

I am assuming that we care about the inner iterations as these are the only one I can set a tolerance too?

sayerhs:
@lucbv The `convergence_tolerance` actually acts in concert with the _outer loop_ `max_iterations` specified in the `equation_systems` block ([see here](https://github.com/Exawind/nalu-wind/blob/master/src/EquationSystems.C#L785)), it doesn't actually stop the [inner iterations](https://github.com/Exawind/nalu-wind/blob/master/src/LowMachEquationSystem.C#L696) if a particular tolerance is reached.

lucbv:
@sayerhs in that case you would mind just telling me how the input file should be modified?
It would be nice to now if this work is providing the proper results.

sayerhs:
@lucbv In my tests with Hypre, I changed the `max_iterations` (outer one in the `equation_systems` block) to a large value like 10, set the tolerance on the different equation systems and ran both the segregated and unsegregated approaches, and monitored the non-linear residuals convergence history for all the equation systems, and ensured that if the non-linear iterations quit early, they did so for the same number of iterations for both approaches.

lucbv:
@sayerhs alright, I am going to start with `max_iterations=10` as recommended and with `convergence_tolerance: 1.0e-8` for the three systems: `LowMachEOM`, `Enthalpy` and `TurbKineticEnergy`, I'll report with log files and summary of results when the simulations are completed.

lucbv:
@sayerhs I did not notice any early quit of the non-linear solver and looking at the values reported in the log file, even with a `convergence_tolerance: 1.0e-5` for all equations would like not lead to the non-linear solver stopping early as the `TurbKineticEnergy` system does not seem to reach that tolerance.
On the other hand both `LowMachEOM` and `Enthalpy` converge to much tighter tolerance and in similar fashion for both unsegregated and segregated systems.

I am attaching both log files here. I could attempt to reset all `convergence_tolerance: 1.0e-5` while increasing further `max_iterations` if we feel that it would lead to more interesting results?

lucbv:
Actually uploading the log files...

[ablNeutralEdgeNLTol.log](https://github.com/Exawind/nalu-wind/files/3335928/ablNeutralEdgeNLTol.log)
[ablNeutralEdgeNLTolSeg.log](https://github.com/Exawind/nalu-wind/files/3335929/ablNeutralEdgeNLTolSeg.log)

sayerhs:
Thanks @lucbv. That is a nice speed-up you are seeing on the linear solver timings! 

Can you explain how the RHS norms are computed within Trilinos for the two approaches? The `NLinear Res` printed for the first outer iteration of the first timestep should be exactly the same for both approaches as they have are the RHS residuals computed based on initial conditions (and shouldn't be impacted by the differences in iterative solver pathways for the two runs). 

```
        MomentumEQS            4           34.5369         493647             1                      
        MomentumEQS            3           34.5398          50945             1
```

lucbv:
@sayerhs good, catch and actually there was a mistake related to @rcknaus comment.
The way the norm was computed is wrong although the calculations are correct.
Here are updated results with the proper norm calculation:
```
        MomentumEQS            4           34.5369         493647             1
        MomentumEQS            3           34.5398         552308             1
```
the two residual are different since they are computed based on the solution of the first non-linear system which has the same initial condition but the update coming from the linear solver is different.

I am attaching the updated log file for the segregated system.
[ablNeutralEdgeNLTolSeg.log](https://github.com/Exawind/nalu-wind/files/3336602/ablNeutralEdgeNLTolSeg.log)

src/TpetraSegregatedLinearSystem.C
lucbv: @rcknaus this has been modified in latest patch

src/TpetraSegregatedLinearSystem.C
```diff
@@ -0,0 +1,1832 @@
...
+  copy_tpetra_to_stk(sln_, linearSolutionField);
+  sync_field(linearSolutionField);
+
+  // computeL2 norm
+  Teuchos::Array<double> mv_norm(ownedRhs_->getNumVectors());
+  ownedRhs_->norm2(mv_norm());
+  double norm2 = 0.0;
+  for(size_t vecIdx = 0; vecIdx < ownedRhs_->getNumVectors(); ++vecIdx) {
+    norm2 += std::sqrt(mv_norm[vecIdx]*mv_norm[vecIdx]);
```
sayerhs: @lucbv Shouldn't the `std::sqrt` happen outside of the for loop? Otherwise you are just adding the `2-norms` for each component.

lucbv: @sayerhs correct!

lucbv: Thanks for catching this...

sayerhs: @lucbv With this change I have a feeling that the first `NLinear Res` for both approaches should be the same.

lucbv:
The two approaches now match on the first non-linear iteration residual norm, I guess we are not updating the residual after the linear solver has converged?
```
        MomentumEQS            4           34.5369         493647             1
        MomentumEQS            3           34.5398         493647             1
```
Thanks for catching my mistake!

lucbv:
Latest log file attached
[ablNeutralEdgeNLTolSeg.log](https://github.com/Exawind/nalu-wind/files/3336763/ablNeutralEdgeNLTolSeg.log)

sayerhs:
@lucbv That is correct, `NLinear Res` column is reporting the residual of the RHS vector (which is the residual of the equation system the way we solve it in Nalu - delta form). If you `vimdiff` the latest log file with the original non-segregated version, the non-linear convergence trends look quite similar. 

At this point, I think the algorithm has been implemented correctly. However, it would be nice to remove the code duplication by overriding `numDof()` method to indicate `3x3` block matrix structure to `Assemble*Alg` and removing duplicate graph creation routines across the two classes, before we merge this. @alanw0 Any other thoughts?

alanw0:
@sayerhs @lucbv I have a plan to pull graph-creation code into a separate class, such that graph-creation can be done once and reused by the various equation systems. But I'm not sure exactly how soon I'll get that done. If Luc's work is ready to push now, I wouldn't object to that. I don't think it would be very hard to do the graph-creation consolidation after he pushes. The code duplication would not exist for very long.

lucbv:
@sayerhs I would prefer that we merge this now.
I am planning on working on the changes outlined earlier and will probably add a variable called `numVectors_` to allow to pass correct information to `Assemble*Alg`.
The `sumInto` will indeed still need to be overloaded and the `segregate_sum_into` will still be required.

sayerhs:
@lucbv @alanw0 OK. Then the last thing we need is a regression test that exercises the segregated solver pathway in the test suite. @lucbv if you can add the `ablNeutralEdgeSeg` case we will have coverage, and also provide an example for others who want to use this in their runs.

lucbv:
@sayerhs I have added a regression test called `ablNeutralEdgeSegregated` and generated the corresponding files. Since I am able to run the regression tests up to tolerance on my machine I am expecting that it should work just fine on the nightly test machines (`rhodes`, `mac` and `eagle`).

jhux2:
@sayerhs @alanw0 Think we can merge this today?

jhux2:
:tada:

PR #264, opened 2019-06-20
Jroverf/turb kinetic energy rodi node kernel

overfelt:

PR #269, opened 2019-06-27
New GPU kernel Turb Kinetic Energy Rodi

overfelt:
Second try. Since the test is over one
element change the old non-GPU test to
be over one element so the reference
solution can be the same between the
two.

### DIFF
--- a/unit_tests/kernels/UnitTestKernelUtils.h
+++ b/unit_tests/kernels/UnitTestKernelUtils.h
@@ -470,44 +470,6 @@ class MomentumEdgeHex8Mesh : public MomentumKernelHex8Mesh
 class MomentumNodeHex8Mesh : public MomentumKernelHex8Mesh
 {};
 
-/** Test Fixture for the Ksgs Kernels
- *
- */
-class KsgsKernelHex8Mesh : public LowMachKernelHex8Mesh
-{
-public:
-  KsgsKernelHex8Mesh()
-    : LowMachKernelHex8Mesh(),
-      tke_(&meta_.declare_field<ScalarFieldType>(
-        stk::topology::NODE_RANK, "turbulent_ke")),
-      tvisc_(&meta_.declare_field<ScalarFieldType>(
-        stk::topology::NODE_RANK, "turbulent_viscosity")),
-      dudx_(&meta_.declare_field<GenericFieldType>(
-          stk::topology::NODE_RANK, "dudx"))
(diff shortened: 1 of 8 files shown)


unit_tests/node_kernels/UnitTestTurbKineticEnergyRodiNodeKernel.C
```diff
@@ -0,0 +1,74 @@
...
+  for (unsigned i=0; i < lhs.extent(0); ++i)
+     for (unsigned j=0; j < lhs.extent(1); ++j)
+        lhs_norm += lhs(i,j)*lhs(i,j);
+
+  lhs_norm = std::sqrt(lhs_norm/(lhs.extent(0)*lhs.extent(1)/3));
+  rhs_norm = std::sqrt(rhs_norm/(rhs.extent(0)/3));
+
+  const double tol = 1e-14;
+  const double lhs_gold_norm = 0.0;
+  const double rhs_gold_norm = nprocs==1 ? 0.00030133929246584567 : 0.0002760523778561557;
```
sayerhs: Given the `if condition` on line 18, this feels redundant.

unit_tests/kernels/UnitTestKernelUtils.h
```diff
@@ -557,6 +557,84 @@ class SSTKernelHex8Mesh : public LowMachKernelHex8Mesh
   ScalarFieldType* fOneBlend_{nullptr};
 };
 
+/** Test Fixture for the Turbulence Kernels
+ *
+ */
+class TurbulenceKernelHex8Mesh : public LowMachKernelHex8Mesh
```
sayerhs: The Rodi term is going to be used with `k-sgs` model, so can we just extend the `KsgsKernelHex8Mesh` instead of creating an entirely new class here?

overfelt: Uh, I'll have to look at what the KsgsKernelHex8Mesh is.......

unit_tests/node_kernels/UnitTestTurbKineticEnergyRodiNodeKernel.C
overfelt: I assumed the if condition on line 18 was temporary.  Surly we will be able to run in parallel some day?

sayerhs:
Related to #264

PR #271, opened 2019-06-28
Convert AssembleScalarEdgeSolverAlgorithm to NGP

sayerhs:


src/edge_kernels/ScalarEdgeSolverAlg.C
```diff
@@ -0,0 +1,195 @@
...
+      const DblType qCds = 0.5 * (qHatL + qHatR);
+
+      const DblType adv_flux = mdot * (pecfac * qUpw + om_pecfac * qCds);
+      smdata.rhs(0) -= adv_flux;
+      smdata.rhs(1) += adv_flux;
+
+      // Left node contribution; upwind terms
+      DblType alhsfac = 0.5 * (mdot + stk::math::abs(mdot))
+        * pecfac * alphaUpw + 0.5 * alpha * om_pecfac * mdot;
+      smdata.lhs(0, 0) += alhsfac / relaxFac;
```
alanw0: I wonder if it's worth storing alhsfac/relaxFac in a local since it gets used 3 times...

sayerhs: @alanw0 Not sure I understand, `alhsfac` gets recomputed thrice before adding to `smdata.lhs`, and it does both with and without `relaxFac` for the L/R nodes per edge.

alanw0: @sayerhs yes I guess you're right. alhsfac/relaxFac gets used 4 times, but as you point out, alhsfac gets changed a couple times. alhsfac/relaxFac is only unchanged for the last 2, maybe it's not worth trying to save a divide or two.

PR #272, opened 2019-06-28
NGP implementation of Edge BC algorithms

sayerhs:
Use MomentumSymmetry BC as an example to demonstrate conversion of edge BC
algorithms.

Add necessary KOKKOS_FUNCTION decorators to MasterElement methods.


sayerhs:
@alanw0 Yes, `FaceElem` and `Edge` helper objects need to be fixed for GPU checks. May be when the STK team gets on a Nalu sprint they can remove `#ifndef KOKKOS_ENABLE_CUDA` from all the NGP tests and make sure they all work?

alanw0:
@sayerhs Yes that does seem like a good "come up to speed" task for the stk team to do.

Issue #255, opened 2019-06-17
regression test: tolerance testing


lucbv:
This issue is meant to continue the discussion started in PR #251

lucbv:
@sayerhs @alanw0 @rcknaus 
One of the things I would like to point is the [this](https://github.com/Exawind/nalu-wind/blob/ba2d9809c1a124c136ecb1232472c4f9547b8f1d/reg_tests/CMakeLists.txt#L33) tolerance criteria.
It pretty much means that all version of the GNU compiler are supposed to lead to the exact same convergence behavior which seems pretty strict to me.
Would it be OK to have other compiler/TPL version converge up to the default tolerance for `Linux` as set [here](https://github.com/Exawind/nalu-wind/blob/ba2d9809c1a124c136ecb1232472c4f9547b8f1d/reg_tests/CMakeLists.txt#L38)?

alanw0:
@lucbv I think it's reasonable to have slightly looser tolerances for certain platforms or compilers.  The fundamental issue is that we don't want anything to change unless we understand it. i.e., we want to know if/when a code change alters the answers. If we loosen tolerances to get tests to pass, then we're just hiding stuff. As long as we maintain some tight tolerances on the nightly dashboard then we'll still be able to know if someone pushes a bad change. Perhaps ultimately we need to set up a system like Trilinos has, to do automated PR testing. That uses a controlled set of flags and libraries on certain systems, so that the results can be controlled.

lucbv:
@alanw0 I understand that the dashboard needs to have a tight tolerance for quality assurance but that can be achieved using the configure option `-D TEST_TOLERANCE=1e-N`.
On the other hand the current default for `GCC` is bit to bit equality which is pretty hard to achieve with software version deviation...
Also I have access to `clang` and `intel` compilers which I could use on my workstation to go around the tight requirement for `GCC`.
I have decided against it because it does not feel right but eventually I might resort to it to get work done.

lucbv:
@sayerhs reading [pass_fail.sh]() it seems that absolute difference between the gold file and the test file is compared to a fixed tolerance. Should not the check be performed with a relative difference?
For solutions greater than `10` there might not be enough significant digits to reach `1e-15` tolerance?

sayerhs:
@lucbv Assuming I understood your points correctly, here are my thoughts on some of the points you raised:

[...]

PR #251, opened 2019-06-12
Luc: changing sln and rhs from Vector to MultiVector

lucbv:
This is the first step to introduce the TpetraSegergatedLinearSystem.
This will allow to store the solution and rhs correctly.

This is still a work in progress, I will run some checks first before asking for this to be merged.

@sayerhs is there a specific suite of builds/tests that I need to perform before merging something in?

PR #275, opened 2019-07-03
Allow user-specified custom golds directory

sayerhs:
Adds a CMake configuration option to allow user to override the default gold norms and specify a machine-specific custom gold norms directory. 

- Automatically sets a `1.0e-15` tolerance when using this feature. 
- The custom directory should contain `test_files/*` layout similar to the one in [nalu-wind repo](https://github.com/Exawind/nalu-wind/tree/master/reg_tests/test_files). 
- The input files are still read from `nalu-wind/reg_tests/test_files` directory, only the `norm.gold` files are consulted from the custom directory. 

Usage: 

```
cmake -DNALU_GOLD_NORMS_DIR=<path>
```

### DIFF
--- a/reg_tests/CMakeLists.txt
+++ b/reg_tests/CMakeLists.txt
@@ -12,6 +12,14 @@ if(RES_LEN EQUAL 0)
  before running the tests.")
 endif()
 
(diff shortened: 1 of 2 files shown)


sayerhs:
See discussions in #251 and #255, for some motivation.

Issue #238, opened 2019-05-28
Nalu-Wind Edge Algorithms – GPU Transition


sayerhs:
Track progress of the conversion of Edge algorithms in Nalu-Wind to be capable of running on GPUs. 

## Edge Interior algorithms

**Status:** _Ready for conversion_

Basic implementation of the NGP-version of *interior edge algorithm* was demonstrated in #232, see [ContinuityEdgeSolverAlg](https://github.com/Exawind/nalu-wind/blob/master/src/edge_kernels/ContinuityEdgeSolverAlg.C) 

~Outstanding issues~: Momentum and Scalar AdvDiff algorithm will require non-SIMD versions of [ngp_create_peclet_function](https://github.com/Exawind/nalu-wind/blob/master/src/EquationSystem.C#L521) – should be a templated function like the [non-NGP version](https://github.com/Exawind/nalu-wind/blob/master/include/EquationSystem.h#L355). Fixed in d2157ff60d67, use 

```C++
eqSystem->ngp_create_peclet_function<double>(velocity->name())
```

- [x] `AssembleContinuityEdgeSolverAlgorithm => edge_kernels/ContinuityEdgeSolverAlg` (#232)
- [x] `AssembleMomentumEdgeSolverAlgorithm` (#267)
- [x] `AssembleScalarEdgeSolverAlgorithm` (#271)
- [x] `AssembleWallDistEdgeSolverAlgorithm` (#241)
- [x] `AssembleNodalGradEdgeAlgorithm`
- [x] `AssembleNodalGradUEdgeAlgorithm`
- [ ] `AssembleScalarEdgeDiffSolverAlgorithm`
- [ ] `AssembleScalarEigenEdgeSolverAlgorithm`

## Edge Boundary algorithms

**Status:** _Ready for conversion_

These algorithms will require access to `meSCS` for the element connected to the face, so will require a hybrid approach where we do use scratch views to assemble the element data **and** be templated on element traits. 

- [x] `AssembleMomentumEdgeSymmetrySolverAlgorithm => edge_kernels/MomentumSymmetryEdgeKernel` (#272)  
- [x] `AssembleMomentumEdgeABLWallFunctionSolverAlgorithm` (#274)
- [x] `AssembleTemperatureNormalGradientBCSolverAlgorithm` (#278)
- [x] `AssembleScalarEdgeOpenSolverAlgorithm` (#306)
- [x] `AssembleContinuityEdgeOpenSolverAlgorithm` (#417)
[...]

PR #287, opened 2019-07-08
NGP looping abstractions

sayerhs:
Introduce wrapper functions that accept a C++ lambda or functor to abstract out
the looping process over buckets, and the handling of ScratchViews for element,
face/element based looping constructs. These constructs will be used to replace
current algorithms that deal with field modification so that they can occur on
device instead of host.

- NgpTypes: Introduce data structures to ease handling NGP data. In preparation
for replacing logic within the code that currently use MetaData/Bulkdata with
ngp::Mesh and ngp::Field<T>, start introducing some convenience types. Replace
handling of ngp data in Realm with the one using wrapper.

- Introduce convenience wrappers to update NGP field data with 
`stk::simd::Double` right hand sides.

- Check looping strategy and field modification using mass flow rate computation in unit tests.

- Convert massFlowRate calculator in unit tests to use NGP loops

### DIFF
--- a/include/ngp_utils/NgpFieldOps.h
+++ b/include/ngp_utils/NgpFieldOps.h
@@ -220,11 +220,13 @@ struct ElemFieldOp
     {
       const auto& fld = obj_.ngpField_;
       const auto* einfo = obj_.edata_.elemInfo;
+
+      // No atomic_add here as only one element active per thread
 #ifdef STK_SIMD_NONE
-      Kokkos::atomic_add(fld.get(einfo[0].meshIdx, ic), stk::simd::get_data(val, 0));
+      fld.get(einfo[0].meshIdx, ic) += stk::simd::get_data(val, 0);
 #else
       for (int is=0; is < obj_.edata_.numSimdElems; ++is) {
-        Kokkos::atomic_add(fld.get(einfo[is].meshIdx, ic), stk::simd::get_data(val, is));
+        fld.get(einfo[is].meshIdx, ic) += stk::simd::get_data(val, is);
       }
 #endif
     }
@@ -249,10 +251,10 @@ struct ElemFieldOp
       const auto& fld = obj_.ngpField_;
       const auto* einfo = obj_.edata_.elemInfo;
 #ifdef STK_SIMD_NONE
-      Kokkos::atomic_add(fld.get(einfo[0].meshIdx, ic), val);
+      fld.get(einfo[0].meshIdx, ic) += val;
 #else
       for (int is=0; is < obj_.edata_.numSimdElems; ++is) {
-        Kokkos::atomic_add(fld.get(einfo[is].meshIdx, ic), val);
+        fld.get(einfo[is].meshIdx, ic) += val;
       }
 #endif
     }
--- a/include/ngp_utils/NgpFieldOps.h
+++ b/include/ngp_utils/NgpFieldOps.h
@@ -9,7 +9,10 @@
 #define NGPFIELDOPS_H
 
 /** \file
- *  \brief Helper objects for field updates when using SIMD datatypes
+ *  \brief Field update utilities within element loops with SIMD data.
+ *
+ *  NgpFieldOps provides two utility classes that deal with updates of nodal and
+ *  element fields from within a SIMD-ized Kokkos::parallel_for loop.
  */
 
 #include "ngp_utils/NgpTypes.h"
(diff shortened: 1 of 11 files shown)


sayerhs:
@alanw0 @overfelt This is a preliminary design to pull out the logic in `Assemble[Edge/Elem]SolverAlgorithm::run_algorithm` into standalone functions that can be used elsewhere in the code. Now that the first round of #238 is mostly complete, I am starting to look at other operations we will need to do on device (e.g., update mdot, recompute nodal gradients on scalar and vector fields, etc.) within each timestep beyond just solving the linear system. There is a bit of boilerplate that makes this request seem big, so it might be easier to start from the unit tests [here](https://github.com/sayerhs/nalu-wind/blob/1f8a36b745f7ccdc8633679fcd0911b34bb490ac/unit_tests/ngp_kernels/UnitTestNgpLoops.C#L337-L371) and  [here](https://github.com/sayerhs/nalu-wind/blob/1f8a36b745f7ccdc8633679fcd0911b34bb490ac/unit_tests/ngp_kernels/UnitTestNgpLoops.C#L251-L272).

I am not crazy about the field operator, but I couldn't think of a better way to handle field updates when the RHS is a SIMD datatype and we need to broadcast to different elements. I thought about extending `ElemDataRequests` to allow for registering fields that should be updated and then doing it outside (like it is done for RHS/LHS), but that meant changing a lot of code (`ElemDataRequests`, `ElemDataRequestsGPU`, etc.). I feel we have painted ourselves into a corner since all the device-capable MasterElement methods are closely tied to `DoubleType` that it would take a lot of work to enable GPU execution of `double` methods. I am not sure what is the impact on performance for gathering and SIMD assumptions on GPU, but the `Kokkos::View` of gathered element data make the actual loop-body concise and easier to read/implement. 

@alanw0 In the current looping abstraction, I opted to use `MeshIndex` instead of `FastMeshIndex`, I am curious if you think I should switch that? Since the `Kokkos::parallel_for` already provides bucket and ordinal, it seemed like an extra hop to generate FastMeshIndex.

include/ngp_utils/NgpLoopUtils.h
```diff
@@ -0,0 +1,401 @@
...
+ *  element and side/face ranks with the same function.
+ *
+ *  @param mesh A STK NGP mesh instance
+ *  @param sel  STK mesh selector to choose buckets for looping
+ *  @param algorithm A functor that will be executed for each entity
+ */
+template<typename Mesh, typename AlgFunctor>
+void run_elem_algorithm(
+  Mesh& mesh,
+  const stk::topology::rank_t rank,
```
alanw0: At first I was perplexed by having 'run_elem_algorithm' take a rank. But I know AssembleElemSolverAlgorithm does get used for non-elem-rank entities. Maybe I would call this 'run_entity_algorithm' or just 'run_algorithm' but it's not a big deal either way.

include/ngp_utils/NgpLoopUtils.h
```diff
@@ -0,0 +1,401 @@
...
+  const AlgFunctor algorithm)
+{
+  using Traits    = NGPMeshTraits<Mesh>;
+  using MeshIndex = typename Traits::MeshIndex;
+
+  run_entity_algorithm(
+    mesh, rank, sel,
+    KOKKOS_LAMBDA(MeshIndex& meshIdx) {
+      algorithm(
+        EntityInfo<Mesh>{meshIdx, (*meshIdx.bucket)[meshIdx.bucketOrd],
```
alanw0: I agree that MeshIndex is the right thing to use here. The stk team has a task (which we haven't gotten to yet) to clean up the mesh-indexes. We have at least 3 of them. The one you're using here has a bucket-pointer, which means it can't be copied from host to device because the pointer would be bad, but if it's created and used within the device, like you're doing here, then I think it is the best one to use.

include/ngp_utils/NgpLoopUtils.h
sayerhs: @alanw0 I already have a `run_entity_algorithm`, this is a specialization that passes the EntityInfo object. We loop over faces and deal with them as 2D elements, so hence this weird name/issue. I'll think about this a bit more.

sayerhs: @alanw0 I already have a `run_entity_algorithm`, this is a specialization that passes the EntityInfo object. We loop over faces and deal with them as 2D elements, so hence this weird name/issue. I'll think about this a bit more.

include/ngp_utils/NgpLoopUtils.h
```diff
@@ -0,0 +1,401 @@
...
+  NGP_ThrowRequire(npe != 0);
+  return npe;
+}
+
+/** Get the nodes per face from the MasterElement registered
+ *
+ *  We need a convenience function because the ME instance is created on the device.
+ */
+template<typename DataReqType>
+inline int nodes_per_face(const DataReqType& dataReqIn)
```
alanw0: @sayerhs @overfelt I think we need one of these for num_integration_points to fix the issue at line 56 in AssembleFaceElemSolverAlgorithm.h

sayerhs:
@alanw0 @overfelt Can you take a look at 3f58f3a that I added to this request since @alanw0 looked at it? I added a little bit of code that eliminates the need to use `ip_set` `ip_add` etc. The syntax now feels more natural, but wonder if there are any pitfalls to this.

include/ngp_utils/NgpFieldOps.h
```diff
@@ -18,163 +21,309 @@
...
+#endif
+    }
+
+    KOKKOS_INLINE_FUNCTION
+    void operator+= (const DoubleType& val) const
+    {
+      const auto& fld = obj_.ngpField_;
+      const auto* einfo = obj_.edata_.elemInfo;
+#ifdef STK_SIMD_NONE
+      Kokkos::atomic_add(fld.get(einfo[0].meshIdx, ic), stk::simd::get_data(val, 0));
```
alanw0: It seems like we shouldn't need atomic-add for element fields. We definitely do for nodal fields since nodes can be shared by different elements which are being worked on by different threads at the same time. But for element fields our current parallelism doesn't have two threads working on the same element at the same time.

sayerhs: @alanw0 That is a great point. I will update the code to reflect our current looping pattern/strategy.

sayerhs:
@alanw0 Fixed the element update in f7612af. Let me know what you think.

alanw0:
@sayerhs it looks good to me.

overfelt:
I like the operator approach and can think of no reason not to implement it this way.

PR #288, opened 2019-07-09
Fix Ksgs unit test to handle dual nodal volume with 2 states

jrood-nrel:
I should have caught this before my other pull requests. Making this a pull request to be sure this is all I need to do to fix it. I ran all the unit tests now and they pass.


sayerhs:
@jrood-nrel `KsgsKerelHex8Mesh` inherits from `TestKernelHex8Mesh` and  now the `dnvField_` is shadowing the `dualNodalVolume_` field in the original mesh. I think a cleanup action should be to provide all geometry based fields in `TestKernelHex8Mesh` and remove them from the `EquationSystem` specific classes.

Issue #286, opened 2019-07-07
FaceElemSolver run_algorithm behavior


sayerhs:
I am curious about the logic for creation of `ScratchMeInfo` in`AssembleFaceElemSolverAlgorithm::run_algorithm` on GPUs here

https://github.com/Exawind/nalu-wind/blob/da5aa392e078e0705de5cede227b96fdd537b064/include/AssembleFaceElemSolverAlgorithm.h#L48-L59

This snippet of code is being executed in host space and the `MasterElement*` pointers are created on device for _converted_ BC kernels, for example, 

https://github.com/Exawind/nalu-wind/blob/da5aa392e078e0705de5cede227b96fdd537b064/src/edge_kernels/MomentumSymmetryEdgeKernel.C#L39-L42

So my question is, how is line 56 working in the first snippet? The code should crash when executed on host? Yet, CDash indicates that the unit tests work just fine.

alanw0:
I agree it looks like meFC should be a device pointer, so line 56 should seg-fault when attempting to call the 'num_integration_points()' method on host. I'll try to dig into it and see what's actually happening.

overfelt:
In my GPU build the code runs because the Kokkos::DefaultExecutionSpace::memory_space comes up Kokkos::CudaUVMSpace so the GPU allocation of the master elements works fine on the CPU.  Where is this set?  I thought something was set to avoid UVM allocations.

alanw0:
That makes sense. I just added some print statements and saw that the pointer value of the master-element created by create_device_expression was the same pointer that was later used in AssembleFaceElemSolverAlgorithm. A 'MemSpace' value is set in KokkosInterface.h. I don't know about DefaultExecutionSpace::memory_space… That maybe influenced by a cmake configuration parameter for the trilinos build?

sayerhs:
Argh! I should've guessed about the UVM. It is basically required by trilinos Tpetra. So there's no way to disable that I think, unless we can ask Mark if things have changed.

alanw0:
I'll ask Mark about that. He recently told me that tpetra still needs UVM, but I don't know if Kokkos::DefaultExecutionSpace::memory_space has to be UVM because of that.

overfelt:
So if tpetra depends on Kokkos::DefaultExecutionSpace::memory_space being Kokkos::CudaUVMSpace then we would need to define Nalu::DefaultExecutionSpace that would be Kokkos::CudaSpace on the GPU?

[...]

PR #293, opened 2019-07-14
Fix calls to device instances of MasterElements from within host code

sayerhs:
This pull request addresses the issues discussed in #286. The commits create query methods to device ME instances (within `ElemDataRequests` or `ElemDataRequestsGPU`) from host code, and replaces the necessary calls in `ScratchViews.C` and `AssembleFaceElemSolverAlgorithm.h` with these query methods. 

Other changes:
- Refactor `ElemDataRequests.h` and move the method implementations to a C++ source file
- Refactor NGP loop utilities to use the new ME query methods and remove unused code.

### DIFF
--- a/include/AssembleFaceElemSolverAlgorithm.h
+++ b/include/AssembleFaceElemSolverAlgorithm.h
@@ -16,6 +16,7 @@
 #include <SimdInterface.h>
 #include <SharedMemData.h>
 #include <CopyAndInterleave.h>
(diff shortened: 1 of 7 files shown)


sayerhs:
@alanw0 One additional thing we need to address before we can replace `run_algorithm` is the handling of LHS and RHS from within the loop-abstractions. I was hoping we can discuss that when I am at SNL this week.

alanw0:
ok sounds good.

PR #298, opened 2019-07-18
Switch the default GPU memory space from CudaUVMSpace to CudaSpace

sayerhs:
This pull request implements the necessary changes to switch the default memory space to CudaSpace

- Introduce multiple `MemSpace` definitions so that Nalu can allocate to `CudaSpace` while `TpetraLinearSystem` can continue to use `CudaUVMSpace`
- Fix `ElemDataRequestsGPU` and `ScratchViews` so that they access the correct `View` in host/device

### DIFF
--- a/include/ElemDataRequestsGPU.h
+++ b/include/ElemDataRequestsGPU.h
@@ -46,6 +46,28 @@ struct FieldInfoNGP {
   unsigned scalarsDim2;
 };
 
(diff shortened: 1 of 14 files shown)


sayerhs:
@alanw0 @overfelt Can you guys run it on ascicgpu machines also, so that we are absolutely sure we didn't overlook something before we merge?

PR #305, opened 2019-08-04
Fix AssembleEdgeSolverAlgorithm to use coeff-applier...

alanw0:
...so that the edge-based kernels can sum contributions into the
linear system on the GPU.
Verified that the unit-test in UnitTestContinuityAdvEdge works
on GPU with the ifdefs removed so that the gold-value comparisons
happen now.

### DIFF
--- a/include/AssembleEdgeSolverAlgorithm.h
+++ b/include/AssembleEdgeSolverAlgorithm.h
@@ -14,6 +14,8 @@
 #include "Realm.h"
 #include "ScratchViews.h"
 #include "SharedMemData.h"
+#include "EquationSystem.h"
+#include "LinearSystem.h"
 
 namespace stk {
 namespace mesh {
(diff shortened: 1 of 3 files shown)


alanw0:
Strangely I'm getting some small diffs in a couple of the edge nightly tests in my gcc (non cuda) build. I'm surprised because I don't see how these changes can affect the non-cuda tests... Maybe it's because I haven't correctly updated my mesh-files submodule since Shreyas made some changes recently.

sayerhs:
@alanw0 do you mean diffs in regression tests? The changes to mesh submodule was a revert of the erroneous push from Chris. Overall there should be no changes in the mesh repo

alanw0:
@sayerhs yes regression tests. the unit-tests are fine. I'll double-check that I've got everything updated.

alanw0:
@sayerhs in a clean master-branch project I'm getting these diffs for a gcc 7.2 build. (I'm not sure exactly how old my trilinos is, but shouldn't be too old)
```
FAIL: ablNeutralEdge..........................    45.9347s 5.7383e-11 2.5392e-08
FAIL: dgNonConformalFluidsEdge................    22.9624s 2.5700e-15 4.1576e-14
FAIL: dgNonConformalFluids....................    26.6742s 4.6200e-15 6.4358e-14
```

jrood-nrel:
Thanks for doing this! Saves me some confusion with the ifdefs.

PR #306, opened 2019-08-06
Converting ScalarEdgeOpenSolverAlg kernel to run on GPU

jvo1012:
Also added unit test of old kernel to get gold results

### DIFF
--- a/include/ElemDataRequests.h
+++ b/include/ElemDataRequests.h
@@ -138,16 +138,6 @@ public:
     add_face_field(*meta_.get_fields()[field], scalarsPerFace);
   }
 
-  inline void add_ip_field(unsigned field, unsigned scalarsPerNode)
-  {
-    add_ip_field(*meta_.get_fields()[field], scalarsPerNode);
-  }
-
-  inline void add_ip_field(unsigned field, unsigned tensorDim1, unsigned tensorDim2)
-  {
-    add_ip_field(*meta_.get_fields()[field], tensorDim1, tensorDim2);
-  }
-
   inline void add_element_field(unsigned field, unsigned scalarsPerElement)
   {
     add_element_field(*meta_.get_fields()[field], scalarsPerElement);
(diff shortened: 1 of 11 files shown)


include/ElemDataRequests.h
```diff
@@ -138,6 +138,16 @@ class ElemDataRequests
     add_face_field(*meta_.get_fields()[field], scalarsPerFace);
   }
 
+  inline void add_ip_field(unsigned field, unsigned scalarsPerNode)
```
sayerhs: All other kernels (see `ScalarOpenAdvElemKernel`) used `add_face_field`, can we not reuse that method to be consistent with the rest of the code.

unit_tests/edge_kernels/UnitTestScalarOpenEdge.C
```diff
@@ -0,0 +1,73 @@
...
+{0, 0, 0, 0, 0, 0, 0, 0, },
+{0, 0, 0, 0, 0, 10, 0, 0, },
+{0, 0, 0, 0, 0, 0, 10, 0, },
+{0, 0, 0, 0, 0, 0, 0, 0, },
+};
+}
+}
+#endif
+
+TEST_F(ScalarOpenEdgeKernelHex8Mesh, NGP_scalar_edge_open_solver_alg)
```
sayerhs: There is already a `SSTKernelHex8Mesh` that provides all the fields necessary to test this, can we reuse that instead of creating this additional class just for this BC?

unit_tests/kernels/UnitTestKernelUtils.h
```diff
@@ -894,6 +894,56 @@ class MixtureFractionKernelHex8Mesh : public TestKernelHex8Mesh
   const double viscSecondary_;
 };
 
+/** Test fixture for generic scalar open edge kernels
+ *
+ *  This test fixture performs the following actions:
+ *    - Create a HEX8 mesh with one element
+ */
+class ScalarOpenEdgeKernelHex8Mesh : public TestKernelHex8Mesh
```
sayerhs: Reuse `SSTKernelHex8Mesh`?

PR #300, opened 2019-07-21
Transition terms in EquationSystems to their NGP versions

sayerhs:
This pull-request continues the effort from #277 to replace edge/node algorithms/supplemental algorithms with their NGP counterparts (see #238 for details on the kernel transition status). With this change, we have all the necessary NGP _assembly terms_ to be able to run an ABL precursor case with periodic, symmetry, and wall function terms. The final commit in this pull-request also removes the old files from the codebase. 

This changes introduced by the NGP version of the terms introduce diffs in the following unit tests

```console
FAIL: ablNeutralEdge..........................    28.9449s 4.0045e-11 6.0600e-09
FAIL: ablNeutralEdgeSegregated................    26.6025s 7.6349e-11 6.1849e-09
FAIL: ablStableElem...........................    25.2395s 1.3249e-11 7.4047e-09
FAIL: ablUnstableEdge.........................    22.7967s 5.0552e-11 8.6368e-09
FAIL: ablUnstableEdge_ra......................    13.1100s 3.7790e-12 2.4520e-12
FAIL: airfoilRANSEdge.........................     3.2003s 7.6678e+04 3.0733e+00
FAIL: airfoilRANSElem.........................     3.5980s 1.2545e+05 2.8433e+00
FAIL: elemBackStepLRSST.......................     1.3090s 1.9000e-13 2.7148e-13
FAIL: elemClosedDomain........................     1.1478s 3.4062e-09 7.3567e-07
FAIL: heatedWaterChannelEdge..................     5.6754s 4.6965e-08 1.8982e-09
FAIL: heliumPlume_rst.........................    63.1259s 3.1000e-14 2.2186e-14
FAIL: milestoneRun............................    36.4316s 1.0001e-15 7.8701e-16
FAIL: movingCylinder..........................     3.6705s 1.3120e-12 1.3114e-12
FAIL: uqSlidingMeshDG.........................   110.3636s 2.3400e-14 2.7394e-14
FAIL: waleElemXflowMixFrac3.5m................   184.9962s 6.2000e-15 1.9339e-14
```

The biggest change is in the `airfoilRANS*` tests because the old version of the SST source terms weren't scaling the LHS entries by the relaxation factor. The new version of the nodal source terms scale these terms correctly, so this diff can be accepted. I have tested this case against the baseline NREL5MW blade resolved simulations. The next big change arisess to be from the density time derivative term, but is within acceptable limits.

PR #310, opened 2019-08-08
Allow UnitTestScalarOpenAlgorithm to run with unitTest2

jrood-nrel:


sayerhs:
Pull request #300  is going to remove `AssembleScalarEdgeOpenSolverAlgorithm`, so this unit test will go away at that time also. I don't think there is need to do this change.

jrood-nrel:
This can stop that test from failing until then though.

sayerhs:
This test is not checking anything, it was created to provide gold values for another test. My vote will be to remove the test (as the base source code is also going to go away). In other tests we do this: https://github.com/Exawind/nalu-wind/blob/master/unit_tests/edge_kernels/UnitTestScalarAdvDiffEdge.C#L42

jrood-nrel:
Fair enough.

PR #315, opened 2019-08-13
Fix for NALU_ALIGN for Summit.

overfelt:
We were using KOKKOS_MEMORY_ALIGNMENT (which is 64) to align
static arrays.  On Summit this gives a warning that it
is greater than 16 and it appears that because of the warning
the alignment attribute is ignored.

This commmit will change to alignof(std::max_align_t)
if compiling for the GPU.  Not sure this will compile
everywhere.

I seem to recall that KOKKOS_MEMORY_ALIGNMENT was set to 64
because this was needed for SIMD and I think Nalu even
pushed for this a couple of years ago. So I'm not sure the
Kokkos macro can be changed.

### DIFF
--- a/include/KokkosInterface.h
+++ b/include/KokkosInterface.h
@@ -15,7 +15,7 @@
 #include <Kokkos_Core.hpp>
 
 #ifdef KOKKOS_ENABLE_CUDA
-#define NALU_ALIGNED alignof(std::max_align_t)
+#define NALU_ALIGNED alignas(sizeof(long double))
 #else
 #define NALU_ALIGNED alignas(KOKKOS_MEMORY_ALIGNMENT)
 #endif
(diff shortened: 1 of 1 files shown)


sayerhs:
@overfelt Should we be conservative and just set 

```C++
#ifdef KOKKOS_ENABLE_CUDA
#define NALU_ALIGNED alignof(sizeof(double))
#else
#define NALU_ALIGNED alignas(KOKKOS_MEMORY_ALIGNMENT)
#endif 
```

overfelt:
I think the most conservative in terms of getting the best performance is using a large value for the alignment.  I don't know that alignof(sizeof(long double)) is going to get better performance than alignof(sizeof(double)) but it will certainly not be worse.  And it will only use a bit more memory.

Maybe we can add this to the list of things to check once we get good profiling tools in place.

sayerhs:
@overfelt Sorry, I should have clarified. By conservative, I meant using something like `sizeof(X)` instead of `std::max_align_t` which might not be supported on all compiler versions we need to use. I wasn't thinking of conservative in terms of size.

PR #317, opened 2019-08-15
Remove more CUDA ifdefs from node kernels

jrood-nrel:
#316 allows me to finish 7 more node kernels. Looks like the remaining ones have more suminto issues I will see if I can handle.

### DIFF
--- a/unit_tests/node_kernels/UnitTestContinuityMassBDFNodeKernel.C
+++ b/unit_tests/node_kernels/UnitTestContinuityMassBDFNodeKernel.C
@@ -34,12 +34,10 @@ TEST_F(ContinuityKernelHex8Mesh, NGP_continuity_mass_node)
 
   helperObjs.execute();
 
-#ifndef KOKKOS_ENABLE_CUDA
   EXPECT_EQ(helperObjs.linsys->lhs_.extent(0), 8u);
   EXPECT_EQ(helperObjs.linsys->lhs_.extent(1), 8u);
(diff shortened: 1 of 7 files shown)


sayerhs:
@jrood-nrel can you provide an example node kernel that is having issues?

jrood-nrel:
Here is an example of a node kernel that still fails:
```
1: /home/jrood/exawind/my-nalu-wind/unit_tests/node_kernels/UnitTestEnthalpyABLForceNode.C:37: Failure
1:       Expected: helperObjs.linsys->numSumIntoCalls_
1:       Which is: 0
1: To be equal to: 8
1: [  FAILED  ] EnthalpyKernelHex8Mesh.NGP_abl_force (8 ms)
```

sayerhs:
Ah yes, it seems that @alanw0 decided to remove the counter and in his example removed that line from the unit test that he had converted: https://github.com/Exawind/nalu-wind/pull/305/files#diff-8e04f73617cb40e46948a977a476db26L67

jrood-nrel:
Ok, are we fine if I just remove these lines then?

sayerhs:
It would be good to get @alanw0's feedback on why he decided to take it out, before we remove `EXPECT_*` statements from unit tests.

alanw0:
Hmm, I had forgotten about that. I'm not sure. I may have removed it from a test I was fixing, and then forgot to go back to it and get that part working. I think it would require putting that member in a view and copying it back up to the host to check it at the end, but maybe there's a simpler way to do it.

PR #318, opened 2019-08-16
Fix production limiting in specific dissipation rate equation for SSTDES

rcknaus:
* The kinetic energy production term in the specific dissipation rate was always getting clipped by ten times the RANS dissipation rate instead of ten times the DES dissipation rate
* Small change.  It'll only effect clipping in one term of the equation.
* I reblessed the tests and added back the "_Input" elem.  I think the NREL machines won't show diffs, but they might.
* See https://github.com/NaluCFD/Nalu/pull/462 for the corresponding NaluCFD change


reg_tests/CTestList.cmake
```diff
@@ -23,6 +23,13 @@ function(add_test_r_rst testname np)
     file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/test_files/${testname})
 endfunction(add_test_r_rst)
 
+# Regression test with input
+function(add_test_r_inp testname np)
```
sayerhs: I like the solution in NaluCFD/Nalu#462. Is there a reason to not rename `*_Input.i` to `*_rst.i`?

src/SpecificDissipationRateEquationSystem.C
```diff
@@ -349,10 +358,18 @@ SpecificDissipationRateEquationSystem::register_interior_algorithm(
         (partTopo, *this, activeKernels, "sst",
          realm_.bulk_data(), *realm_.solutionOptions_, dataPreReqs, false);
 
+      build_topo_kernel_if_requested<SpecificDissipationRateSSTDESSrcElemKernel>
+        (partTopo, *this, activeKernels, "sst",
```
sayerhs: This should be `"sst_des"`?

sayerhs:
Thanks @rcknaus, good catch!

reg_tests/CTestList.cmake
rcknaus: Well, it's not a restart so I thought it was better to keep it named something separate.  Switching it to `_rst` will work too and it won't require the extra block in the CTestList file, but it's less clear what the test is doing.

jamelvin:
Looks good!

src/node_kernels/SDRSSTDESNodeKernel.C
```diff
@@ -0,0 +1,120 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2019 National Renewable Energy Laboratory.                  */
+/*  This software is released under the license detailed                  */
+/*  in the file, LICENSE, which is located in the top-level Nalu          */
+/*  directory structure                                                   */
+/*------------------------------------------------------------------------*/
+
+#include "node_kernels/SDRSSTDESNodeKernel.h"
```
jrood-nrel: @rcknaus I think this file is missing in this pull request. When building, I'm getting an error.

rcknaus: I forgot to include it.  I just pushed a fix to master.

PR #312, opened 2019-08-08
Fix Wale model to use the deviatoric part of the square of the velocity gradient tensor

rcknaus:
* The WALE model in nalu-wind uses the full square of velocity gradient tensor rather than just the deviatoric part like in the Nicoud and Ducros paper.  

* This will cause a bunch of tests to diff:
  * TestTurbulenceAlgorithm.turbviscwalealgorithm
  * ablUnstableEdge
  * ablUnstableEdge_ra
  * ductElemWedge
  * heatedWaterChannelEdge
  * ablStableElem
  * heatedWaterChannelElem
  * hoHelium
  *  nonIsoNonUniformEdgeOpenJet 
  * nonIsoNonUniformElemOpenJet
  * waleElemXflowMixFrac3.5m

I think it's probably best to rebless them with gold files from the NREL machine.

PR #319, opened 2019-08-16
Update gold solution for TestTurbulenceAlgorithm.turbviscwalealgorithm

jrood-nrel:
This what we're doing with this one? I pulled it from here https://my.cdash.org/testDetails.php?test=49237521&build=1689746


sayerhs:
Related to #312, @rcknaus can you check this too?

PR #321, opened 2019-08-19
Fix parallel consistency for the minimum distance clipping utility in SST

rcknaus:
* The minimum distance clipping utility was missing a parallel consistency operation
* Doesn't affect any of the tests.  It's somewhat unlikely to come up as a problem.

### DIFF
--- a/src/ShearStressTransportEquationSystem.C
+++ b/src/ShearStressTransportEquationSystem.C
@@ -24,6 +24,8 @@
 // stk_mesh/base/fem
 #include <stk_mesh/base/BulkData.hpp>
 #include <stk_mesh/base/Field.hpp>
+#include <stk_mesh/base/FieldParallel.hpp>
+
 #include <stk_mesh/base/MetaData.hpp>
 
 // stk_io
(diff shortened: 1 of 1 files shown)


rcknaus:
While the exposed boundary faces may be (always) locally owned, the nodes on the exposed faces might not be.  We don't need the sync if we require that the cells across the periodic interface are also periodic, which is a good property to maintain but we definitely don't check for it---just that the cell vertices line up at the periodic interface.

sayerhs:
I agree this is a good thing to ensure consistency, and that it should be merged. 

Just trying to understand the situations where this could be a problem. We ensure vertices line up, so `nodeL` and `nodeR` both lie on the periodic face (as they would be normal to the wall face), and `exposed_area_vector` is computed globally, so the computation at the mirror node should the same as the original node?

rcknaus:
The exposed area vector is computed local to the face, so if the orientation of the area vector across the periodic interface are different, the result will be different depending on the parallel decomposition.  We just require the nodes to be periodic, so the exposed faces can have different orientations.

PR #329, opened 2019-08-29
Velocity initial condition with wake and induction using mom. theory

gantech:
Create an initial condition for a turbine simulation with wake and induction calculated using momentum theory. This should help speed up any simulations of turbines. The model accounts for wake expansion and is designed to taper off after 8 diameters over a distance of 1D. Tested to work with actuator line simulations.

velprofile_mom_wake_induction

```
      - user_function: ic_1
        target_name: fluid
        user_function_name:
         velocity: wind_energy_mom_wake_induction
        user_function_parameters:
          velocity:
            - 0.0  # Turbine location - x
            - 0.0  # Turbine location - y
            - 0.0  # Turbine location - z
            - 0.43  # Thrust coefficient
            - 126.0  # Turbine diameter
            - 8.0  # U infinity - x
            - 0.0  # U infinity - y
            - 0.0  # U infinity - z

### DIFF
--- a/include/user_functions/WindEnergyMomWakeInductionAuxFunction.h
+++ b/include/user_functions/WindEnergyMomWakeInductionAuxFunction.h
@@ -1,0 +1,75 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2019 National Renewable Energy Laboratory.                  */
+/*  This software is released under the license detailed                  */
+/*  in the file, LICENSE, which is located in the top-level Nalu          */
+/*  directory structure                                                   */
+/*------------------------------------------------------------------------*/
+
+#ifndef WINDENERGYMOMWAKEINDUCTIONAUXFUNCTION_H
+#define WINDENERGYMOMWAKEINDUCTIONAUXFUNCTION_H
+
+#include "AuxFunction.h"
+#include <vector>
+#include <array>
+
+namespace sierra{
+namespace nalu{
+
+
+
+/** Create power law velocity profile aux function for wind energy applications
+ *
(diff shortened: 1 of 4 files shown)


tonyinme:
Hey @gantech, this is a really good idea and it will be very helpful.
The implementation looks good.

I am not sure if this fits better within nalu-wind or nalu-wind-utils.
Maybe @sayerhs can comment on that?

I have only 2 minor comments/suggestions:
1. Can you add comments to describe the formulations in /WindEnergyMomWakeInductionAuxFunction.C?
Also, as part of the comments can you add reference to the paper/book where you got the formulas from (I think this was Emmanuel's book)?

2. It would be helpful to have the upstream and downstream extents as a parameter to the input. Something similar to what nalu-wind-utils has for the refinements. Instead of manually specifying in the formula the 16R, it would be good to have that as an input with a default value of 16R. This would be helpful in case you have a smaller domain and don't want the wake to reach the boundary.

jrood-nrel (approved):

sayerhs:
@tonyinme @gantech has referenced Emmanuel's paper and the relevant equations in the doxygen-style class documentation here: https://github.com/Exawind/nalu-wind/pull/329/files#diff-60774d98aedf120a9beb0b857a88c563R22

tonyinme (approved):
Looks good!

gantech:
Based on further analysis using actuator line simulations, the initial condition seems to make no difference to the time of convergence of the loads. So I say we dump this initial condition.

Compare_RotTorq

Compare_Wind1VelX

PR #330, opened 2019-08-29
2-MPI-rank test of tpetra assembly, loadComplete on GPU

alanw0:
This will verify the first item in issue 301.


alanw0:
Oversight or lack of clarity on my part: this is a test to verify loadComplete on GPU, sets up a mesh on 2 MPI ranks and adds the infrastructure to check the resulting sparse matrix on both procs, but is not yet actually calling loadComplete in the helperObjects… That's next. if you want to wait on the approval until I add that commit, that's ok.

alanw0:
OK, the missing commit is now in place, TpetraHelperObjectsEdge is now calling loadComplete.

Issue #301, opened 2019-07-21
NGP transition of TpetraLinearSystem


sayerhs:
`TpetraLinearSystem::sumInto` has been transitioned to NGP in #234. The following might need to be transitioned before a linear system solve can be performed on a GPU and the field updates are handled properly. Since Tpetra uses UVM space, the update of `deltaSolution` and the remaining fields will happen in the STK mesh space without modifications. However, the solution will need to be pushed down to device before the next solve. 

- [x] [TpetraLinearSystem::loadComplete](https://github.com/Exawind/nalu-wind/blob/master/src/TpetraLinearSystem.C#L1475)
- [x] [TpetraLinearSystem::copy_tpetra_to_stk](https://github.com/Exawind/nalu-wind/blob/master/src/TpetraLinearSystem.C#L1834) 
- [x] [LinearSystem::sync_field](https://github.com/Exawind/nalu-wind/blob/master/src/LinearSystem.C#L120)
- [x] [EquationSystem::assemble_and_solve](https://github.com/Exawind/nalu-wind/blob/master/src/EquationSystem.C#L299) - `periodic_delta_solution_update`
- [x] [field_axpby](https://github.com/Exawind/nalu-wind/blob/master/src/LowMachEquationSystem.C#L712)

sayerhs:
`sync_field` fixed in #420

PR #332, opened 2019-09-01
Convert field updates to NGP versions

sayerhs:
Synchronize fields to device after a linear system solve. Addresses the
`field_axpby` part in #301

### DIFF
--- a/include/EquationSystem.h
+++ b/include/EquationSystem.h
@@ -236,6 +236,17 @@ public:
     get_required(node, "convergence_tolerance", convergenceTolerance_);
   }
 
+  /** Update field with the delta solution from linear solve
+   *
+   *  This method updates the solution from the linear system solve on host
+   *  (within BulkData) and then synchornizes the field on the device.
+   */
+  virtual void solution_update(
+    const double delta_frac,
+    const stk::mesh::FieldBase& delta,
+    const double field_frac,
+    stk::mesh::FieldBase& field);
+
   Simulation *root();
   EquationSystems *parent();
 
--- a/include/ngp_utils/NgpFieldUtils.h
+++ b/include/ngp_utils/NgpFieldUtils.h
@@ -1,0 +1,49 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2019 National Renewable Energy Laboratory.                  */
+/*  This software is released under the license detailed                  */
+/*  in the file, LICENSE, which is located in the top-level Nalu          */
+/*  directory structure                                                   */
+/*------------------------------------------------------------------------*/
+
+#ifndef NGPFIELDUTILS_H
+#define NGPFIELDUTILS_H
+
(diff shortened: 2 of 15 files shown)


sayerhs:
@alanw0 This pull request just creates a wrapper to ensure that the updates to the field by `field_axpby` is pushed down to the device for use after a linear system solve. For the september GPU demonstration target, my assumption is that `copy_tpetra_to_stk` 

https://github.com/Exawind/nalu-wind/blob/0e87d37630507f4ea8ce5e1f0078b6af5d4323c3/src/TpetraLinearSystem.C#L1833

will use Tpetra's UVM to copy the solution vector on host. If we go this route, then all the MPI send/recv activity that happens in the subsequent calls listed in #301 will be taken care of. Since periodic involves some ghosting (although it doesn't change), it allows us to get to a working prototype while STK team is implementing the on-device send/recv for `ngpField`. When that happens it is just a matter of replacing the implementation of `solution_update` instead of modifying code in all the `solve_and_update` implementations.

sayerhs:
@alanw0 No, I haven't touched the other items in #301 as I figured you were already dealing with those. For `copy_tpetra_to_stk` we are just making sure that the device solution can be updated on `BulkData` fields correct? Or are you planning to move the logic in there to device? I suspect since we have to do MPI communications, it is probably better to leave it as it is and just ensure that the solution is transferred after a solve. Thoughts?

alanw0:
@sayerhs currently the (simple) parallel_sum for ngp::Fields brings the data from device to host, does MPI communication, then brings the data back to device. In other words the idea is that the data is on device before and after parallel_sum. The (smart) implementation is under way, which keeps the data on device and does MPI communication with device pointers. What other MPI communication do we need to do? communicate_field_data for periodic? We'll work on the ngp version of that next. Once that is done, we might be able to keep data on device through multiple time-steps right? If so, then maybe we should make copy_tpetra_to_stk copy the data straight from the device-resident tpetra solution vector into the device-resident ngp::Field. As you mentioned, tpetra's UVM model will help us out when we reference data in the wrong place, but I think we're getting close to keeping data on device and just bringing it back to host for I/O.

sayerhs:
@alanw0 Yes, if we can get 

- `stk::mesh::copy_owned_to_shared`
- `stk::mesh::communicate_field_data`

working to synchronize `ngp::Field` then we can update the field from linear solve directly on device, and then avoid having to move things to CPU. We might still have some algorithms on Nalu side that are still running on CPU and, therefore, we might still explicitly move things in the short term.

jrood-nrel:
Do we need a matching update to Trilinos/STK for this? Has it already happened?

alanw0:
@jrood-nrel yes we'll need a separate stk-to-trilinos update for copy_owned_to_shared and communicate_field_data. We'll try to get those in soon.

sayerhs:
Will implement using NgpFieldBLAS functions introduced in #342

PR #333, opened 2019-09-02
Prototype NGP versions of field update algorithms

sayerhs:
This draft pull request creates a prototype design for NGP equivalent of the `AssembleNodalGrad{,U}{Edge,Elem}Algorithm` classes. The design should be extensible to other field update algorithms for fields such as mass flux , turbulent viscosity, CFL/Re, etc. The corresponding Boundary algorithms can be prototyped after the `NodalGradElemAlg` class design.


sayerhs:
@alanw0 Yes, building some scaffolding to get unit tests for these algs.

sayerhs:
@alanw0 @overfelt @jrood-nrel The draft pull request has been updated with unit tests. I have confirmed that it is working as the original versions. 

Currently `NodalGradAlgDriver<GradPhiType>::post_work()` still does `parallel_sum` on bulk_data objects instead of the `ngp::Field` as it is not part of trilinos master yet. Either way, until the `communicate_field_data` and `copy_owned_to_shared` is available, it has to come to host anyway so the rest of the logic should remain.

alanw0:
@sayerhs ok. So until we have parallel_sum, communicate_field_data and copy_owned_to_shared all on Trilinos/master for ngp-fields, we'll restrict our demonstration runs to 1 mpi rank.

PR #334, opened 2019-09-03
Add Kokkos based calls to Edge master elements.

overfelt:


alanw0:
@overfelt To be clear, I wasn't insisting on unit-tests for this pull-request, just asking the question.

overfelt:
@alanw0 Yes there is the UnitTestKokkosME.C test to compare the old master element calls with the new Kokkos View versions.  When I tried to add the Edge AlgTraits to the list of elements tested it would not compile.  These AlgTraits structures are very light for the Edge elements so did not have a few things.  Adding to AlgTraits seemed inappropriate as they do not make sense for edges.  I figured I needed to do an std::enable_if<> to avoid compiling stuff when Edges are used and at that point I gave up to work on something else.

sayerhs:
@alanw0 I am going to merge this to allow me to proceed with next round of NGP work. I have created an issue #343 to remind us that we need to add unit tests.

PR #338, opened 2019-09-05
Improve Tpetra matrix initialization performance

DrBooom:
This modifications speeds up the Tpetra branch on the 5MW problem.

Additional  code optimization will follow this base PR.

```      
         5/19 head       This PR
cores    total time      total time     % change
-----    ----------      ----------    ---------
 4096     1391.0          1323.4         4.8
 8192      807.6           797.6         1.2
16384      619             518.8        16.2
```

### DIFF
--- a/include/TpetraLinearSystemHelpers.h
+++ b/include/TpetraLinearSystemHelpers.h
@@ -74,6 +74,16 @@ void add_lengths_to_comm(const stk::mesh::BulkData&  /* bulk */,
                          const stk::mesh::EntityId* colEntityIds,
                          const int* colOwners);
 
+void add_lengths_to_comm_tpet(const stk::mesh::BulkData&  /* bulk */,
+                              TpetIDFieldType * tpetGID_label,
+                         stk::CommNeighbors& commNeighbors,
+                         int entity_a_owner,
+                         stk::mesh::EntityId entityId_a,
+                              //                         unsigned numDof,
+                         unsigned numColEntities,
+                         const stk::mesh::EntityId* colEntityIds,
+                         const int* colOwners);
+
 void communicate_remote_columns(const stk::mesh::BulkData& bulk,
                                 const std::vector<int>& neighborProcs,
                                 stk::CommNeighbors& commNeighbors,
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -877,6 +877,9 @@ Realm::setup_nodal_fields()
   hypreGlobalId_ = &(metaData_->declare_field<HypreIDFieldType>(
                        stk::topology::NODE_RANK, "hypre_global_id"));
 #endif
+  tpetGlobalId_ = &(metaData_->declare_field<TpetIDFieldType>(
+                       stk::topology::NODE_RANK, "tpet_global_id"));
+
   // register global id and rank fields on all parts
   const stk::mesh::PartVector parts = metaData_->get_parts();
   for ( size_t ipart = 0; ipart < parts.size(); ++ipart ) {
@@ -886,6 +889,8 @@ Realm::setup_nodal_fields()
 #ifdef NALU_USES_HYPRE
     stk::mesh::put_field_on_mesh(*hypreGlobalId_, *parts[ipart], nullptr);
 #endif
+    stk::mesh::put_field_on_mesh(*tpetGlobalId_, *parts[ipart], nullptr);
+    stk::mesh::field_fill(std::numeric_limits<LinSys::GlobalOrdinal>::max(), *tpetGlobalId_);
   }
 
 
--- a/src/TpetraLinearSystem.C
+++ b/src/TpetraLinearSystem.C
@@ -241,10 +241,6 @@ void TpetraLinearSystem::beginLinearSystemConstruction()
   }
 
   std::sort(owned_nodes.begin(), owned_nodes.end(), CompareEntityById(bulkData, realm_.naluGlobalId_) );
-  std::vector<stk::mesh::Entity>::iterator iter = std::unique(owned_nodes.begin(), owned_nodes.end(), CompareEntityEqualById(bulkData, realm_.naluGlobalId_));
-  owned_nodes.erase(iter, owned_nodes.end());
-
-  ThrowRequire((int)maxOwnedRowId_ == (int)(owned_nodes.size()*numDof_)); // if this is never tripped, then the above unique is useless. !!!
 
   // use the Contiguous Map constructor. 
 
@@ -261,19 +257,10 @@ void TpetraLinearSystem::beginLinearSystemConstruction()
 
   GlobalOrdinal gomin   = ownedRowsMap_->getMinGlobalIndex();
 
-  if(realm_.tpetGlobalId_ == nullptr) {
-    std::cout << " beginLinearSystemConstruction tpetGlobalID_ is null"
-              <<std::endl<<std::flush;
-  }
-  ThrowRequire(realm_.naluGlobalId_ != nullptr);
-  ThrowRequire(realm_.tpetGlobalId_ != nullptr);
-
   for(stk::mesh::Entity entity : owned_nodes) {
     const stk::mesh::EntityId entityId = *stk::mesh::field_data(*realm_.naluGlobalId_, entity);
     myLIDs_[entityId] = numDof_*localId++;
     auto *  thisgid = stk::mesh::field_data(*realm_.tpetGlobalId_, entity);
-    ThrowRequire(thisgid != nullptr);
-    ThrowRequireMsg( gomin + numDof_ * gstart != 0 , "gid == 0 in owned node assignment loop");
     auto basegid = gomin + numDof_ * gstart;
     (*thisgid) = basegid;
     for(unsigned idof=0; idof < numDof_; ++ idof) ownedGids.push_back(basegid + idof);
@@ -322,8 +309,6 @@ void TpetraLinearSystem::beginLinearSystemConstruction()
     int owner = bulkData.parallel_owner_rank(masterentity);
     auto basegid = *stk::mesh::field_data(*realm_.tpetGlobalId_, masterentity);
 
-    ThrowRequireMsg(basegid!=0, " shared node loop master tpet gid ");
-    
     if(entity != masterentity) 
       *stk::mesh::field_data(*realm_.tpetGlobalId_, entity) = basegid;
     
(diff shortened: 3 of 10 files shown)


src/TpetraLinearSystem.C
```diff
@@ -245,17 +244,62 @@ void TpetraLinearSystem::beginLinearSystemConstruction()
   std::vector<stk::mesh::Entity>::iterator iter = std::unique(owned_nodes.begin(), owned_nodes.end(), CompareEntityEqualById(bulkData, realm_.naluGlobalId_));
   owned_nodes.erase(iter, owned_nodes.end());
 
+  ThrowRequire((int)maxOwnedRowId_ == (int)(owned_nodes.size()*numDof_)); // if this is never tripped, then the above unique is useless. !!!
+
```
alanw0: So is it useless or not? If it's useless, remove it. If not useless, remove this comment.

include/TpetraLinearSystemHelpers.h
```diff
@@ -74,6 +74,16 @@ void add_lengths_to_comm(const stk::mesh::BulkData&  /* bulk */,
                          const stk::mesh::EntityId* colEntityIds,
                          const int* colOwners);
 
+void add_lengths_to_comm_tpet(const stk::mesh::BulkData&  /* bulk */,
+                              TpetIDFieldType * tpetGID_label,
```
alanw0: Why the commented-out arguments?

src/Realm.C
```diff
@@ -886,6 +889,8 @@ Realm::setup_nodal_fields()
 #ifdef NALU_USES_HYPRE
     stk::mesh::put_field_on_mesh(*hypreGlobalId_, *parts[ipart], nullptr);
 #endif
+    stk::mesh::put_field_on_mesh(*tpetGlobalId_, *parts[ipart], nullptr);
+    stk::mesh::field_fill(std::numeric_limits<LinSys::GlobalOrdinal>::max(), *tpetGlobalId_);
```
alanw0: We'll have to do some follow-up work to avoid having multiple id fields. We've talked about using the same id field for hypre and tpetra.

src/TpetraLinearSystem.C
```diff
@@ -798,16 +852,41 @@ void TpetraLinearSystem::fill_entity_to_row_LID_mapping()
...
+                gid =  * stk::mesh::field_data(*realm_.tpetGlobalId_, node);
+            }
+            else
+              gid =  * stk::mesh::field_data(*realm_.tpetGlobalId_, node);
+
+            // GlobalOrdinal gid = tpetGIds[i];
+            
+            if(gid == 0 || gid == -1 || gid == std::numeric_limits<LinSys::GlobalOrdinal>::max() ) {
+              // This is needed because of a very small chance a ghost node doesn't have a master
+              // with a valid gid. ctests shows that entityToColLID_ is never accessed TpetraLinearSystem
```
alanw0: A very small chance? Can it happen or not?

src/TpetraLinearSystem.C
```diff
@@ -245,17 +244,62 @@ void TpetraLinearSystem::beginLinearSystemConstruction()
...
+
   for(stk::mesh::Entity entity : owned_nodes) {
     const stk::mesh::EntityId entityId = *stk::mesh::field_data(*realm_.naluGlobalId_, entity);
     myLIDs_[entityId] = numDof_*localId++;
-    for(unsigned idof=0; idof < numDof_; ++ idof) {
-      const GlobalOrdinal gid = GID_(entityId, numDof_, idof);
-      ownedGids.push_back(gid);
-    }
+    auto *  thisgid = stk::mesh::field_data(*realm_.tpetGlobalId_, entity);
+    ThrowRequire(thisgid != nullptr);
```
sayerhs: Is this throw really necessary? Also should this and the next one be `ThrowAssert`? @alanw0  comments?

src/TpetraLinearSystem.C
```diff
@@ -245,17 +244,62 @@ void TpetraLinearSystem::beginLinearSystemConstruction()
...
+  int gstart = 0;
+
+  GlobalOrdinal gomin   = ownedRowsMap_->getMinGlobalIndex();
+
+  if(realm_.tpetGlobalId_ == nullptr) {
+    std::cout << " beginLinearSystemConstruction tpetGlobalID_ is null"
+              <<std::endl<<std::flush;
+  }
+  ThrowRequire(realm_.naluGlobalId_ != nullptr);
+  ThrowRequire(realm_.tpetGlobalId_ != nullptr);
```
sayerhs: Prefer converting this to `ThrowRequireMsg` and eliminating lines 264-267.

src/TpetraLinearSystem.C
```diff
@@ -245,17 +244,62 @@ void TpetraLinearSystem::beginLinearSystemConstruction()
...
+  LocalOrdinal localId = 0;
+  int gstart = 0;
+
+  GlobalOrdinal gomin   = ownedRowsMap_->getMinGlobalIndex();
+
+  if(realm_.tpetGlobalId_ == nullptr) {
+    std::cout << " beginLinearSystemConstruction tpetGlobalID_ is null"
+              <<std::endl<<std::flush;
+  }
+  ThrowRequire(realm_.naluGlobalId_ != nullptr);
```
sayerhs: Is this throw really necessary?

src/TpetraLinearSystem.C
sayerhs: What is the situation where a ghosted node is not a `master` i.e., `(naluGlobalId_[node] != bulk.identifier(node)`?

sayerhs:
@alanw0 has already hit on the main points regarding this commit, and I agree with all of them. 

Looking at the code, I don't follow why a completely different field is necessary to track Tpetra Global IDs, and why we can't just rename `hypreGlobalID_` to `linSysGlobalID_` or something neutral and use it within Trilinos and Hypre? If the concern is the integer type `LinSys::GlobalOrdinal`, then we can change `hypreGlobalID_` to use that and HypreLinearSystem interface can handle it just fine. The current linear system global ID numbering has been extensively tested for all use-cases, so duplicating the logic without full test coverage feels unnecessary. Can we add some comments explaining the design choices so that we have a record of the differences for this pull-request?

src/TpetraLinearSystem.C
DrBooom: I'll remove it.

src/Realm.C
DrBooom: I will remove the useless std::unique as part of this PR, the other tasks are < 2 days of coding, assuming a working build, more like 1 day if all goes well. 
Moving tpetGlobalID and hypreGlobalID into a new unified type is tedious typing,but should only be a day to get to a PR.

src/TpetraLinearSystem.C
DrBooom: This happens only in ctest::unit_test1. In the case that a ghosted node does not have a master with a valid tpetGlobalID, it is never accessed downstream. 

This case should never happen, if my understanding of the STK documentation is correct, however unit_test1 does trigger this behavior. It may be that unit_test1 has a broken mesh set up.

I will modify this comment.

alanw0: It's puzzling that it could appear in unit_test1. Unit_test1 is the single-processor run of the unit-tests, and there is no ghosting when there is only 1 MPI rank.

DrBooom:
> @alanw0 has already hit on the main points regarding this commit, and I agree with all of them.
> 
> Looking at the code, I don't follow why a completely different field is necessary to track Tpetra Global IDs, and why we can't just rename `hypreGlobalID_` to `linSysGlobalID_` or something neutral and use it within Trilinos and Hypre? 

    I fully agree that this should get cleaned up, in a future PR.  

Can we add some comments explaining the design choices so that we have a record of the differences for this pull-request?
     I'll do that here, and if requested can paste the comments into the code:
========================================================

   This PR implements a change to take advantage of the fact that Tpetra can use internal optimizations if the GlobalID space for the Map/Graph/Matrix are contiguous and compact. 

During the assignment of GlobalId's to the owned nodes, the sorting of those nodes by naluGlobalId is needed as the solve is slowed down if unsorted owned nodes are used to build the Tpetra objects.

src/TpetraLinearSystem.C
```diff
@@ -876,13 +861,8 @@ void TpetraLinearSystem::fill_entity_to_col_LID_mapping()
...
               gid =  * stk::mesh::field_data(*realm_.tpetGlobalId_, node);
 
-            // GlobalOrdinal gid = tpetGIds[i];
-            
             if(gid == 0 || gid == -1 || gid == std::numeric_limits<LinSys::GlobalOrdinal>::max() ) {
-              // This is needed because of a very small chance a ghost node doesn't have a master
-              // with a valid gid. ctests shows that entityToColLID_ is never accessed TpetraLinearSystem
-              // at those locations: CF SumInto and insert_graph_connections.
-              // Question: Can the selector be modified to eliminate ghost nodes from this loop?
+	      // unit_test1 does produce ghost nodes that have a master that don't have a valid tpetGlobalId_
```
alanw0: I will look into this. I don't think this needs to hold up this PR.

sayerhs:
@DrBooom As @alanw0 said, the speed-ups are great, and I think that is a good argument for merging this so that we can benefit from this. It would be good to have a definite plan on how we go about merging the two linear solver global IDs. I'll let you and @alanw0 decide on that. 

Regarding the following comments: 

> This PR implements a change to take advantage of the fact that Tpetra can use internal optimizations if the GlobalID space for the Map/Graph/Matrix are contiguous and compact.

I would say `hypreGlobalID_` is contiguous and compact. Is there something missing there? 

> During the assignment of GlobalId's to the owned nodes, the sorting of those nodes by naluGlobalId is needed as the solve is slowed down if unsorted owned nodes are used to build the Tpetra objects.

`hypreGlobalID_` is sorted to ensure that the ordering is the same as naluGlobalID (see snippet below), is there something extra that the logic for `tpetraGlobalID` does that I overlooked?

https://github.com/Exawind/nalu-wind/blob/3b9098b0965d421cf97d95f7b1cc8fe70412f41c/src/Realm.C#L3532-L3569

DrBooom:
> I would say `hypreGlobalID_` is contiguous and compact. Is there something missing there?

I think you are correct, however I've not delved into anything other than the setup of hypreGlobalID in Realm. 

Is there an example problem (or can we make one) that runs the same equation in Hypre and in Tpetra, and simply compare the global ID's? If they match, then a linux replace of hypreGlobalId and tpetGlobalId to some third named type accomplishes this goal (with a few additional edits). 

> 
> > During the assignment of GlobalId's to the owned nodes, the sorting of those nodes by naluGlobalId is needed as the solve is slowed down if unsorted owned nodes are used to build the Tpetra objects.
> 
> `hypreGlobalID_` is sorted to ensure that the ordering is the same as naluGlobalID (see snippet below), is there something extra that the logic for `tpetraGlobalID` does that I overlooked?
> 
    No, the comment was based on the attempt to same come cycles by not doing the sort.

DrBooom:
Well.... I did the conversion to combine the hypre and tpet GlobalID's. It passes all ctests.

I would prefer to have the existing PR approved, then submit this change as a new PR, since it touches so many files. 

Thoughts?

jhux2:
Let's capture the next steps in a new issue.

PR #341, opened 2019-09-06
Convert TurbViscKsgsAlg to NGP

jrood-nrel:


unit_tests/ngp_algorithms/UnitTestTurbViscKsgsAlg.C
```diff
@@ -0,0 +1,63 @@
...
+#include "ngp_algorithms/TurbViscKsgsAlg.h"
+
+TEST_F(KsgsKernelHex8Mesh, NGP_turb_visc_ksgs_alg)
+{
+  // Only execute for 1 processor runs
+  if (bulk_.parallel_size() > 1) return;
+
+  fill_mesh_and_init_fields();
+
+/*
```
sayerhs: Candidate for cleanup?

src/ngp_algorithms/TurbViscKsgsAlg.C
```diff
@@ -0,0 +1,66 @@
...
+  const auto dualNodalVolume = fieldMgr.get_field<double>(dualNodalVolume_);
+  auto tvisc = fieldMgr.get_field<double>(tvisc_);
+
+  const DblType cmuEps = cmuEps_;
+  const DblType invNdim = 1.0 / meta.spatial_dimension();
+
+  nalu_ngp::run_entity_algorithm(
+    ngpMesh, stk::topology::NODE_RANK, sel,
+    KOKKOS_LAMBDA(const Traits::MeshIndex& meshIdx) {
+      const DblType filter = std::pow(dualNodalVolume.get(meshIdx, 0), invNdim);
```
sayerhs: Is there a benefit to using `std::sqrt` and `std::cbrt` here instead of generic `std::pow`? We did that in a Kernel 

https://github.com/Exawind/nalu-wind/blob/3b9098b0965d421cf97d95f7b1cc8fe70412f41c/src/kernel/TurbKineticEnergyKsgsSrcElemKernel.C#L89-L91

alanw0: Does this run on GPU? I didn't think std:: functions would work on GPU, instead we need to use the stk::math:: versions.

sayerhs: @alanw0 `stk::math::pow` in turn calls `std::pow` anyway, so I don't think that will be an issue on device. The pitfall is that if we ever switch `DblType` to `stk::simd::Double` then the code would break on `simd`.

https://github.com/trilinos/Trilinos/blob/ae1e3b063186d08c9dade332701e92257a6fe24f/packages/stk/stk_math/stk_math/StkMath.hpp#L79-L81

PR #352, opened 2019-09-10
Convert diff flux coeff and turb visc computations to NGP

sayerhs:
This pull-request implements the effective diffusive flux coefficient and turbulent viscosity computations to run on device for the Momentum equation system.


sayerhs:
This is built upon #351, so it will show an extra commit that will go away when #351 is merged.

PR #355, opened 2019-09-11
Combine hypreGlobalId and tpetGlobalId

DrBooom:
I took a chance and did a fast merge of hypreGlobalID and tpetGlobalID, and it passed ctests with minimal effort. This addresses one of the major concerns with PR #338

### DIFF
--- a/include/HypreDirectSolver.h
+++ b/include/HypreDirectSolver.h
@@ -93,28 +93,28 @@ protected:
 
   mutable HYPRE_Solver precond_;
 
-  HypreIntType (*solverCreatePtr_)(MPI_Comm, HYPRE_Solver*);
-  HypreIntType (*solverDestroyPtr_)(HYPRE_Solver);
-  HypreIntType (*solverSetupPtr_)(
+  linSysIntType (*solverCreatePtr_)(MPI_Comm, HYPRE_Solver*);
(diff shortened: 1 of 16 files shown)


include/HypreDirectSolver.h
```diff
@@ -93,28 +93,28 @@ class HypreDirectSolver: public LinearSolver
 
   mutable HYPRE_Solver precond_;
 
-  HypreIntType (*solverCreatePtr_)(MPI_Comm, HYPRE_Solver*);
-  HypreIntType (*solverDestroyPtr_)(HYPRE_Solver);
-  HypreIntType (*solverSetupPtr_)(
+  linSysIntType (*solverCreatePtr_)(MPI_Comm, HYPRE_Solver*);
```
sayerhs: We cannot replace `HypreIntType` with Tpetra's GlobalOrdinal here. None of the return types or arguments should be changed

DrBooom: I'll revise and re-test. I thought Hyper was turned on in this build, apparently not...

DrBooom:
So, it appears that LinSys uses GlobalOrdinal  long, while hypre uses BIGINT long long. 

This is going to take more rework than the time I have right now. , I'm going to withdraw this PR...

PR #362, opened 2019-09-13
Extend unit test to include some Edge Master Element checking.

overfelt:
Seems a bit complicated to use the face/edge combination scratchview in this unit test. Maybe this is not the best example of this kind of thing.


alanw0:
Yes, I'm having trouble thinking about a better way to do this kind of stuff right now...

sayerhs:
@overfelt Any reason to wait on merging this one?

overfelt:
Oh, forgot about this.  I think the unit test should still work and it does provide a tiny bit of checking for one of the master element functions.  So, yes, let's merge.

sayerhs:
Seems to be addressed in #422

PR #365, opened 2019-09-13
Cleanup periodic_field_update in preparation for ngp conversion

alanw0:
This commit makes MPI communication more efficient. There were
some extra (un-needed) calls to parallel_communicate_field. Also,
Instead of making separate calls to copy_owned_to_shared and
communicate_field_data(aura,...), it is more efficient to make a
single call to communicate_field_data(bulk, ...) because this call
includes shared and all ghostings for a single round of MPI calls.

### DIFF
--- a/include/PeriodicManager.h
+++ b/include/PeriodicManager.h
@@ -60,15 +60,15 @@ class PeriodicManager {
   // holder for master += slave; slave = master
   void apply_constraints(
     stk::mesh::FieldBase *,
-    const unsigned &sizeOfField,
-    const bool &bypassFieldCheck,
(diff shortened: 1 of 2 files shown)


alanw0:
My test-run is still in progress and I now see that I have a couple of small diffs. I'll confirm whether they are due to this commit or not, before merging.

alanw0:
ok, re-ran the diffs without my commit, they are pre-existing. It's just 3 tests, and the diffs are around 1.e-14 so Jon has done a good job of cleaning up gold files, this is much better than it was yesterday.

PR #366, opened 2019-09-13
Convert ComputeABLWallFrictionVelocity to NGP

sayerhs:
This pull-request converts ABL wall friction velocity calculations to NGP version. The original algorithm is split into two: 1. a geometry computation part (assembled wall area and normals) that is no only performed once during initialization and then only recomputed if there is mesh motion, and 2. the wall friction velocity computation that is called in place of the old algorithm.

### DIFF
--- a/src/ngp_algorithms/ABLWallFrictionVelAlg.C
+++ b/src/ngp_algorithms/ABLWallFrictionVelAlg.C
@@ -1,0 +1,310 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2019 National Renewable Energy Laboratory.                  */
+/*  This software is released under the license detailed                  */
+/*  in the file, LICENSE, which is located in the top-level Nalu          */
+/*  directory structure                                                   */
+/*------------------------------------------------------------------------*/
+
+#include "ngp_algorithms/ABLWallFrictionVelAlg.h"
+
+#include "BuildTemplates.h"
+#include "master_element/MasterElement.h"
+#include "master_element/MasterElementFactory.h"
+#include "ngp_utils/NgpLoopUtils.h"
+#include "ngp_utils/NgpFieldOps.h"
+#include "ScratchViews.h"
+#include "SolutionOptions.h"
+#include "utils/StkHelpers.h"
+#include "wind_energy/MoninObukhov.h"
+
+#include "stk_mesh/base/Field.hpp"
+
+namespace sierra {
+namespace nalu {
+
+namespace {
+
+template <typename PsiFunc>
+KOKKOS_FUNCTION
+double calc_utau(
+  const double uh,
+  const double zh,
+  const double term1,
+  const double Tflux,
+  const double Lfac,
+  const double kappa,
+  PsiFunc psi_func,
+  const double psi_fac)
+{
+  const double eps = 1.0e-8;
+  const double convTol = 1.0e-7;
+  const double perturb = 1.0e-3;
+  const int maxIters = 40;
+
+  // Return utau as epsilon if the velocity at surface is zero
+  if (stk::math::abs(uh) < eps)
+    return eps;
+
+  double utau0, utau1;
+  // Proceed with normal computations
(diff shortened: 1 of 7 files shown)


sayerhs:
Diffs introduced by this commit 

```
FAIL: ablNeutralEdge..........................    44.5935s 5.0635e-11 7.6628e-09
FAIL: ablNeutralEdgeSegregated................    40.6261s 2.3196e-11 1.8791e-09
FAIL: ablStableElem...........................    25.2016s 7.2112e-12 3.9548e-09
FAIL: ablUnstableEdge.........................    26.2378s 2.6215e-11 4.4789e-09
FAIL: ablUnstableEdge_ra......................    14.4914s 5.1577e-12 8.3990e-12
FAIL: ablUnstableEdge_rst.....................    14.9938s 3.6460e-11 2.0183e-12
```

src/ngp_algorithms/ABLWallFrictionVelAlg.C
```diff
@@ -0,0 +1,310 @@
...
+#include "stk_mesh/base/Field.hpp"
+
+namespace sierra {
+namespace nalu {
+
+namespace {
+
+template <typename PsiFunc>
+KOKKOS_FUNCTION
+double calc_utau(
```
sayerhs: @alanw0 @overfelt Can this function be converted to SIMD version? The convergence criteria seems a bit tricky.

src/ngp_algorithms/ABLWallFrictionVelAlg.C
```diff
@@ -0,0 +1,310 @@
...
+        uTangential = stk::math::sqrt(uTangential);
+
+        const DoubleType Tflux = heatFluxIp / (rhoIp * CpIp);
+        const DoubleType Lfac = stk::math::if_then_else(
+          (stk::math::abs(Tflux) < eps), Lmax,
+          (-Tref / (kappa * gravity * Tflux)));
+        const DoubleType term = stk::math::log(zh / z0);
+
+        DoubleType utau_calc;
+        for (int si = 0; si < edata.numSimdElems; ++si) {
```
sayerhs: See comment at `calc_utau` function implementation.

src/ngp_algorithms/ABLWallFrictionVelAlg.C
alanw0: At first look, I don't think so. I don't know how it would work if there can be a different number of iterations for different entries within a simd type...

overfelt: @sayerhs Shreyas, 
Would not bother with a SIMD version right now.  I think that for SIMD you just iterate until all are converged.  Still faster than iterating each SIMD entry til convergence separately.  I think the big outer if/then/else at the calling point would need to be pushed into the calc_utau function.  The Tflux==0 version might be tricky but maybe that is a rare case.

sayerhs:
@overfelt `Tflux == 0` is the neutral ABL case. That's all we have been running in Exawind for the last three years and will probably continue for the remainder 😁

sayerhs:
@overfelt how do you suggest we do the convergence tolerance check for SIMD? We still have de-interleave to check the criteria right?

overfelt:
@sayerhs Maybe I'm missing something, but I think you just keep iterating until all residuals are less than tolerance.  So "if(stk::math::abs(f1) < convTol)" becomes "if(max(stk::math::abs(f1))<convTol)".  This assumes that adding extra iterations to the already converged values does not hurt.  This looks like a Newton iteration of some kind. If it is finding a root of a function, normally it is stable in that iterating on the exact solution returns the exact solution.   In other cases the exact solution would hit a 0 divided by 0 line and can be unstable without special logic.  Ahhh, I now see the logic at line 79 to avoid division by zero but I do not see any logic to avoid epsilon divided by epsilon which can be bad.  So, yes, interleave the check criteria to capture the converged solutions once and then ignore the results of any extra iterations on those.

PR #374, opened 2019-09-18
Convert periodic_field_update for GPU execution

alanw0:
Passes all tests on CPU, checked build and execution using
periodic3dEdge on GPU. (Test still doesn't pass, but runs to
completion.)

### DIFF
--- a/include/PeriodicManager.h
+++ b/include/PeriodicManager.h
@@ -131,8 +131,10 @@ class PeriodicManager {
  public:
   // the data structures to hold master/slave information
   typedef std::pair<stk::mesh::Entity, stk::mesh::Entity> EntityPair;
+  typedef Kokkos::pair<stk::mesh::Entity, stk::mesh::Entity> KokkosEntityPair;
   typedef std::pair<stk::mesh::Selector, stk::mesh::Selector> SelectorPair;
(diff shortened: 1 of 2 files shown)


alanw0:
Note that this pull-request shouldn't be merged until nalu-wind has updated Trilinos to be newer than yesterday morning or so. (I can provide sha1 if needed.)

sayerhs:
@alanw0 I think it would be a good idea to allow for both on-host and on-device exchanges for these methods, and we could control at the call site how we want the exchanges happening. There are situations (where we will still need some things to happen on host, e.g., during initialization, transfers, overset ghosting changes) and at that point it might be better to pull the fields to host do the non-ngp algorithms on host, perform parallel communication on host fields, and then push it back to device as necessary. Can we retain the old methods and rename the new ones `ngp_periodic_field_update` and so on, so that we can explicitly transition these calls one by one.

alanw0:
@sayerhs yes that's easy and sounds reasonable. I've got a chopped up day today with sprint meetings but I'll put in a revised commit soon.

alanw0:
Ok I think this is merge-able now. Realm methods are calling the non-ngp versions of the PeriodicManager methods, can be switched to ngp versions when ready. I'll do the apply_max_field conversion next.

Issue #336, opened 2019-09-05
NGP conversion of field update algorithms


sayerhs:
This issue tracks the non-solver field updates that needs to be converted to run on device for wind-energy relevant simulations. 

For a template of edge loop conversion, see conversion of `AssembleNodalGradEdgeAlgorithm` in 

https://github.com/Exawind/nalu-wind/blob/d2d9b58929b9b201b04089dc5a296e8ccb3950ec/include/ngp_algorithms/NodalGradEdgeAlg.h#L20

Similarly the `NodalGradElemAlg` and `NodalGradBndryElemAlg` shows examples of element conversion. We will need all the boundary elem algorithms converted as they are common for both edge and elem solver options.

https://github.com/Exawind/nalu-wind/blob/d2d9b58929b9b201b04089dc5a296e8ccb3950ec/include/ngp_algorithms/NodalGradElemAlg.h#L23
https://github.com/Exawind/nalu-wind/blob/d2d9b58929b9b201b04089dc5a296e8ccb3950ec/include/ngp_algorithms/NodalGradBndryElemAlg.h#L23

Note, that the template arguments are not necessary for most Edge kernels. It was done for NodalGrad to collapse multiple classes into single class that deals with both vector and scalar fields whose projected nodal gradients are computed. 

For node loops, an example is available in #335 and is pending merge. Unit tests for these examples are in the [unit_tests/ngp_algorithms](https://github.com/Exawind/nalu-wind/tree/master/unit_tests/ngp_algorithms) directory. 

# ABL Neutral simulation

Periodic BCs, with wall functions for terrain. 

- [x] `AssembleNodalGradEdgeAlgorithm`
- [x] `AssembleNodalGradElemAlgorithm`
- [x] `AssembleNodalGradUEdgeAlgorithm`
- [x] `AssembleNodalGradUElemAlgorithm`
- [x] `AssembleNodalGradBoundaryAlgorithm`
- [x] `AssembleNodalGradUBoundaryAlgorithm`
- [x] `ComputeMdotEdgeAlgorithm`
- [x] `EffectiveDiffFluxCoeffAlgorithm`
- [x] `EnthalpyEffectiveDiffFluxCoeffAlgorithm`
- [x] `TurbViscKsgsAlgorithm`
- [x] `ComputeABLWallFrictionVelocityAlgorithm`
- [x] `ComputeTurbKineticEnergyWallFunctionAlgorithm`
- [x] `FixPressureAtNodeAlgorithm`

[...]

PR #375, opened 2019-09-20
Turbulent Kinetic Energy Wall Algorithm for ngp.

overfelt:
This passes the unit test but I'm not sure it is all correct.  The use of parallel_sum seems to be odd  for a single processor run but it was the only way I could get non-zero values out of the algorithm.

### DIFF
--- a/include/ScratchViews.h
+++ b/include/ScratchViews.h
@@ -545,7 +545,7 @@ template<typename T, typename TEAMHANDLETYPE, typename SHMEM>
 void MasterElementViews<T, TEAMHANDLETYPE, SHMEM>::fill_master_element_views_new_me(
   const ElemDataRequestsGPU::DataEnumView& dataEnums,
   SharedMemView<double**, SHMEM>* coordsView,
-  MasterElement* /* meFC */,
+  MasterElement* meFC,
   MasterElement* meSCS,
   MasterElement* meSCV,
   MasterElement* meFEM,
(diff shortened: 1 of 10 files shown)


sayerhs:
@overfelt Seems like there was some serious miscommunication here. Thsi algorithm has already been converted here: https://github.com/Exawind/nalu-wind/blob/f8fb51a40cdd27100be35e92a7184b868eebda66/src/ngp_algorithms/TKEWallFuncAlg.C#L26

overfelt:
@sayerhs, I had checked it off on #336 but I guess I was too slow in getting it converted. 

So why do we have to add the coordinates field to the faceData_ requests when they are not used?

sayerhs:
I am sorry @overfelt I think I mistakenly assumed that I had checked it off. WE should have added names to people who had checked things off. It is my mistake. I'll try to be more careful in future. Regarding `coordinates`, good catch! While I was developing, I had registered `shape_fcn` for some reason and then took it out, but forgot to take `coordinates` out when I was cleaning up the code.

overfelt:
I thought that if I checked off a box way down the list I could get ahead of you, but you are too fast.  

Another question:  Why do you have to sync to host before the parallel_sum but don't have to sync_to_device afterwards?  Is that done automatically by one of the following functions?  

And does the stk function ngp::parallel_sum() not work?  I see that it just does a sync-to-host/parallel_sum/sync-to-device but when I tried it it simply did not seem to work for me.  Maybe @alanw0 can comment on that.

jro

sayerhs:
@overfelt That is a bug in the code that I need to fix. The parallel update logic needs to change also, to incorporate Alan's latest updates. One thing we need to figure out is the best way to minimize device-host-device transfers. Right now every call to `ngp::parallel_*` method will force a sync.

alanw0:
I'm interested to hear any further details about the parallel update logic. Functions are in place to do parallel_sum, copy_owned_to_shared and communicate_field_data for ngp::Fields. As Shreyas says, we may still need to reduce the sync_to_host/sync_to_device calls, but as far as I know those functions are working.

PR #376, opened 2019-09-23
Add Cartesian Aligned Strong Symmetry BC

psakievich:
Add option to apply a symmetry bc strongly for a plane who's normal aligns with one of the principle Cartesian directions.

Bonus features:
- Gaussian jet bc added to user functions. This was used for a test case
- Kokkosify tri33D shape function calls

### DIFF
--- a/include/Enums.h
+++ b/include/Enums.h
@@ -42,7 +42,10 @@ enum AlgorithmType{
    */
   REF_PRESSURE = 15, 
  
-  TOP_ABL = 16
+  TOP_ABL = 16,
+  X_SYM_STRONG = 17,
+  Y_SYM_STRONG = 18,
+  Z_SYM_STRONG = 19
 };
 
 enum BoundaryConditionType{
--- a/include/LowMachEquationSystem.h
+++ b/include/LowMachEquationSystem.h
@@ -224,6 +224,7 @@ public:
 
   // saved of mesh parts that are not to be projected
   std::vector<stk::mesh::Part *> notProjectedPart_;
+  std::array<std::vector<stk::mesh::Part*>,3> notProjectedDir_;
 };
 
 class ContinuityEquationSystem : public EquationSystem {
--- a/include/NaluParsing.h
+++ b/include/NaluParsing.h
@@ -312,9 +312,11 @@ struct SymmetryUserData : public UserData {
     Z_DIR_STRONG
   };
   SymmetryTypes symmType_;
+  bool useProjections_;
   SymmetryUserData()
     : UserData(),
-      symmType_(SymmetryTypes::GENERAL_WEAK)
+      symmType_(SymmetryTypes::GENERAL_WEAK),
+      useProjections_(false)
   {}
 };
 
--- a/include/NaluParsing.h
+++ b/include/NaluParsing.h
@@ -305,8 +305,16 @@ struct OversetUserData : public UserData {
 };
 
 struct SymmetryUserData : public UserData {
+  enum class SymmetryTypes{
+    GENERAL_WEAK,
+    X_DIR_STRONG,
+    Y_DIR_STRONG,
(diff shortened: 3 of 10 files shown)


marchdf:
@psakievich this looks good. Quick question: can I specify multiple strong sym BC in different directions? i.e. can I do:
```
    - symmetry_boundary_condition: bc_tunnel_walls
      target_name: tunnel_walls
      symmetry_user_data:
        symmetry_type: y_direction_strong
        symmetry_type: z_direction_strong
```
Or would I have to split the `tunnel_walls` sideset into a front/back and top/bottom type of thing?

gantech:
Should we also add the parts with strong symmetry to the `not_projected_part`?

psakievich:
@marchdf You would need to split them into separate sidesets.  Each symmetry type applies a separate BC's so if you applied them two then those velocity components would be set to zero on each surface in the sideset.

psakievich:
@gantech personally I'd rather not, but we could if that is the preference.  This is a unique bc because we are only constraining one component. Omitting the projection would be more consistent with the other Dirichlet BC's, but then continuity errors can creep into the other two components. I'm not sure if they would grow or not. 

Doing the projection can create some small errors in the bc enforcement, but they will be checked every iteration by the provisional calculation.  From my testing so far the errors have been small.  For the McAlister case with CFL ~10K I was seeing errors on the order of the linear solve convergence in the normal velocity component i.e. for free stream velocity of 46 m/s the max z-velocity for on the symmetry bc was +/- 1.e-3

gantech:
Yeah..you're right. May be we should project in the other two directions and turn off projection for the wall normal component alone? I just think that strong slip should mean wall normal velocity of zero.

psakievich:
@gantech I can prototype this if you'd like before merging. If we skip the projection in one direction then the other components still won't be divergence free since we've already done all the solves, but I doubt it will impact the numerics much.  Or I can just add an option to allow turning projection on/off for these BC's.

gantech:
Thanks. It would be great to turn off the projection in the wall normal direction by default for strong symmetry boundaries in my opinion. The divergence free condition is right now enforced only through the mass flow rate on the edges/scs and not through the nodal velocity anyway. So turning off any projection shouldn't change the divergence free condition. The mass flow rate is assumed to be zero for all equation systems on any symmetry faces already. This change would simply make the nodal velocity field consistent with that assumption.

psakievich:
@gantech here you go

sayerhs:
@psakievich looking at your latest commit, it seems to disable projection for all velocity components on the symmetry part, is that correct? My understanding was that @gantech suggested that we only turn off projection for the velocity component normal to the boundary, but retain projection for the tangential components. That would be more involved and would require special treatment in `project_nodal_velocity`.

psakievich:
@sayerhs you are correct.  I started prototyping turning off the projection for the independent directions, but as I thought about it more I don't really think it will make much of a difference if we don't project the other two.  Maybe I'm missing something here but it seems like doing the projection for two out of the three is not really numerically consistent since the terms in the projection calculation are based off all three components of the provisional velocity and the pressure field has the errors in the normal direction baked in.  To turn off each one independently will require three more kokkos functions for doing the projections so it didn't seem like it was worth it for something that is not consistent. Does that make sense or am I off base here?  It's not really that big of a change to implement the single direction it just seems less numerically-consistent to me. Perhaps we should chat via slack?

psakievich:
Changes have been made.  Will merge today unless there are more comments.

sayerhs:
@psakievich Ganesh is on PTO, can this wait until he gets back and has a chance to look at it?

psakievich:
@sayerhs to document our discussion here: yes to both your questions, but I will add a warning instead of a throw since testing so far shows no difference in the results between the two solver configurations.

PR #378, opened 2019-09-24
Fix GPU compile error.

overfelt:


sayerhs:
@overfelt Why do we need those methods to be public?

overfelt:
It is unfortunate that kernel launches can not even be in protected functions.  Maybe we need to append "private_" to such functions to show they are not for general use.

sayerhs:
How about `impl_method_name` to indicate that this is an implementation detail? Follows the convention of `namespace impl { ... };` etc.

PR #381, opened 2019-09-24
Updates to get working version of ABL simulation on GPU

sayerhs:


src/Realm.C
```diff
@@ -3376,6 +3378,38 @@ void
 Realm::swap_states()
 {
   bulkData_->update_field_data_states();
+
+#ifdef KOKKOS_ENABLE_CUDA
+  if (get_time_step_count() < 2) return;
```
jrood-nrel: I'm curious about why this exists.

sayerhs:
@alanw0 Yes, the hack for enthalpy/temperature requires updating the entire `PropertyEvaluator` algorithms to be converted. In my opinion, they are really low priority for wind applications, so I'll add that to the list of algorithms to be converted, and revisit those.

src/Realm.C
sayerhs: This is because of the partial conversion of nalu-wind to NGP. There are a few initialization things that are still happening on host, so we don't want to pull these fields to device until the equation systems have done all their initialization logic. This is related to the enthalpy/temperature stuff that Alan was commenting about.

jrood-nrel: Thanks. Just curious.

src/Realm.C
```diff
@@ -3376,6 +3378,38 @@ void
...
+        ->mesh_meta_data_ordinal());
+
+      const stk::mesh::Selector sel = 
+        (metaData_->locally_owned_part() | metaData_->globally_shared_part())
+        & (stk::mesh::selectField(*fld));
+      const unsigned numComponents = (fld->name() == "velocity") ?
+        metaData_->spatial_dimension() : 1;
+      nalu_ngp::field_copy(
+          ngp_mesh(), sel, toField, fromField, numComponents);
+      //toField.swap(fromField);
```
sayerhs: @alanw0 OK. I went back and checked the theory and it works. I am now able to get the same results using `ngp::Field::swap` (L3409) as with `field_copy` (L3408). The issue was with lines 3395 and 3398, where we are taking a *shallow copy* of the `ngp::Field` instances that are returned from `ngp::FieldManager`, if we replace those lines with references instead, then everything will work correctly. This occurs because `ngp::Field::swap` is *swapping* the `Kokkos::View` instances within these instances but all other copies of the `ngp::Field` are unaffected and while these instances do have their states updated, the instance within `ngp::FieldManager` itself was unaffected. It took me a while to catch this because we are able to modify the fields themselves using these `ngp::Field` instances without having to use references, but that happens deep within `Kokkos::View` where it stores a single copy of the actual data.

sayerhs: You can replicate this in your STK NGP unit tests, both in `ngp::MultiStateField` and `ngp::Field` like below:

```C++

ngp::Field<double> ngpVel_np1_v1(bulk, velocity);
ngp::Field<double> ngpVel_np1_v2(bulk, velocity);
ngp::Field<double> ngpVel_n_v1(bulk, velocity.field_of_state(stk::mesh::StateN));
ngp::Field<double> ngpVel_n_v2(bulk, velocity.field_of_state(stk::mesh::StateN));

ngpVel_n_v1.swap(ngpVel_np1_v1);

// Check field values in `v2` instances
```

alanw0: Hmm, that's very subtle. That will take a little thought to figure out the best solution in general. Your solution of using references works for nalu-wind, we should probably think about a solution in stk...

PR #384, opened 2019-09-28
Fix setting cudaLimitStackSize for unit tests.

overfelt:
With this and the fix for the NgpLoopTest.NGP_basic_face_elem_loop test, all the NGP unit tests pass for me with all of the NGP unit tests pass.  The link warnings are still there and it seems odd that just those two little master element functions would cause the stack size to increase so.


sayerhs:
I don't see the fix for the face elem loop test did you forget a commit? Or was that already merged to master?

PR #387, opened 2019-10-01
Introduce driver code for geometry & TKE ABL wall function computations

sayerhs:
Introduce two new algorithm drivers `GeometryAlgDriver` and `TKEWallFuncAlgDriver`. These drivers are necessary as the ABL _assembled wall area/normal distance_ and the TKE wall function NGP algorithms are templated on the element topology type. We have a situation where the results will be incorrect if the user had multiple topologies for the wall boundary, because the field gets reset at every invocation of the Algorithm. This commit moves the rest of fields and the parallel/periodic synchronizations to the driver `pre_work` and `post_work` so that the fields are updated correctly. 

This commit also introduces NGP versions of the volume/area computations. There are unit tests to test these new algorithms, but `Realm` continues to use the non-NGP algorithms as we need several master elements to have NGP versions of `determinant` before we can switch completely. At the minimum we need `Tri32D` to have NGP versions of the methods before we can switch to NGP for 3D geometries.

### DIFF
--- a/src/ngp_algorithms/GeometryAlgDriver.C
+++ b/src/ngp_algorithms/GeometryAlgDriver.C
@@ -68,6 +68,7 @@ void GeometryAlgDriver::post_work()
   const auto& ngpMesh = realm_.ngp_mesh();
   const auto& fieldMgr = realm_.ngp_field_manager();
 
+  // TODO: Convert to NGP version of parallel and periodic updates
   auto* dualVol = meta.template get_field<ScalarFieldType>(
     stk::topology::NODE_RANK, "dual_nodal_volume");
   std::vector<const stk::mesh::FieldBase*> fields{dualVol};
--- a/src/ngp_algorithms/GeometryAlgDriver.C
+++ b/src/ngp_algorithms/GeometryAlgDriver.C
@@ -1,0 +1,145 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2019 National Renewable Energy Laboratory.                  */
+/*  This software is released under the license detailed                  */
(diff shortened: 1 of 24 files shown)


src/ngp_algorithms/GeometryAlgDriver.C
```diff
@@ -0,0 +1,145 @@
...
+    fields.push_back(wallAreaF);
+    fields.push_back(wallDistF);
+  }
+
+  for (auto* fld: fields) {
+    auto ngpFld = fieldMgr.get_field<double>(fld->mesh_meta_data_ordinal());
+    ngpFld.modify_on_device();
+    ngpFld.sync_to_host();
+  }
+
```
alanw0: Add 'TODO' item about converting these field updates and parallel_sum to ngp versions?

src/ngp_algorithms/WallFuncGeometryAlg.C
```diff
@@ -122,50 +121,50 @@ void WallFuncGeometryAlg<BcAlgTraits>::execute()
...
-      KOKKOS_LAMBDA(const MeshIndex& mi) {
-        wdist.get(mi, 0) /= warea.get(mi, 0);
-      });
-
-    // Indicate that we have modified but don't sync it
-    wdist.modify_on_device();
-    warea.modify_on_device();
-    wdistBip.modify_on_device();
-  }
+  // {
```
alanw0: Did you mean to leave all this code commented out?

src/ngp_algorithms/GeometryAlgDriver.C
sayerhs: @alanw0 There is a TODO on line 71

PR #393, opened 2019-10-01
Fix for bad merge.

overfelt:
Not sure what happened with the last merge, all the functions ended up multiply defined so nothing would compile even though it compiled fine in my own project.


sayerhs:
@overfelt Seems like Phil had implemented these in #376 also.

PR #398, opened 2019-10-04
Enable ABL forcing algorithm to work on NGP

sayerhs:
This pull request introduces the following changes:

- Convert `TurbulenceAveragingPostProcessing` and ABL statistics classes to use STK NGP and perform parallel for/reduce on device.
- Reintroduce ability to compute area-weighted ABL friction velocity (utau) statistics that got removed when transitioning ABL wall friction algorithm to NGP. 
- Introduce custom Kokkos reduction types for use with Nalu-Wind parallel reduction algorithms
- Introduce SIMD element parallel reduce utility in NGP loop utilities

Marking as draft for the time being to enable more rigorous testing on GPUs before merge.

### DIFF
--- a/include/LowMachEquationSystem.h
+++ b/include/LowMachEquationSystem.h
@@ -14,6 +14,7 @@
 #include "NaluParsing.h"
 
 #include "ngp_algorithms/NodalGradAlgDriver.h"
+#include "ngp_algorithms/WallFricVelAlgDriver.h"
 #include "ngp_algorithms/EffDiffFluxCoeffAlg.h"
 
 namespace stk{
@@ -212,7 +213,7 @@ public:
   ScalarFieldType* Udiag_{nullptr};
 
   VectorNodalGradAlgDriver nodalGradAlgDriver_;
-  NgpAlgDriver wallFuncAlgDriver_;
+  WallFricVelAlgDriver wallFuncAlgDriver_;
   std::unique_ptr<EffDiffFluxCoeffAlg> diffFluxCoeffAlg_{nullptr};
   std::unique_ptr<Algorithm> tviscAlg_{nullptr};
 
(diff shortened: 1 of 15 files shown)


sayerhs:
@alanw0 @overfelt One design question regarding the changes to `TurbulenceAveragingPostProcessing` is if we should think about minimizing the number of kernel launches across the different `compute_*` methods, and instead consolidate the logic into `NodeKernels` that execute within a single `Kokkos::parallel_for` loop over the mesh. @jrood-nrel said that it probably doesn't matter on the modern NVIDIA cards if we had asynchronous kernel execution. However, we are explicitly setting `CUDA_LAUNCH_BLOCKING=1` which I believe is a Kokkos/Tpetra requirement.

overfelt:
Are we sure we have to have CUDA_LAUNCH_BLOCKING?  And is this restriction going to be around a while?  It seems that one drawback of big kernels is taking up more memory.  Another would be trying to debug a long function instead of a bunch of short ones.  

If fewer launches is something we will have to have by 2023 in order to meet performance goals, then we should try building them into the code now.

jrood-nrel:
There are already so many kernels in trilinos that it seems futile to bother with reducing the number of kernels.

sayerhs:
@overfelt I am not suggesting big kernels, I was mostly concerned that the kernels within `TurbulenceAveragingPostProcessing` was looping over the mesh several times without having enough work within the loop for the threads. So mostly the question was trying to make sure that there was enough work. 

Regarding `CUDA_LAUNCH_BLOCKING` I think that is a question for Kokkos/Tpetra experts. 

@jrood-nrel we launch kernels for things like `create_on_device()` that creates a GPU instance of a class, so the number of kernels is a lost cause. :)

PR #399, opened 2019-10-04
Fix bug in ELEM_DATA_NEEDED enum

sayerhs:
Move `SCV_MIJ` from being counted under `SCS` data to `SCV` data.


sayerhs:
@alanw0 Yes. There is a check in the NGP loops that checks if the user requests a particular surface/volume quantity and hasn't registered that type of ME. The check uses start/end values and @marchdf tripped that error in the TAMS model implementation.

alanw0:
@sayerhs is it possible to make that TAMS check more robust? Maybe it needs a switch statement instead of start/end values? Just trying to prevent some future person from falling into a debugging hole if the enum values change again. Maybe there just needs to be a unit-test for the expected ordering of the enum values?

sayerhs:
The issue is not in TAMS. It is in one of the places that I processs elem data requests. 

I will change it to a switch statement. It is annoying because if we add another enum we have to change the code in a dozen different places that have switch statements.

alanw0:
Maybe instead of adding a switch statement, you could add a unit-test that would fail if the enums change. That way the usage of start/end values would be protected.

sayerhs:
Ok. That's a good idea. I'll add that.

marchdf:
Thanks for fixing this @sayerhs. In my fix of this type of thing in the TAMS branch there is also at line 411 of ScratchViews.h a `numScsIp` that should be `numScvIp`.

Issue #233, opened 2019-05-17
Consolidated Elem + modified timestep algorithm broken since #227


sayerhs:
#227 introduced GPU friendly `sumInto` logic for the linear system and removed the call to [apply_coeff](https://github.com/Exawind/nalu-wind/blob/master/src/SolverAlgorithm.C#L151) in [AssembleElemSolverAlgorithm::execute](https://github.com/Exawind/nalu-wind/blob/master/src/AssembleElemSolverAlgorithm.C#L127). Now the momentum diagonal is no longer populated into the field as a _side effect_ of `apply_coeff` 

Related issue: we have no regression test to catch this behavior.

alanw0:
oh, oops, I remember noticing that side-effect when I was working on the coeff-applier, and forgot about it before I finished. I'll take care of that.

sayerhs:
@alanw0 A stopgap would be to revert 

https://github.com/Exawind/nalu-wind/blob/master/src/AssembleElemSolverAlgorithm.C#L127 

back to `apply_coeff` when executing on HostSpace.

alanw0:
@sayerhs yes that would be fine. Do you want to go ahead and do that? I'm out of the office today, so I won't get to this until tonight or tomorrow.

sayerhs:
@alanw0 I've reverted back to `apply_coeff` on host. I will leave this issue open until we have a proper fix to this issue, and a regression test to catch such issues in future.

sayerhs:
Fixed in #438

PR #402, opened 2019-10-08
Fix memory leak in FixPressureAtNodeAlgorithm.

alanw0:


alanw0:
@sayerhs would a unique_ptr survive being copied into the lambda of the parallel_for (for the cpu case)?

sayerhs:
Hmm... I forgot that `KOKKOS_LAMBDA` does `[=]`, so obviously that won't work. Why does `get_coeff_applier` return a pointer? Can it return an object, that can be moved? Either way, just curiosity on my part. Don't let this hold up the pull request.

overfelt:
Just curious:
1. Is the coeffApplyer/deviceCoeffApplier pattern inside of a ifdef KOKKOS_ENABLE_CUDA the long term solution or just a short term thing until all CoeffAppliers are device compatable?
2. Alan, you are going to fix the other translation units with the free inside the ifdef?

sayerhs:
@overfelt Good point regarding (1), I think there should be an implementation of `device_pointer()` for HostOnlyCoeffApplier too. So that we don't have additional `#ifdef`s further down. 

Another thing of concern is that #233 is now proliferating to more algorithms

alanw0:
Dammit, you guys are both raising good points. Regarding 1., I agree there needs to be some cleanup to eliminate the ifdefs. Regarding James' point 2., the other occurrences aren't currently leaks because both the creation and the deletion are inside ifdefs due to the fact that the non-cuda usage still goes through the traditional apply_coeff(..) function instead of the coeffApplier. So it needs unification, but it's not a leak.
Finally, @sayerhs I agree that needs to be cleaned up. I'll look at how to clean up the save-off-diagonal feature so that this isn't a bifurcation point for that.

I'll go ahead and merge this leak-fix now, but make these cleanups next on my to-do list.

PR #403, opened 2019-10-10
Disable copy/assignment construction of `MeshInfo` object

sayerhs:
Disabling to prevent creating accidental copy of `ngp::FieldManager` object so as to ensure that we always get the same instance of `ngp::Field` that have already been instantiated on device.

### DIFF
--- a/include/Realm.h
+++ b/include/Realm.h
@@ -27,7 +27,6 @@
 #include <stk_ngp/NgpFieldManager.hpp>
 
 #include "ngp_utils/NgpMeshInfo.h"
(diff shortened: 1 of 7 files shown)


alanw0:
Sorry for being late to review this. It looks fine, I just had a question for clarification: when a copy of MeshInfo was made, was it copying FieldManager, or was it making a new FieldManager? It seems like making a new FieldManager would result in the ngp::Field issue we were discussing yesterday, but copying the FieldManager might not.

sayerhs:
The best I can tell, it was making a copy of `ngp::FieldManager` but I am not sure how to tell if it was constructing or copying.

PR #408, opened 2019-10-11
Provide Kokkos configuration and Teuchos Timer output

sayerhs:
For NGP runs print out Kokkos configuration for debugging purposes. Always
output Teuchos timers to get Trilinos solver timings based on settings.

### DIFF
--- a/nalu.C
+++ b/nalu.C
@@ -27,6 +27,7 @@
 
 // Kokkos
 #include <Kokkos_Core.hpp>
+#include <Teuchos_TimeMonitor.hpp>
 
 #include <iostream>
 #include <fstream>
@@ -238,6 +239,10 @@ int main( int argc, char ** argv )
                               false, naluEnv.parallel_comm());
 
   stk::diag::deleteRootTimer(sierra::nalu::Simulation::rootTimer());
+
+  // Write out Trilinos timers
+  Teuchos::TimeMonitor::summarize(
+    naluEnv.naluOutputP0(), false, true, false, Teuchos::Union);
   }
   Kokkos::finalize_all();
   // all done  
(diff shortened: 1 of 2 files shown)


nalu.C
```diff
@@ -238,6 +239,10 @@ int main( int argc, char ** argv )
                               false, naluEnv.parallel_comm());
 
   stk::diag::deleteRootTimer(sierra::nalu::Simulation::rootTimer());
+
+  // Write out Trilinos timers
+  Teuchos::TimeMonitor::summarize(
```
jhux2: You might consider using [TimeMonitor::report](https://trilinos.org/docs/dev/packages/teuchos/doc/html/classTeuchos_1_1TimeMonitor.html#aa7c31c3850cbf30315dd363b777c54fa) instead:

```
        RCP<ParameterList> reportParams = rcp(new ParameterList);
        reportParams->set("How to merge timer sets",   "Union");
        reportParams->set("alwaysWriteLocal",          false);
        reportParams->set("writeGlobalStats",          true);
        reportParams->set("writeZeroTimers",           false);

        TimeMonitor::report(comm.ptr(), out, reportParams);
```

sayerhs: @jhux2 what's the benefit, if the options are hardcoded in the code? Are you suggesting that we put additional parameters in input file? I just went by what was here: https://github.com/Exawind/nalu-wind/blob/ec5e76ed0590c95c61e182a12b7ea4da474a1138/src/LinearSolver.C#L143-L145

jhux2: Good question.  The parameter list interface makes the choice of options clearer.  It would also make controlling the output easier if so desired.  Finally, it would make switching over to using Teuchos stacked timer easier, if that's something the developers ultimately wanted to do.

However, I am perfectly fine with the code as implemented.

sayerhs: @jhux2 One more question: `stacked timer`? I think it would be a good idea to allow the options to be provided using an input file. Let me merge this and add an issue to take care of that.

PR #409, opened 2019-10-11
Make get_entity_master more forgiving for ghost entities.

alanw0:
Added a flag so that get_entity_master can optionallly return an
invalid entity if master is not found, instead of throwing an exception.

In particular, when filling the entity_to_col_LID mapping-vector,
we traverse all of the entities (including ghosts) and occasionally
an entity won't have a local master. In those cases, this is a
location in the mapping-vector that won't be accessed anyway, so
we just set it to -1 and continue.

### DIFF
--- a/include/TpetraLinearSystemHelpers.h
+++ b/include/TpetraLinearSystemHelpers.h
@@ -55,7 +55,8 @@ void fill_owned_and_shared_then_nonowned_ordered_by_proc(std::vector<LinSys::Glo
 
 stk::mesh::Entity get_entity_master(const stk::mesh::BulkData& bulk,
                                     stk::mesh::Entity entity,
-                                    stk::mesh::EntityId naluId);
+                                    stk::mesh::EntityId naluId,
+                                    bool throwIfMasterNotFound = true);
 
 size_t get_neighbor_index(const std::vector<int>& neighborProcs, int proc);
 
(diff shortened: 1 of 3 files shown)


sayerhs:
@alanw0 Not sure I follow. For overset meshes, we will `ghost` and we will fill to those columns, so it will participate in the graph.

alanw0:
I think overset is fine: you're right that entities ghosted for overset do participate in the graph. Similarly for non-conformal ghosting.
This issue is specifically related to periodic, with slaves having the situation where nalu-id != stk-id. Our periodic scheme ghosts the slave nodes to the same processor as the master proc, but not vice-versa. When a periodic slave is a ghost on a processor where the master is not present, then it is safe to assume that slave won't be represented in the graph. (In fact I think the slaves are not represented in the graph on any processor, except as a zero'd-out dirichlet-like entry.)

sayerhs:
@alanw0 OK Thanks, it makes sense now. The original Tpetra implementation removed rows corresponding to the mirror nodes of a periodic boundary entirely giving a smaller matrix. Did that change with recently to have dummy rows?

alanw0:
@sayerhs No I don't think that changed yet. But if we go ahead and have the tpetra system use the same ids as hypre, then we'll probably change to have dummy rows. So in that case, my comment about the zero'd out dirichlet-like entry is not correct, but would be correct if we start leaving dummy rows in the system.

PR #411, opened 2019-10-16
Update SDR EqSystem to use NGP versions of Algorithms

sayerhs:
Update nodal gradient and effective diffusive flux calculations in Specific Dissipation Rate EqSys. 

Remove more unused files.


sayerhs:
@jrood-nrel This will introduce diffs in the following tests:

```
FAIL: elemBackStepLRSST.......................     1.6617s 1.4200e-14 2.0505e-14
FAIL: elemBackStepLRSST_Input.................     2.0382s 3.9998e-15 6.0826e-16
```

You can re-bless these files.

PR #413, opened 2019-10-17
Step 1 of cleaning up CoeffApplier.

alanw0:
This commit makes CoeffApplier handle extracting/saving the diagonal.

Still more things to do:
1. Resolve coeffApplier vs deviceCoeffApplier usage. Algs shouldn't
need to get coeffApplier and then ask it to produce deviceCoeffApplier.
2. Handle the 'fix_overset_rows' functionality?
3. reduce duplication between TpetraLinearSystem and
TpetraSegregatedLinearSystem

### DIFF
--- a/include/EquationSystem.h
+++ b/include/EquationSystem.h
@@ -353,7 +353,7 @@ public:
   std::string dofName_{"undefined"};
 
   bool extractDiagonal_{false};
-  ScalarFieldType* Udiag_{nullptr};
+  virtual ScalarFieldType* get_diagonal_field() { return nullptr; }
 
   // owner equation system
   /*EquationSystem *ownerEqs_;*/
--- a/include/EquationSystem.h
+++ b/include/EquationSystem.h
@@ -353,6 +353,7 @@ public:
   std::string dofName_{"undefined"};
 
   bool extractDiagonal_{false};
+  ScalarFieldType* Udiag_{nullptr};
 
   // owner equation system
   /*EquationSystem *ownerEqs_;*/
--- a/src/AssembleElemSolverAlgorithm.C
+++ b/src/AssembleElemSolverAlgorithm.C
@@ -89,14 +89,12 @@ AssembleElemSolverAlgorithm::execute()
 
   auto ngpKernels = nalu_ngp::create_ngp_view<Kernel>(activeKernels_);
 
-#ifdef KOKKOS_ENABLE_CUDA
   CoeffApplier* coeffApplier = eqSystem_->linsys_->get_coeff_applier();
   CoeffApplier* deviceCoeffApplier = coeffApplier->device_pointer();
 
   double diagRelaxFactor = diagRelaxFactor_;
   int rhsSize = rhsSize_;
   unsigned nodesPerEntity = nodesPerEntity_;
-#endif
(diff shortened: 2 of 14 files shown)


include/EquationSystem.h
```diff
@@ -353,6 +353,7 @@ class EquationSystem
   std::string dofName_{"undefined"};
 
   bool extractDiagonal_{false};
+  ScalarFieldType* Udiag_{nullptr};
```
sayerhs: Could this be a method instead? something like `get_diagonal_field()` that returns `nullptr` in base implementation and is specialized for `MomentumEquationSystem` where it returns `Udiag_`?

src/AssembleElemSolverAlgorithm.C
```diff
@@ -112,25 +110,23 @@ AssembleElemSolverAlgorithm::execute()
...
         extract_vector_lane(smdata.simdlhs, simdElemIndex, smdata.lhs);
-        for (int ir=0; ir < rhsSize_; ++ir)
-          smdata.lhs(ir, ir) /= diagRelaxFactor_;
-        apply_coeff(nodesPerEntity_, smdata.ngpElemNodes[simdElemIndex],
+        for (int ir=0; ir < rhsSize; ++ir)
+          smdata.lhs(ir, ir) /= diagRelaxFactor;
+        (*deviceCoeffApplier)(nodesPerEntity, smdata.ngpElemNodes[simdElemIndex],
                     smdata.scratchIds, smdata.sortPermutation, smdata.rhs, smdata.lhs, __FILE__);
       }
 #else
```
sayerhs: Is there a reason to keep this `#ifdef`? Or can we collapse this?

src/AssembleNGPNodeSolverAlgorithm.C
```diff
@@ -91,10 +91,8 @@ AssembleNGPNodeSolverAlgorithm::execute()
   const stk::mesh::EntityRank entityRank = stk::topology::NODE_RANK;
   const int rhsSize = rhsSize_;
 
-#ifdef KOKKOS_ENABLE_CUDA
   CoeffApplier* coeffApplier = eqSystem_->linsys_->get_coeff_applier();
   CoeffApplier* deviceCoeffApplier = coeffApplier->device_pointer();
```
sayerhs: Are you thinking that `CoeffApplier` returned currently by `get_coeff_applier()` will be managed internally within LinearSystem and will in future return the pointer currently accessed by `device_pointer()`? That would be a good cleanup.

include/EquationSystem.h
alanw0: Yes that's a good idea, I'll do that.

src/AssembleElemSolverAlgorithm.C
alanw0: That ifdef is around the simd-loop, we probably could collapse that. Possibly set num-simd-elems to 1 if cuda or something like that?

src/AssembleNGPNodeSolverAlgorithm.C
alanw0: Yes I agree that would be a good next step.

Issue #418, opened 2019-10-22
Errors building with CUDA


jrood-nrel:
@alanw0 I haven't tested non-CUDA, but I was trying to run the ablNeutralEdge case on the GPU and it looks like some errors popped up in #413.

```
.../nalu-wind/src/LinearSystem.C(132): error: name followed by "::" must be a class or namespace name
.../nalu-wind/src/LinearSystem.C(141): error: identifier "linSys_" is undefined
```

PR #419, opened 2019-10-23
Fix compile-error for cuda build.

alanw0:


alanw0:
This fixes the issue in #418

PR #427, opened 2019-10-30
Revert 98bc2ac and update master elements.

overfelt:
It was pointed out that some of the master element calls used
local scratch when it would be better to pass in a Kokkos View
to save stack space.


overfelt:
Yes, all the tests are fixed.  It turned out to be nothing.  Needed to up the reserved stack size in Simulation.h.

PR #429, opened 2019-10-31
Check for nvcc compiler in preprocessor to avoid unreachable statement warnings

jrood-nrel:
This doesn't get rid of all the warnings, but should stop most of these `unreachable statement` warnings. It seems `nvcc` uses an older GCC because these warnings are deprecated in more recent versions of GCC.


jrood-nrel:
Oops. I need to actually fix a couple things on this I just realized. To account for some other preprocessor macros.

PR #432, opened 2019-11-03
Step 2 of coeff-applier cleanup.

alanw0:
Make coeff-applier pointers be owned/managed by
LinearSystem. By default, get_coeff_applier returns a device pointer.
This also makes sense when running a non-cuda build, since the
host is the same as the device.
If we need a host pointer when running a cuda build, we can add a
get_host_coeff_applier.

Next step will be to try to move the extract-diagonal code into
something (a functor?) owned by SolverAlgorithm, and also put
the fix-overset code in that.

### DIFF
--- a/include/LinearSystem.h
+++ b/include/LinearSystem.h
@@ -79,7 +79,14 @@ public:
     EquationSystem *eqSys,
     LinearSolver *linearSolver);
 
-  virtual ~LinearSystem() {}
+  virtual ~LinearSystem() {
+    if (hostCoeffApplier != nullptr) {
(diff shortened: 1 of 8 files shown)


include/LinearSystem.h
```diff
@@ -266,6 +276,9 @@ class LinearSystem
   bool recomputePreconditioner_;
   bool reusePreconditioner_;
 
+  CoeffApplier* hostCoeffApplier = nullptr;
```
sayerhs: `std::unique_ptr` here?

PR #435, opened 2019-11-07
Reinstate fix_overset_rows for CPU only runs

sayerhs:
This fix is needed for some of the turbine production runs that are breaking with the latest master. Eventually this will be handled by the CoeffApplier like logic for NGP simulations.


gantech:
This made a great difference to my turbine runs and allowed the run to go uninterrupted for 5+ days . My runs were constantly blowing up at inconsistent times before this fix.

PR #436, opened 2019-11-08
Some steps to more modern CMake

jrood-nrel:
This is a first step to localizing source and include files. This seems to work by merely removing the global source file property, enforcing CMake >= 3.13, and using `target_sources`.

@sayerhs I started to look at using `target_include_directories` in the `src` directory here, but since there is currently one large `include` directory, would we have to merge the `include` directory with the `src` directory to localize the include paths per directory? Or would we just cut off the directory prefixes and use relative paths in the `.C` `#include` lines and use the explicit paths in `target_include_directories` that pair each `src` subdirectory with its `include` subdirectory? Hard to explain my thought with that correctly I think.

I just realized we also don't seem to need a C compiler here so I removed that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,10 +27,17 @@ set(CMAKE_CXX_STANDARD 11)       # Set nalu-wind C++11 standard
 set(CMAKE_CXX_EXTENSIONS OFF)    # Do not enable GNU extensions
 set(CMAKE_CXX_STANDARD_REQUIRED) # Force error if C++11 standard is not supported
 
+# Create targets
+set(nalu_ex_name "naluX")
+set(utest_ex_name "unittestX")
+add_library(nalu "")
+add_executable(${nalu_ex_name} nalu.C)
+add_executable(${utest_ex_name} unit_tests.C)
+
 find_package(MPI REQUIRED)
-include_directories(SYSTEM ${MPI_C_INCLUDE_PATH})
-include_directories(SYSTEM ${MPI_CXX_INCLUDE_PATH})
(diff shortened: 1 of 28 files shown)


CMakeLists.txt
```diff
@@ -236,22 +236,6 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/NaluVersionInfo.h.in"
...
-   endforeach()
-endfunction(add_sources)
-add_subdirectory(src)
-add_subdirectory(unit_tests)
-get_property(NALU_SOURCES GLOBAL PROPERTY GlobalSourceList)
-get_property(NALU_UNIT_SOURCES GLOBAL PROPERTY GlobalUnitSourceList)
-
 include_directories(${CMAKE_SOURCE_DIR}/include)
 include_directories(${CMAKE_BINARY_DIR}/include)
 add_library(nalu ${NALU_SOURCES})
```
sayerhs: This line can just be `add_library(nalu)`? Similarly with `add_executable` for `unittest`

jrood-nrel: Ah, yeah I noticed that after I submitted this.

jrood-nrel:
I made a few more changes to localize to the `nalu` target. We'll talk this over before I merge it though. I have a somewhat loose understanding of exactly what I'm doing with the latest commit but it compiled for me.

jrood-nrel:
Ok, I think I'm going to merge this today, but I will keep a close eye on if this will break compiling for anyone. I will try to broadcast that we will require CMake 3.13 for this.

PR #439, opened 2019-11-13
Ensure gradient call happens regardless of sym BC type.

marchdf:
The calculation of the gradient of velocity was not happening when the
symmetry BC type was something other than GENERAL_WEAK.


psakievich:
Good catch thanks!

PR #443, opened 2019-11-14
Require CMake 3.13.0. Update CMake code to use more modern practices.

jrood-nrel:
Since this requires a newer version of CMake and can have other unintended consequences, we will merge it once developers can verify this will work on our main development machines.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 ############################ BASE ######################################
-cmake_minimum_required (VERSION 3.10.0)
-project(Nalu CXX C Fortran)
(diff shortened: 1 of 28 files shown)


sayerhs:
@jrood-nrel I would suggest a two stage approach, move to modern CMake code but retain the same CMake minimum version as Trilinos and move to 3.13 at a later date. This will allow the other good changes to get into the code base quicker and not be held up because of a version requirement that is not necessary for the changes. Or is there a change that absolute requires this version requirement change?

jrood-nrel:
@sayerhs I will take your suggestion. I have solved the need to update the required version. I will make a new pull request.

PR #444, opened 2019-11-15
Update CMake code to use more modern practices

jrood-nrel:
This version of my changes does not require any newer version of CMake than we are already using. After some research, it turns out `target_sources` was only allowed to take relative file paths in version 3.13.0, so I had to make the paths to the source files explicit. I also tested this again for the GPU build.


sayerhs:
@jrood-nrel Should we make sure that the yaml-cpp version change won't break builds on SNL machines?

jrood-nrel:
I will just reopen this with the yaml-cpp 0.5.3 logic in.

PR #470, opened 2019-12-17
Sampling plane capability

lawrenceccheung:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [ X ] Feature enhancement

Modified `DataProbePostProcessing.C` to output sample planes directly from Nalu-Wind (without having to specify them as a series of lines).

The `plane_specifications` can co-exist with `line_of_site_specifications` and uses the same data structures.

Documentation of this feature has also been added to `nalu_inp.rst`.

## Checklist

of reviewers and developers.

*All pull requests*
- [X ] Builds successfully (*must test on at least one system/compiler combination*)
  - Operating systems 
    - [X ] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [X ] Intel compilers
    - [ ] NVIDIA CUDA
- [X ] Compiles without warnings
- [X ] Passes all unit tests

*For new code updates*
- [X ] Documentation updates - additions to user, theory, and verification manuals


src/DataProbePostProcessing.C
```diff
@@ -758,7 +966,102 @@ DataProbePostProcessing::provide_output_txt(
...
+	    if ( processorId == NaluEnv::self().parallel_rank()) {    
+
+	      const int N1 = probeInfo->edge1NumPoints_[inp];
+	      const int N2 = probeInfo->edge2NumPoints_[inp];
+	      const int pointsPerPlane = N1*N2;
+	      const int numPlanes = probeInfo->offsetSpacings_[inp].size();
+
+	      // Get the path to the file name, and create any directories necessary
+	      // - Get the path
+	      size_t pathfound;
```
sayerhs: Please convert lines 990-1007 to C++. `sys/stat.h` is not standard if I remember correctly. Has this been tested on different compilers/operating systems that this will be expected to run on. 

While, this capability would be useful elsewhere too, e.g., SurfaceForceAndMoment, I would prefer to find some portable/standard way of implementing this and have it tested across different platforms. At that point, I would also recommend elevating this to a utility function that we can use in other classes that create text files.

src/xfer/Transfer.C
```diff
@@ -96,9 +96,9 @@ Transfer::load(const YAML::Node & node)
   realmPairName_.second = realmPair[1].as<std::string>() ;
 
   // set bools for variety of mesh part declarations
-  const bool hasOld = node["mesh_part_pair"];
-  const bool hasNewFrom = node["from_target_name"];
-  const bool hasNewTo = node["to_target_name"];
+  const bool hasOld = node["mesh_part_pair"]?true:false;
```
sayerhs: are these changes to handle the latest version of `yaml-cpp`?

lawrenceccheung: Yes, I couldn't compile it with the latest version of `yaml-cpp` unless I made those changes.

sayerhs:
I wanted to loop @tonyinme in on the discussion. He has been using STK Transfers with sampling planes and has python scripts that he has used to extract and process the SWiFT data for the IEA task. From an I/O perspective, the Exodus/NetCDF approach is better than text files. Also eliminates  things like _"I think we should really consider making all the data probes comma delineated to make them easier to read into post-processing scripts."_

tonyinme:
hey all, thanks for including me in the discussion.

@sayerhs, I think the best is to have both options available (text and exodus).

@lawrenceccheung , I see that you added options for text and exodus but they are not in the documentation. Can you please add these to the documentation?

lawrenceccheung:
Thanks @tonyinme.  Good catch, let me add that to the `nalu_inp.rst` file.

I'm still looking for a nice, clean way to recursively create directories in C++.  There's some nifty options out there, but they depend on C++17 or boost libraries, so it may not work everywhere.  I'll update the code once I find something good.

sayerhs:
@lawrenceccheung Yeah, I can't wait for us to be able to turn on `-std=c++17` and start using `std::filesystem`. I've been hesitating to use `boost::filesystem` because of Trilinos' push to remove boost dependency. That said, we do use  `boost::program_options` and `boost::format` in Nalu-Wind, so we already have a boost dependency. I think `boost::filesystem` is probably a cleaner short-term route and we can swap it out with `std::filesystem` eventually.

lawrenceccheung:
@sayerhs: Okay, I changed things to use the `boost::filesystem` calls and that does make things cleaner for now.   However, apparently `boost::filesystem` isn't a header-only library and needs to be linked against `libboost_filesystem` at compile-time.  That shouldn't be an issue if other parts of Nalu-Wind also call boost, so the library should already exist.  

I had to also change `CMakeLists.txt` to include the library when linking.  When compiling, just make sure to call cmake with the flag `-DBOOST_DIR=/path/to/boost` so that it can locate the library and link to it.

CMakeLists.txt
```diff
@@ -54,6 +54,14 @@ else(Trilinos_BUILD_SHARED_LIBS)
   set(Nalu_LIBRARY_TYPE STATIC)
 endif(Trilinos_BUILD_SHARED_LIBS)
 
+############################ BOOST #####################################
+find_library(BOOST_LIBRARIES NAMES boost_filesystem PATHS ${BOOST_DIR}/lib)
```
sayerhs: Can you please use https://cmake.org/cmake/help/latest/module/FindBoost.html instead?

lawrenceccheung: Thanks for pointing me to `FindBoost`.  I modified `CMakeLists.txt` to use that instead.

CMakeLists.txt
```diff
@@ -61,6 +61,14 @@ else(Trilinos_BUILD_SHARED_LIBS)
   set(Nalu_LIBRARY_TYPE STATIC)
 endif(Trilinos_BUILD_SHARED_LIBS)
 
+############################ BOOST #####################################
+find_package(Boost REQUIRED COMPONENTS filesystem)
+if (DEFINED Boost_LIBRARIES)
```
sayerhs: `find_package` with `REQUIRED` will throw a fatal error, so you don't need to check it manually as you are doing :)

lawrenceccheung: Haha, you're right.  I wanted to be doubly-sure that it was linking to something that it found, but yes, the manual check is unnecessary.

PR #471, opened 2019-12-18
add parser option to rebalance mesh using stk_balance

jvo1012:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

This change adds the parser option to turn on rebalance using stk_balance with the input file parameter ``rebalance_mesh``. Users need to provide a decomposition method ``stk_rebalance_method`` in the input file as well. 

## Checklist

of reviewers and developers.

*All pull requests*
- [ x] Builds successfully (*must test on at least one system/compiler combination*)
  - Operating systems 
    - [x ] Linux
    - [ ] MacOS
  - Compilers 
    - [x ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [x ] NVIDIA CUDA
- [x ] Compiles without warnings
- [x ] Passes all unit tests
- [x ] Passes all regression tests

*For new code updates*
- [x ] Documentation updates - additions to user, theory, and verification manuals

### DIFF
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -499,6 +499,11 @@ Realm::initialize()
 
   // rebalance mesh using stk_balance
   if (rebalanceMesh_) {
+#ifndef HAVE_ZOLTAN2_PARMETIS
+  if (rebalanceMethod_ == "parmetis")
+    throw std::runtime_error("Zoltan2 is not built with parmetis enabled, "
+                             "try a geometric balance method instead (rcb or rib)");
+#endif
     stk::balance::GraphCreationSettings rebalanceSettings;
     rebalanceSettings.setDecompMethod(rebalanceMethod_);
     stk::balance::balanceStkMesh(rebalanceSettings, *bulkData_);
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -475,10 +475,6 @@ Realm::initialize()
   timerPopulateMesh_ += time;
   NaluEnv::self().naluOutputP0() << "Realm::ioBroker_->populate_mesh() End" << std::endl;
 
-  if (doBalanceNodes_) {
-    balance_nodes();
-  }
-
   // If we want to create all internal edges, we want to do it before
   // field-data is allocated because that allows better performance in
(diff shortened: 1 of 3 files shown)


src/Realm.C
```diff
@@ -497,6 +501,13 @@ Realm::initialize()
   timerPopulateFieldData_ += time;
   NaluEnv::self().naluOutputP0() << "Realm::ioBroker_->populate_field_data() End" << std::endl;
 
+  // rebalance mesh using stk_balance
+  if (rebalanceMesh_) {
```
sayerhs: Shouldn't rebalance happen before `create_edges` (possibly?) but definitely before `populate_field_data`?

jvo1012: The rebalance depends on the coordinates so it needs to happen after populate_field_data(), otherwise it will segfault.

sayerhs: Interesting, I didn't know that. I was under the impression that Zoltan only required the connectivity graph. Do you know what it uses coordinates for? 

So the next question is how do we handle the interaction with `balance_nodes()` (called above in line 479). If the user activates both, then this Zoltan rebalance would undo what was done in `balance_nodes`, correct? So should we be setting `balance_nodes()` to happen after `balanceStkMesh`? I assume since only ownership is changing, fields are consistent and doesn't get affected?

alanw0: @sayerhs rebalance has a couple of different options for the algorithm that is used internally. If a graph-based algorithm is used, then coordinates would not be needed. But if a geometric algorithm is used (like RCB) then coordinates are needed. Also, stk-balance generally does a proximity search so that contact surfaces can be kept together in the decomposition, and the proximity search needs coordinates.

Good point about balance_nodes, that needs to happen after the rebalance. And yes the fields should be fine after rebalance, field-data gets moved with the entities that are being moved.

Ideally rebalance would be a little cheaper if it is done prior to create_edges. But that's less important than putting rebalance after populate_field_data (which is mandatory).

PR #478, opened 2019-12-20
More shuffling around of CMake

jrood-nrel:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement
- [x] Cleanup

Just shuffling around CMake code to make it more consistent.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [x] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,6 @@
 ############################ BASE ######################################
 cmake_minimum_required (VERSION 3.10.0 FATAL_ERROR)
 project(Nalu CXX Fortran)
-
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 
 ########################## OPTIONS #####################################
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -34,40 +33,36 @@ add_library(nalu "")
...
-  separate_arguments(OpenMP_CXX_FLAGS)
-  separate_arguments(OpenMP_Fortran_FLAGS)
-  target_compile_options(nalu PUBLIC $<$<COMPILE_LANGUAGE:CXX>:${OpenMP_CXX_FLAGS}>)
-  target_compile_options(nalu PUBLIC $<$<COMPILE_LANGUAGE:Fortran>:${OpenMP_Fortran_FLAGS}>)
-endif()
+target_link_libraries(nalu PUBLIC $<$<BOOL:${MPI_CXX_FOUND}>:MPI::MPI_CXX>)
+target_link_libraries(nalu PUBLIC $<$<BOOL:${MPI_Fortran_FOUND}>:MPI::MPI_Fortran>)
 
 ########################## TRILINOS ####################################
+set(USER_Trilinos_DIR ${Trilinos_DIR})
```
sayerhs: What is the reason for this additional variable?

CMakeLists.txt
```diff
@@ -34,40 +33,36 @@ add_library(nalu "")
...
 set(CMAKE_PREFIX_PATH ${Trilinos_DIR} ${CMAKE_PREFIX_PATH})
-find_package(Trilinos REQUIRED)
+find_package(Trilinos QUIET REQUIRED)
 if(Trilinos_FOUND)
-  message(STATUS "Found Trilinos = ${Trilinos_DIR}")
+  message(STATUS "Found Trilinos = ${USER_Trilinos_DIR}")
```
sayerhs: How can we be certain that `find_package` settled on the Trilinos in `Trilinos_DIR`? It seems that we are just depending on the fact that we pushed it to the top of `CMAKE_PREFIX_PATH` stack.

CMakeLists.txt
```diff
@@ -192,7 +194,7 @@ if(ENABLE_ALL_WARNINGS)
   endif()
 endif()
 
-# Add our flags according to language
+# Add our extra flags according to language
 separate_arguments(NALU_CXX_FLAGS)
 separate_arguments(NALU_Fortran_FLAGS)
 target_compile_options(nalu PUBLIC $<$<COMPILE_LANGUAGE:CXX>:${NALU_CXX_FLAGS}>)
```
sayerhs: We used to have the `NALU_CXX_FLAGS` printed out in the CMake output. It looks like that got lost during the CMake modernization refactor. Can we reinstate that?

CMakeLists.txt
jrood-nrel: That's true. I didn't like it reporting the path with `lib/cmake` in it and overwriting `Trilinos_DIR`. The only way I could figure out how to avoid it was to report the path by saving the original. The only way to get consistent directories for the user seems to be by using `CMAKE_PREFIX_PATH`. I'm not sure of the best way to get this all consistent.

sayerhs: May be we should switch to [<PackageName>_ROOT](https://cmake.org/cmake/help/latest/policy/CMP0074.html). But I see what you are saying about reporting paths.

CMakeLists.txt
jrood-nrel: I actually removed that by choice. I found it really difficult to report the _final_ set of compiler flags during configure that CMake will actually use. Before, I was reporting the build type flags, Trilinos flags, and our own custom flags, but I would prefer to just report the final flags for C++ and Fortran. I could reinstate the old flag reporting, but I would like to find a better way. I found I could only rely on `CMAKE_EXPORT_COMPILE_COMMANDS` to tell me that.

jrood-nrel: Since I call `target_compile_options` to add flags rather than build up a variable with all of them, which I could report, I was hoping there was some kind of reporting feature for it, but there doesn't seem to be. I'm still thinking about this, but I don't have a good solution. The old way of reporting started to become really incomplete after using `target_compile_options`.

jrood-nrel:
I'm going to merge this because I need it to enable `Boost_DIR` to get CDash working again. It probably makes sense to move to `_ROOT` instead in the future.

PR #488, opened 2020-01-11
Address code review of first AssembleNodalGradPOpenBoundaryAlgorithm

overfelt:
The changes were extensive enough that a new branch was created.

This has been tested with gcc 7.2 and I still need to compile and
test for Cuda.

**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [X] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [X] Linux
    - [ ] MacOS
  - Compilers 
    - [X] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [X] Passes all unit tests

*For new code updates*
- [X] New regression tests exercising the new code

### DIFF
--- a/src/ngp_algorithms/NodalGradPOpenBoundaryAlg.C
+++ b/src/ngp_algorithms/NodalGradPOpenBoundaryAlg.C
@@ -115,14 +115,15 @@ NodalGradPOpenBoundary<AlgTraits>::execute()
       const int* ipNodeMap = meFC->ipNodeMap();
 
       auto& faceView = fdata.simdFaceView;
+      auto& elemView = fdata.simdElemView;
       const auto v_areav        = faceView.get_scratch_view_2D(exposedAreaVec);
       const auto v_dnv          = faceView.get_scratch_view_1D(dualNodalVol);
       const auto face_p_field   = faceView.get_scratch_view_1D(exposedPressureField);
-      const auto elem_p_field   = faceView.get_scratch_view_1D(pressureField);
       const auto v_coord        = faceView.get_scratch_view_2D(coordsID);
+      const auto elem_p_field   = elemView.get_scratch_view_1D(pressureField);
 
       const auto meFaceViews = fdata.simdFaceView.get_me_views(CURRENT_COORDINATES);
-      const auto meElemViews = fdata.simdElemView.get_me_views(CURRENT_COORDINATES);
+      const auto meElemViews = elemView.get_me_views(CURRENT_COORDINATES);
       const auto v_shape_fcn = useShifted ? meFaceViews.fc_shifted_shape_fcn :  meFaceViews.fc_shape_fcn;
       const auto e_shape_fcn = useShifted ? meElemViews.scs_shifted_shape_fcn:  meElemViews.scs_shape_fcn;
 
--- a/src/ngp_algorithms/NodalGradPOpenBoundaryAlg.C
+++ b/src/ngp_algorithms/NodalGradPOpenBoundaryAlg.C
@@ -1,0 +1,161 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2019 Sandia Corporation.                                    */
+/*  This software is released under the license detailed                  */
+/*  in the file, LICENSE, which is located in the top-level Nalu          */
+/*  directory structure                                                   */
+/*------------------------------------------------------------------------*/
+
+#include "ngp_algorithms/NodalGradPOpenBoundaryAlg.h"
+#include "Algorithm.h"
+
+#include "BuildTemplates.h"
+#include "master_element/MasterElement.h"
+#include "master_element/MasterElementFactory.h"
+#include "ngp_algorithms/ViewHelper.h"
+#include "ngp_utils/NgpLoopUtils.h"
+#include "ngp_utils/NgpFieldOps.h"
+#include "Realm.h"
+#include "ScratchViews.h"
+#include "SolutionOptions.h"
+#include "utils/StkHelpers.h"
+
+namespace sierra{
+namespace nalu{
+
+//==========================================================================
+// Class Definition
+//==========================================================================
+// NodalGradPOpenBoundary - adds in boundary contribution
+//                          for elem/edge proj nodal gradient
+//==========================================================================
+//--------------------------------------------------------------------------
+//-------- constructor -----------------------------------------------------
+//--------------------------------------------------------------------------
+template <typename AlgTraits>
+NodalGradPOpenBoundary<AlgTraits>::NodalGradPOpenBoundary(
+  Realm &realm,
+  stk::mesh::Part *part,
+  const bool useShifted)
+  : Algorithm(realm, part),
+    useShifted_(useShifted),
+    zeroGrad_(realm_.solutionOptions_->explicitlyZeroOpenPressureGradient_),
+    massCorr_(realm_.solutionOptions_->activateOpenMdotCorrection_),
+    exposedAreaVec_ (get_field_ordinal(realm_.meta_data(), "exposed_area_vector", realm.meta_data().side_rank())),
+    dualNodalVol_   (get_field_ordinal(realm_.meta_data(), "dual_nodal_volume")),
+    exposedPressureField_   (get_field_ordinal(realm_.meta_data(), (massCorr_ ? "pressure" : "pressure_bc"))),
+    pressureField_ (get_field_ordinal(realm_.meta_data(), "pressure")),
+    gradP_         (get_field_ordinal(realm_.meta_data(), "dpdx")),
+    coordinates_   (get_field_ordinal(realm_.meta_data(), realm.get_coordinates_name())),
+    meFC_  (MasterElementRepo::get_surface_master_element<typename AlgTraits::FaceTraits>()),
+    meSCS_ (MasterElementRepo::get_surface_master_element<typename AlgTraits::ElemTraits>()),
+    faceData_(realm.meta_data()),
+    elemData_(realm.meta_data())
+{
+  const int coordinates = get_field_ordinal(realm.meta_data(), realm.get_coordinates_name());
+  faceData_.add_coordinates_field(coordinates, AlgTraits::nDim_, CURRENT_COORDINATES);
+  elemData_.add_coordinates_field(coordinates, AlgTraits::nDim_, CURRENT_COORDINATES);
+  faceData_.add_cvfem_face_me   (meFC_);
+  elemData_.add_cvfem_surface_me(meSCS_);
+
+  faceData_.add_face_field(exposedAreaVec_, AlgTraits::numFaceIp_, AlgTraits::nDim_);
+  faceData_.add_gathered_nodal_field(dualNodalVol_,1);
+  faceData_.add_gathered_nodal_field(exposedPressureField_,1);
+  faceData_.add_gathered_nodal_field(pressureField_,1);
+  faceData_.add_gathered_nodal_field(gradP_,       AlgTraits::nodesPerFace_, AlgTraits::nDim_);
+  faceData_.add_coordinates_field   (coordinates_, AlgTraits::nDim_, CURRENT_COORDINATES);
+
+  const ELEM_DATA_NEEDED fc_shape_fcn  = useShifted_ ?  FC_SHIFTED_SHAPE_FCN :  FC_SHAPE_FCN;
+  const ELEM_DATA_NEEDED scs_shape_fcn = useShifted_ ? SCS_SHIFTED_SHAPE_FCN : SCS_SHAPE_FCN;
+  faceData_.add_master_element_call( fc_shape_fcn, CURRENT_COORDINATES);
+  elemData_.add_master_element_call(scs_shape_fcn, CURRENT_COORDINATES);
+}
+
+
+
+
+//--------------------------------------------------------------------------
+//-------- execute ---------------------------------------------------------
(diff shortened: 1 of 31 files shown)


unit_tests/edge_kernels/UnitTestMomentumABLWallFuncEdge.C
```diff
@@ -20,9 +20,10 @@ TEST_F(MomentumABLKernelHex8Mesh, NGP_abl_wall_func)
 {
   if (bulk_.parallel_size() > 1) return;
 
+  const std::string meshSpec;
   const bool doPerturb = false;
   const bool generateSidesets = true;
-  fill_mesh_and_init_fields(doPerturb, generateSidesets);
+  fill_mesh_and_init_fields(meshSpec, doPerturb, generateSidesets);
```
sayerhs: I find it really unintuitive and weird to declare `meshSpec` as an empty string and then pass that to a method. I'd to go look at the method to understand what is going on, and I see why you did it. Why not have an overload for `fill_mesh_and_init_fields` 

```C++
// Old behavior
void fill_mesh_and_init_fields(bool doPeturb, bool generateSidesets); 
{
  const std::string meshSpec = "generated:1x1x" + std::to_string(bulk.parallel_size());
  fill_mesh_and_init_fields(meshSpec, doPerturb, generateSidesets);
}

void fill_mesh_and_init_fields(std::string meshSpec, bool doPerturb, bool generateSidesets);
{
 // implementation
}
```

Now only the second method needs to be overridden by the subclasses. 

void

unit_tests/ngp_algorithms/UnitTestNodalGradPOpenBoundary.C
```diff
@@ -0,0 +1,122 @@
...
+#include "ngp_algorithms/GeometryBoundaryAlg.h"
+#include "ngp_algorithms/NodalGradAlgDriver.h"
+#include "ngp_algorithms/GeometryAlgDriver.h"
+
+TEST_F(LowMachKernelHex8Mesh, NGP_nodal_grad_popen)
+{
+  // Only execute for 1 processor runs
+  if (bulk_.parallel_size() > 1) return;
+
+  const std::string meshSpec = "generated:20x20x20";
```
sayerhs: What is the motivation to do this for a `20x20` face instead of a reference element?

unit_tests/ngp_algorithms/UnitTestNodalGradPOpenBoundary.C
```diff
@@ -0,0 +1,122 @@
...
+  unit_test_utils::HelperObjects helperObjs(bulk_, stk::topology::HEX_8, 1, partVec_[0]);
+
+  const double zero= 0.0;
+  const double one = 1.0;
+  const double two = 2.0;
+  const double oneVecTwelve[12] = {one, one, one, one, one, one, one, one, one, one, one, one};
+  stk::mesh::field_fill(zero, *dpdx_);
+  stk::mesh::field_fill(two, *dnvField_);
+  stk::mesh::field_fill(one, *pressure_);
+  stk::mesh::field_fill_component(oneVecTwelve, *exposedAreaVec_);
```
sayerhs: `dnvField_` and `exposedAreaVec_` are computed using master elements within `fill_mesh_and_init_fields()`, what is the motivation for overriding these? And they get overwrriten by lines 43-48 anyway.

unit_tests/ngp_algorithms/UnitTestNodalGradPOpenBoundary.C
```diff
@@ -0,0 +1,122 @@
...
+
+  stk::mesh::Part* surface1 = meta_.get_part("surface_1");
+  {
+    sierra::nalu::GeometryAlgDriver geomAlgDriver(helperObjs.realm);
+    geomAlgDriver.register_face_algorithm<sierra::nalu::GeometryBoundaryAlg>(
+      sierra::nalu::WALL, surface1, "geometry");
+    geomAlgDriver.execute();
+  }
+
+  stk::mesh::field_fill(2.0, *dnvField_);
```
sayerhs: Not needed?

unit_tests/ngp_algorithms/UnitTestNodalGradPOpenBoundary.C
```diff
@@ -0,0 +1,122 @@
...
+                                  { 0.0,  0.0,   0.0}};
+
+    const double tol = 1.0e-16;
+
+    ngp::Field<double> ngpdpdx(bulk_, *dpdx_);
+    ngp::Field<double> ngpcord(bulk_, *coordinates_);
+    ngpdpdx.modify_on_device();
+    ngpdpdx.sync_to_host();
+    ngpcord.modify_on_device();
+    ngpcord.sync_to_host();
```
sayerhs: I don't understand why this is necessary.

unit_tests/ngp_algorithms/UnitTestNodalGradPOpenBoundary.C
```diff
@@ -0,0 +1,122 @@
...
+<<" "<<cord[2]
+<<" dpdxref:"
+<<" "<<dpdxref[j][0]
+<<" "<<dpdxref[j][1]
+<<" "<<dpdxref[j][2]
+<<" dpdx:"
+<<" "<<dpdx[0]
+<<" "<<dpdx[1]
+<<" "<<cord[2]
+<<std::endl;
```
sayerhs: Is this left over from debugging?

unit_tests/edge_kernels/UnitTestMomentumABLWallFuncEdge.C
overfelt: We could do that.   But then there are two functions to have virtual overrides all the way down the subclasses.

unit_tests/ngp_algorithms/UnitTestNodalGradPOpenBoundary.C
overfelt: In order to have the ngp exactly match the non-ngp and give better coverage.  It may be overkill but the 20x20 runs in less than a second.

unit_tests/edge_kernels/UnitTestMomentumABLWallFuncEdge.C
sayerhs: @overfelt can you explain a bit more? I was thinking that since the first method was implemented in terms of the secoyit wouldn't need to be overridden as often by sub classes. Am I overlooking something?

unit_tests/ngp_algorithms/UnitTestNodalGradPOpenBoundary.C
sayerhs: I wasn't worried about the runtime. I guess I am unclear why a 1x1 face (or a 3x3) won't match the non-ngp exactly. Also, I was curious if you had found some differences in results that didn't show up in a 1x1 or a 3x3 face that we should look into with other algorithms also.

alanw0: I'm curious about this too. I agree that running 8000 elements is probably not a significant runtime issue, but is it the case that 1x1 or 2x2 gave non-matching results? That would seem to be unlikely...

src/ngp_algorithms/NodalGradPOpenBoundaryAlg.C
```diff
@@ -0,0 +1,161 @@
...
+
+      const int faceOrdinal = fdata.faceOrd;
+
+      for (int ip = 0; ip < AlgTraits::numFaceIp_; ++ip) {
+        DoubleType pIp = 0.0;
+        if (zeroGrad) {
+          // evaluate pressure at opposing face.  
+          const int oip = meSCS->opposingFace(faceOrdinal, ip);
+          for (int n = 0; n < AlgTraits::nodesPerElement_; ++n) {
+            pIp += e_shape_fcn(oip,n) * elem_p_field(n);
```
sayerhs: @overfelt Sorry, I should have looked closer. This is still a bug, you are accessing `elem_p_field` as a HEX field, but it is registered as a face (QUAD_4) field (See line 65).

unit_tests/ngp_algorithms/UnitTestNodalGradPOpenBoundary.C
overfelt: The question arose as to why a 20x20 test mesh and not 1x1 or 2x2.  I wanted a 3x3 or larger  because in this case the number of exposed faces makes a difference in the answer so I just wanted a variety of element exposed faces to get varied answers at sides, edges and corners.  As Shreyas pointed out below there is still a major bug that is not caught by this unit test.  In general I worried that a single element unit test might not be enough.

unit_tests/ngp_algorithms/UnitTestNodalGradPOpenBoundary.C
overfelt: Yes, it is not needed to initialize the dual nodal volume and exposed area vectors.

unit_tests/ngp_algorithms/UnitTestNodalGradPOpenBoundary.C
overfelt: I think the dual nodal volume was initialized to something besides 2 in the fill_mesh_and_init_fields for the ngp mesh while it was initialized to 2 for the non-ngp unit test and I just wanted to use the same values for both.

unit_tests/ngp_algorithms/UnitTestNodalGradPOpenBoundary.C
overfelt: Yes, that is crud left over from debugging.

unit_tests/ngp_algorithms/UnitTestNodalGradPOpenBoundary.C
overfelt: Shreyas,
Yes, syncing the coordinates back to the host is unnecessary.  I was just syncing all of the fields and included the coordinates even though they started on the host to begin with.
(I wish the email version of these comments included the previous comments)

src/ngp_algorithms/NodalGradPOpenBoundaryAlg.C
overfelt: Shreyas,  Yes, this is a bug.  The pressure field should have been registered as an element gathered field, not a face gathered field.

src/ngp_algorithms/NodalGradPOpenBoundaryAlg.C
```diff
@@ -0,0 +1,162 @@
+/*------------------------------------------------------------------------*/
```
sayerhs: Replace license boilerplate with new joint license text.

src/ngp_algorithms/NodalGradPOpenBoundaryAlg.C
```diff
@@ -0,0 +1,162 @@
...
+    exposedPressureField_   (get_field_ordinal(realm_.meta_data(), (massCorr_ ? "pressure" : "pressure_bc"))),
+    pressureField_ (get_field_ordinal(realm_.meta_data(), "pressure")),
+    gradP_         (get_field_ordinal(realm_.meta_data(), "dpdx")),
+    coordinates_   (get_field_ordinal(realm_.meta_data(), realm.get_coordinates_name())),
+    meFC_  (MasterElementRepo::get_surface_master_element<typename AlgTraits::FaceTraits>()),
+    meSCS_ (MasterElementRepo::get_surface_master_element<typename AlgTraits::ElemTraits>()),
+    faceData_(realm.meta_data()),
+    elemData_(realm.meta_data())
+{
+  const int coordinates = get_field_ordinal(realm.meta_data(), realm.get_coordinates_name());
```
sayerhs: Already have `coordinates_` field as class instance member

unit_tests/UnitTestNodalGradPOpenBoundary.C
```diff
@@ -0,0 +1,101 @@
+/*------------------------------------------------------------------------*/
```
sayerhs: Update license boilerplate with the new joint-license language

unit_tests/ngp_algorithms/UnitTestNodalGradPOpenBoundary.C
sayerhs: @overfelt Because of the dual volume approach, the number of element sub-control faces connected to a node will accumulate different values to an interior node vs a boundary node along an edge or at a corner. However, the contributions to a node per-element should still be  the same.

sayerhs:
@overfelt The code is looking good now. Let us know when you think you're ready for this to merge and then @alanw0 and I can take a look and merge it.

overfelt:
@sayerhs Thanks for your help.

Turns out my cuda build and run was just a gcc run on the cuda platform and I'm still working on an actual cuda build on Sandia's ascicgpu platforms.  Maybe @alanw0 got his build to work?

alanw0:
@overfelt my build worked but not exactly turn-key... I'll send you an email with the "recipe" that I did, and I'll copy it to Shreyas.

sayerhs:
@overfelt @alanw0 I would like to merge this, any objections?

alanw0:
@sayerhs go ahead, I think it's ready.

overfelt:
No objections.

jro

PR #495, opened 2020-01-16
Convert Udiag field update to NGP and other NGP cleanups

sayerhs:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

- Convert Udiag synchronization to NGP, this was not active for `ablNeutralEdge`, but will be for `airfoilRANSEdge` and McAlister wing test cases
- Clean up LinearSystem classes to remove unsued code
- Convert field update logic in `GeometryAlgDriver` and `NodalGradAlgDriver` to NGP versions. 

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [x] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*


src/LowMachEquationSystem.C
```diff
@@ -2719,13 +2720,19 @@ MomentumEquationSystem::assemble_and_solve(
...
+    bool doFinalSyncBackToDevice = false;
+    ngp::parallel_sum(realm_.bulk_data(), fVec, doFinalSyncBackToDevice);
+
     const auto sel = stk::mesh::selectField(*Udiag_)
       & meta.locally_owned_part()
       & !(stk::mesh::selectUnion(realm_.get_slave_part_vector()))
       & !(realm_.get_inactive_selector());
     const auto& bkts = bulk.get_buckets(stk::topology::NODE_RANK, sel);
 
+    // Perform update on host to save on future parallel syncs
```
alanw0: Are we performing this update on host just until we convert the below non-conformal stuff, or is there more to it than that?

sayerhs: @alanw0 Line 2727 calls `parallel_sum` which calls `stk::mesh::parallel_sum` internally, and later we call `copy_owned_to_shared` etc. (lines 2750-2757) which all call `stk::mesh` variants internally. If this loop was performed on device using `ngp::Mesh` then it would have pushed the values to device and then we would have to bring it back up to host again to do the work in lines 2750-2757. So it seems that we are better off performing this on the host so as to avoid `sync_to_host -> sync_to_device -> sync_to_host -> sync_to_device` roundtrip and just do 1 sync back and forth. If you see a better way of doing it, let me know and I'll update it.

PR #507, opened 2020-01-29
Overset: decoupled overset solves and related search/interpolation changes

sayerhs:
This pull-request introduces the ability to select _decoupled overset method_ for the linear solves. This feature is necessary for ECP ExaWind FY20-Q3 milestone performance studies. 

- Implement decoupled solves where no _constraint equations_ are added to the linear system for the fringe rows. Instead the solution is lagged and then updated using an explicit interpolation step.
- Introduce ability to perform _solution exchange_ using the native TIOGA data exchange mechanism. This is the first step to removing the need for ghosting in overset runs. 
- When using decoupled solver methodology for all equation systems, introduce the ability to skip ghosting entirely. 

## Checklist

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [x] MacOS
  - Compilers 
    - [x] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*


sayerhs:
@alanw0 Yes, I should have mentioned. With the default settings (_coupled overset_) the code will not introduce any diffs in the regression tests. I have also tested this with our production runs PhaseVI and NREL-5MW and I have verified that we get the same results with both overset coupling approaches. @ashesh2512 has run several rotating cylinder cases with this change and has verified that independently for that problem. 

For the NREL5MW problem, eliminating ghosting of about 100k entities every timestep offers quite a bit of speed gain per timestep. We also see a drop in the number of linear iterations which provides the largest fraction of the speed gains.

alanw0:
@sayerhs Those speed gains sound great. We're working on performance improvements in stk, including stk ghosting, but there's no substitute for avoiding ghosting altogether.

PR #512, opened 2020-02-04
add  alias for processing all blocks in a mesh

rcknaus:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

The "usability breakout session" during the Q1 meeting requested an alias for specifying all blocks on a mesh, since there can be a lot of them generated with tools like percept.  This adds a special-case for specifying "all_blocks" as the equivalent of listing all element-ranked mesh blocks.  

Another possibility here would instead to have an "omit_targets: " syntax as an alternative way of specifying the target parts of the mesh.  That would better support the use case where there are a lot of physics blocks but a couple of post-processing blocks that the user wants to omit.  I'm not sure which way is better.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -1176,7 +1176,7 @@ Realm::setup_initial_conditions()
   for (size_t j_ic = 0; j_ic < initialConditions_.size(); ++j_ic) {
     InitialCondition& initCond = *initialConditions_[j_ic];
 
-    const std::vector<std::string> targetNames = initCond.targetNames_;
+    const std::vector<std::string> targetNames = handle_all_element_part_alias(initCond.targetNames_);
 
     for (size_t itarget=0; itarget < targetNames.size(); ++itarget) {
       const std::string targetName = physics_part_name(targetNames[itarget]);
@@ -4546,6 +4546,7 @@ Realm::physics_part_name(std::string name) const
 std::vector<std::string>
 Realm::physics_part_names(std::vector<std::string> names) const
 {
+  names = handle_all_element_part_alias(names);
   if (doPromotion_) {
     std::transform(names.begin(), names.end(), names.begin(), [&](const std::string& name) {
       return super_element_part_name(name);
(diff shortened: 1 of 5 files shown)


sayerhs:
Regarding `omit_targets` https://github.com/NaluCFD/Nalu/pull/258 and https://github.com/NaluCFD/Nalu/pull/263 ;)

sayerhs:
While we are at it can we also remove the need to specify it separately twice: once for interior_algorithm and once for material_property?

src/Realm.C
```diff
@@ -4843,6 +4844,27 @@ bool
...
+      }
+    }
+    return new_names;
+  }
+
+  if (std::find(names.begin(), names.end(), allElementPartAlias) != names.end()) {
+    NaluEnv::self().naluOutputP0() << "Part alias " << allElementPartAlias
+        << " present with other parts; " << allElementPartAlias << " must be a valid mesh part" << std::endl;
+  }
+  return names;
```
sayerhs: This brings a question to my mind: Why not let the user leave out `target_name` in the input file and then have `Realm::get_elem_parts_in_mesh()` instead of having this `all_blocks`?

psakievich: Agreed. This could be a nice feature.  It seems logical for IC and materials to apply to the entire domain unless specified.

tonyinme:
hey @rcknaus, this is looking great!
You think we can add some documentation for this feature?

rcknaus:
Added some lines in the documentation talking about the alias.  As for material properties and ICs both requiring targets, I've used different targets there a few times in the past.  It's a pretty concise way to specify discontinuous initial conditions or discontinuous initial fluid properties, and it's not a lot of verbiage in the input file.  I'd prefer to keep it, but I'm OK with getting rid of that too.

PR #514, opened 2020-02-05
Stop accessing operator[] on <simd-type>._data

alanw0:
This unit-test was doing something like:
   simd_value._data[i] = ...

With our coming changes to stk-simd, the '_data' member
won't have operator[]. But the stk-simd type itself has
the operator, so the unit-test code simply needs to use
that:
   simd_value[i] = ...

**Pull-request type:**
- [x ] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
- [x ] Builds successfully (*must test on at least one system/compiler combination*)
  - Operating systems 
    - [x ] Linux
    - [ ] MacOS
  - Compilers 
    - [x ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [ x] Passes all unit tests
- [ x] Passes all regression tests

*For new code updates*

### DIFF
--- a/unit_tests/UnitTestEigenDecomposition.C
+++ b/unit_tests/UnitTestEigenDecomposition.C
@@ -153,12 +153,12 @@ TEST(TestEigen, testeigendecomp3d_simd)
 
   // Initialize matrices
   for (unsigned j = 0; j < stk::simd::ndoubles; ++j) {
-    A3d_fixed_simd[0][0]._data[j] = A3d_fixed[0][0]*(j + 1);
-    A3d_fixed_simd[0][1]._data[j] = A3d_fixed[0][1]*(j + 1);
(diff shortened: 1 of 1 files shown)


sayerhs:
Hmm... there are other parts of the code that have been using 

```C++
stk::simd::get_data(doubleTypeVal, simdIndex);
stk::simd::set_data(doubleTypeVal, simdIndex, doubleVal);
```

Should those be updated to use the `operator[]` too?

alanw0:
@sayerhs I don't think there's really a preference between get_data vs operator[]. In the proposed C++ standard simd type I don't think either will exist and instead there will just be loads and stores. But we have a few years before we are likely to be forced into that.

PR #524, opened 2020-02-17
Fix memory leak in ActuatorDisk

psakievich:
**Pull-request type:**
- [x] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

**1.  Stop creating actuatorPoints every iteration**
- When these redundant points were created one of the node_vec's is not populated/gets a stale pointer
- This leads to an invalid pointer being referenced when the `spread_actuator_force_to_node_vec` method is called.
- Actuator disk regression test now passes clang 7.0 build with asan turned on

**2.  Remove unnecessary re-construction of actuatorMap during initialization**

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/actuator/ActuatorFAST.C
+++ b/src/actuator/ActuatorFAST.C
@@ -435,6 +435,8 @@ ActuatorFAST::initialize()
   // create the ActuatorLineFASTPointInfo
   create_actuator_point_info_map();
 
+  create_point_info_map_class_specific();
+
   // coarse search
   determine_elems_to_ghost();
 
@@ -443,8 +445,6 @@ ActuatorFAST::initialize()
 
   // complete filling in the set of elements connected to the centroid
   complete_search();
-
-  update(); // Update location of actuator points, ghosting etc.
 
 }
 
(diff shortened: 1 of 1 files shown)


sayerhs:
@psakievich Can you also add a brief note about what the memory leak was? While recreating actuator points might have been inefficient, it doesn't explain why it failed Clang ASAN.

psakievich:
@sayerhs I'm not sure exactly where, but one of the node vecs doesn't get populated correctly leading to an Assert_Throw in stk when the entity type is compared between the bucket entity and the node_vec entry's entity.  Could be a stale pointer.  I don't know exactly where this occurred, but I believe it is because the old set of actuatorDisk points never have there node_vec's cleared.

sayerhs:
Thanks! Since we are using similar methods in actuator line code also, I was wondering why it didn't affect that side of the code.

PR #525, opened 2020-02-18
Jroverf/sst max length scale

overfelt:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [ X] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [X] Linux
    - [ ] MacOS
  - Compilers 
    - [X] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [X] NVIDIA CUDA
- [X] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/ngp_algorithms/SSTMaxLengthScaleDriver.C
+++ b/src/ngp_algorithms/SSTMaxLengthScaleDriver.C
@@ -48,21 +48,20 @@ void SSTMaxLengthScaleDriver::post_work()
   const auto& meshInfo = realm_.mesh_info();
   auto& ngpMaxLengthScale = nalu_ngp::get_ngp_field(meshInfo, "sst_max_length_scale");
 
-  std::vector<NGPDoubleFieldType*> fields(1, &ngpMaxLengthScale);
+  const auto& meta = realm_.meta_data();
+  auto* maxLengthScale = meta.template get_field<ScalarFieldType>(
+     stk::topology::NODE_RANK, "sst_max_length_scale");
 
   // Algorithms should have marked the fields as modified, but call this here to
   // ensure the next step does a sync to host
   ngpMaxLengthScale.modify_on_device();
+  ngpMaxLengthScale.sync_to_host();
 
-  bool doFinalSyncToDevice = false;
-  ngp::parallel_sum(realm_.bulk_data(), fields, doFinalSyncToDevice);
+  stk::mesh::parallel_max(realm_.bulk_data(), {maxLengthScale});
 
   if (realm_.hasPeriodic_) {
-    const auto& meta = realm_.meta_data();
-    auto* maxLengthScale = meta.template get_field<ScalarFieldType>(
-       stk::topology::NODE_RANK, "sst_max_length_scale");
     const unsigned nComponents = 1;
-    realm_.periodic_field_update(maxLengthScale, nComponents);
+    realm_.periodic_field_max(maxLengthScale, nComponents);
   }
   ngpMaxLengthScale.modify_on_host();
   ngpMaxLengthScale.sync_to_device();
--- a/src/ngp_algorithms/SSTMaxLengthScaleDriver.C
+++ b/src/ngp_algorithms/SSTMaxLengthScaleDriver.C
@@ -1,0 +1,71 @@
+// Copyright 2017 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS), National Renewable Energy Laboratory, University of Texas Austin,
+// Northwest Research Associates. Under the terms of Contract DE-NA0003525
+// with NTESS, the U.S. Government retains certain rights in this software.
+//
+// This software is released under the BSD 3-clause license. See LICENSE file
+// for more details.
+//
+
+#include "ngp_algorithms/SSTMaxLengthScaleDriver.h"
+#include "ngp_utils/NgpLoopUtils.h"
+#include "ngp_utils/NgpFieldUtils.h"
+#include "Realm.h"
+#include "utils/StkHelpers.h"
+
+#include "stk_mesh/base/Field.hpp"
+#include "stk_mesh/base/FieldParallel.hpp"
+#include "stk_mesh/base/MetaData.hpp"
+#include "stk_mesh/base/FieldBLAS.hpp"
+#include "stk_ngp/NgpFieldParallel.hpp"
+
+namespace sierra {
+namespace nalu {
+
+SSTMaxLengthScaleDriver::SSTMaxLengthScaleDriver(
+  Realm& realm
+) : NgpAlgDriver(realm)
+{}
+
+void SSTMaxLengthScaleDriver::pre_work()
+{
+
+  auto* maxLengthScale = realm_.meta_data().template get_field<ScalarFieldType>(
+    stk::topology::NODE_RANK, "sst_max_length_scale");
+  stk::mesh::field_fill(0.0, *maxLengthScale);
+  
+  const auto& meshInfo = realm_.mesh_info();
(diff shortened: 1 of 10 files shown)


src/ngp_algorithms/SSTMaxLengthScaleDriver.C
```diff
@@ -0,0 +1,70 @@
...
+  ngp::parallel_sum(realm_.bulk_data(), fields, doFinalSyncToDevice);
+
+  if (realm_.hasPeriodic_) {
+    const auto& meta = realm_.meta_data();
+    auto* maxLengthScale = meta.template get_field<ScalarFieldType>(
+       stk::topology::NODE_RANK, "sst_max_length_scale");
+    const unsigned nComponents = 1;
+    realm_.periodic_field_update(maxLengthScale, nComponents);
+  }
+
```
sayerhs: We need to push the fields back to the device after this point 

```C++
ngpMaxLengthScale.modify_on_host();
ngpMaxLengthScale.sync_to_device();
```

unit_tests/ngp_algorithms/UnitTestSSTMaxLengthScaleAlg.C
```diff
@@ -0,0 +1,63 @@
...
+  unit_test_utils::HelperObjects helperObjs(
+    bulk_, stk::topology::HEX_8, 1, partVec_[0]);
+
+  // Force computation of edge area vector
+  helperObjs.realm.realmUsesEdges_ = true;
+
+  const auto& fieldMgr = helperObjs.realm.mesh_info().ngp_field_manager();
+  auto& ngpElemVol = fieldMgr.get_field<double>(
+      elementVolume_->mesh_meta_data_ordinal());
+
```
sayerhs: Looks like this is leftover from cut & paste from another test, and is not needed?

unit_tests/ngp_algorithms/UnitTestSSTMaxLengthScaleAlg.C
```diff
@@ -0,0 +1,63 @@
...
+  sierra::nalu::SSTMaxLengthScaleDriver AlgDriver(helperObjs.realm);
+
+  AlgDriver.register_elem_algorithm<sierra::nalu::SSTMaxLengthScaleAlg>(
+    sierra::nalu::INTERIOR, partVec_[0], "SSTMaxLen");
+
+  AlgDriver.execute();
+
+  ngpElemVol.modify_on_device();
+  ngpElemVol.sync_to_host();
+
```
sayerhs: Same here.

overfelt: Yes, these should both be maxLengthScale.  I'm surprised this passed on the GPU.  I need to check that I really ran on the GPU.

sayerhs: The test passed because `post_work` pulls the field back onto host and weren't modified afterward.

src/ngp_algorithms/SSTMaxLengthScaleDriver.C
```diff
@@ -64,7 +64,8 @@ void SSTMaxLengthScaleDriver::post_work()
     const unsigned nComponents = 1;
     realm_.periodic_field_update(maxLengthScale, nComponents);
   }
-
+  ngpMaxLengthScale.modify_on_host();
```
alanw0: Is this because realm.periodic_field_update modified on host? I forgot why we can't do the ngp periodic_field_update yet... In any case, shouldn't this be inside the if-statement if(realm.has_periodic)?

sayerhs: @alanw0 We need a forced push to device because `doFinalSyncToDevice = false` (in Line 57). The reason to do it this way is because we do `parallel_sum` followed by `periodic_field_update`. Then the operations would be:

```
- Sync from device to host
- Perform parallel_sum
- Sync from host to device 
- If (periodic)
  - Sync from device to host
  - perform periodic exchange
  - Sync from host to device
```

With the above logic we eliminate the intermediate `host -> device -- device -> host` round-trip. It is not too bad here, but there are situations where we do 

```
parallel_sum
copy_owned_to_shared
periodic_field_update
parallel_communicate_field(nonConformalGhosting_)
parallel_communicate_field(oversetGhosting_)
```

alanw0: I agree we don't want the intermediate copies back and forth. Can we switch to using the ngp periodic_field_update? I forgot why we couldn't switch to that.

sayerhs: It would be nice if `ngp::Mesh/ngp::Field` could provide a 

```C++
bool fromDevice = true;
bool fromHost = false;
ngp::field_sync_begin(fromDevice, fromHost);
{
  parallel_sum
  copy_owned_to_shared
  parallel_communicate_field(ghosting)
}
ngp::field_sync_end()
```

alanw0: I'm not clear on what would happen within field_sync_begin and field_sync_end. Currently, parallel_sum etc bring the field up from device only if needed, and put it back to device only if directed by the flag. I think that allows for minimal copies. What scenario am I missing?

sayerhs: @alanw0 It would parallel `bulk.modification_begin() / bulk.modification_end()` and do the necessary synchronizations. It is just a short-hand for remembering to toggle flags... but yes, it doesn't have anything extra than what we can do with manual passing of flags to individual functions. 

I assume you mean `ngp_apply_constraints`? If so, yes, in this case it will work. However, in situations where we have extra ghostings to handle, we will have to add another flag to prevent `ngp_apply_constraints` from syncing back to device.

src/ngp_algorithms/SSTMaxLengthScaleDriver.C
```diff
@@ -0,0 +1,71 @@
...
+  auto& ngpMaxLengthScale = nalu_ngp::get_ngp_field(meshInfo, "sst_max_length_scale");
+
+  std::vector<NGPDoubleFieldType*> fields(1, &ngpMaxLengthScale);
+
+  // Algorithms should have marked the fields as modified, but call this here to
+  // ensure the next step does a sync to host
+  ngpMaxLengthScale.modify_on_device();
+
+  bool doFinalSyncToDevice = false;
+  ngp::parallel_sum(realm_.bulk_data(), fields, doFinalSyncToDevice);
```
sayerhs: @overfelt This should be `ngp::parallel_max` we want the maximum value not sum for the max length scale.

src/ngp_algorithms/SSTMaxLengthScaleDriver.C
```diff
@@ -0,0 +1,71 @@
...
+
+  if (realm_.hasPeriodic_) {
+    const auto& meta = realm_.meta_data();
+    auto* maxLengthScale = meta.template get_field<ScalarFieldType>(
+       stk::topology::NODE_RANK, "sst_max_length_scale");
+    const unsigned nComponents = 1;
+    realm_.periodic_field_update(maxLengthScale, nComponents);
```
sayerhs: Similarly this should use https://github.com/Exawind/nalu-wind/blob/549aa89599428f2e14a500f05eab71b3b5504829/src/PeriodicManager.C#L899-L902 

see here for an example:

https://github.com/Exawind/nalu-wind/blob/549aa89599428f2e14a500f05eab71b3b5504829/src/ShearStressTransportEquationSystem.C#L511-L515

PR #526, opened 2020-02-18
Update SST equation system to NGP 

sayerhs:
This pull-request updates the field update methods in SST equation system to NGP. This is the last bit of code that needs to be ported to GPUs before RANS simulations of airfoils/fixed-wings on GPUs can begin. 

The changes introduced a diff in one of the ~10 SST tests:

```
FAIL: elemBackStepLRSST_Input.................    41.8989s 1.3003e-07 4.6521e-09
```

I think this is acceptable and have re-blessed the gold file. 

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [x] MacOS
  - Compilers 
    - [x] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [x] NVIDIA CUDA
- [x] Passes all unit tests


src/ShearStressTransportEquationSystem.C
```diff
@@ -411,10 +434,33 @@ ShearStressTransportEquationSystem::update_and_clip()
 void
 ShearStressTransportEquationSystem::clip_min_distance_to_wall()
 {
-  // if this is a restart, then min distance has already been clipped
-  if (realm_.restarted_simulation())
+  if (realm_.restarted_simulation() && !realm_.has_mesh_motion())
     return;
 
+#if 1
```
alanw0: Did you mean to leave `#if 1` in here?

sayerhs: @alanw0 Seems like github showed you an older version of the commits? 
0e62f58 should have removed those `#if 1` statements.

alanw0: Hmm. That's weird. In any case, it's fine then. :-)

PR #530, opened 2020-02-19
Rebless ActuatorDisk test

psakievich:
Rebless after fixing memory leak in PR #524


jrood-nrel:
Thanks Phil. I just updated this with the gold file from the test machine to make sure it's exact. So I hope you don't mind if I just close this.

PR #531, opened 2020-02-19
Make pyramid isInElement more robust

rcknaus:
**Pull-request type:**
- [x] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

The pyramid basis functions are well-defined within the element but are singular on a surface outside the element.  During the newton solve in isInElement, the point may be outside of the element and near the singular surface, causing the inversion to produce a NAN or similar bad things to occur.

This tries to make the newton solve more robust, by increasing the tolerance on the `(1-t)` denominator when t approaches 1, providing an early out when the point is actually found, and providing an earlier failure mechanism when the Jacobian is zero.

Also activates the master element unit tests to pyramids, which haven't needed to be skipped in a while.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*

### DIFF
--- a/unit_tests/UnitTestMasterElements.C
+++ b/unit_tests/UnitTestMasterElements.C
@@ -13,6 +13,7 @@
 #include <master_element/MasterElement.h>
 #include <master_element/MasterElementFactory.h>
 #include <master_element/Quad42DCVFEM.h>
+#include <master_element/Pyr5CVFEM.h>
 #include <master_element/TensorOps.h>
 
 #include <memory>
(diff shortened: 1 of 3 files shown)


unit_tests/UnitTestMasterElements.C
```diff
@@ -610,14 +616,14 @@ protected:
     TEST_F(x, hex8##_##y)   { y(stk::topology::HEX_8); }
```
psakievich: Delete these stale comments porfavor.

unit_tests/UnitTestMasterElements.C
```diff
@@ -610,14 +616,14 @@ protected:
...
-TEST_F_ALL_TOPOS_NO_PYR(MasterElement, scs_interpolation)
-TEST_F_ALL_TOPOS_NO_PYR(MasterElement, scs_derivative)
-TEST_F_ALL_TOPOS_NO_PYR(MasterElement, scv_interpolation)
-TEST_F_ALL_TOPOS_NO_PYR(MasterElement, volume_integration)
+TEST_F_ALL_TOPOS(MasterElement, scs_interpolation)
+TEST_F_ALL_TOPOS(MasterElement, scs_derivative)
+TEST_F_ALL_TOPOS(MasterElement, scv_interpolation)
+TEST_F_ALL_TOPOS(MasterElement, volume_integration)
 
 // Pyramid fails since the reference element
```
psakievich: Ditto

PR #544, opened 2020-03-19
Improve unit-testing for sparse-matrix values.

alanw0:
It used to just compare each matrix column and coefficient to
the expected gold values, expecting position to match as well.
Now it searchs the matrix row for each gold column and checks
if the coefficient matches, and doesn't fail if there are extra
columns in the matrix.

**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*

### DIFF
--- a/unit_tests/UnitTestTpetraHelperObjects.h
+++ b/unit_tests/UnitTestTpetraHelperObjects.h
@@ -109,10 +109,16 @@ struct TpetraHelperObjectsBase {
 
     for(int i=0; i<localMatrix.numRows(); ++i) {
       KokkosSparse::SparseRowViewConst<MatrixType> constRowView = localMatrix.rowConst(i);
-
-      for(int j=0; j<constRowView.length; ++j) {
-        EXPECT_EQ(cols[rowOffsets[i]+j], constRowView.colidx(j));
-        EXPECT_NEAR(vals[rowOffsets[i]+j], constRowView.value(j), 1.e-14)<<"i: "<<i<<", j: "<<j;
+      for(int offset=rowOffsets[i]; offset<rowOffsets[i+1]; ++offset) {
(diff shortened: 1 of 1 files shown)


sayerhs:
@alanw0 @jhux2 brings up a good point, may be there should be a check to ensure that the non-gold entries are near zero?

alanw0:
@jhux2 @sayerhs yes that is a good point, I'll add a commit to include that check.

unit_tests/UnitTestTpetraHelperObjects.h
```diff
@@ -117,6 +130,11 @@ struct TpetraHelperObjectsBase {
             foundGoldCol = true;
             EXPECT_NEAR(vals[offset], constRowView.value(j), 1.e-14)<<"i: "<<i<<", j: "<<j;
           }
+          else if (!find_col(constRowView.colidx(j),
```
jhux2: @alanw0 Shouldn't this be `if (find_col(...`?

alanw0: Nope, it's checking the case where the column from the matrix is not found in the "gold" cols. In other words the column from the matrix is one of the "extra" graph columns which should correspond to a zero coefficient value.

jhux2: Ah, ok, got it.  `cols` are the "gold" columns.

PR #546, opened 2020-03-20
Fix bugs in production actuator code

psakievich:
A couple of bugs in production actuator code found during refactor of the actuator code

 - Unit tensor was wrong for actuator lines (affects hub and tower)

 - Actuator disk swept points were being created twice doubling the
   force added to the flow field and the computational work for the
   swept points of the actuator disk

Please note that this will likely require a rebless of all actuator regression tests. The actuator line tests passed with a loose tolerance on my local machine but will likely diff on production platforms. The actuator disk test failed, but the calculations with this fix are correct and match the results of the new code that is being developed for the HFM Q2 milestone.

**Pull-request type:**
- [x] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [x] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*


tonyinme:
hey @psakievich, I think that changing the tensor will change the blade orientation in the wrong direction. Did you check that after making this change?

src/actuator/ActuatorLineFAST.C
```diff
@@ -103,7 +103,7 @@ ActuatorLineFAST::execute_class_specific(
     // inverted after the rotation is done inside the 
     // spread_actuator_force_to_node_vec function.
     std::vector<double> orientation_tensor
-        {0.0, 1.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 1.0};
+        {1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0};
```
gantech: @tonyinme Wasn't this done on purpose? I thought it had something to do with the right-handedness of the orientation tensor inside OpenFAST vs. how you used them here for the direction vector along chord, thickness and span wise direction.

psakievich: I didn’t check this I just saw it was this way in the actuator disk and the comments. I’ll check it and revert if necessary

psakievich: @tonyinme @gantech I've changed it so the tensor out of openfast gets set to `(chord, thick, span)` instead of trying to change all the other tensors to match openfast.
@tonyinme will you run this with one of your advanced actuator line problems to see if the results come out correct?  So far we've just got uniform gaussians in the reg tests.  We should add a nonuniform gaussian that takes advantage of chord scaling to the regression tests as well

psakievich: row swap was the wrong operation.  @tony and I have verified that the latest commits provide correct

psakievich:
@tonyinme @gantech do you think this is ready to merge?  There is an issue with `epsilon_tower` when using the anisotropic gaussian that I've identified, but that seems to be a long standing bug.  I'm working on fixing it in the new code.

gantech:
Thanks. I'll leave it to @tonyinme I don't know enough about the actuator disk stuff. If you've verified the orientation tensor to take the conventions in OpenFAST into account, then I'm good.

psakievich:
Addressed in #550

PR #563, opened 2020-04-08
Convert nalu-wind to use new stk::mesh::NgpMesh and stk::mesh::NgpField API

djglaze:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

Convert nalu-wind to use new stk::mesh::NgpMesh and stk::mesh::NgpField API

nalu_ngp::FieldManager was added as a lightweight interface for the old ngp::FieldManager workflow.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [x] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [x] NVIDIA CUDA
- [ ] Compiles without warnings (pre-existing Kokkos warnings)
- [x] Passes all unit tests

*For new code updates*

### DIFF
--- a/unit_tests/kernels/UnitTestKernelUtils.C
+++ b/unit_tests/kernels/UnitTestKernelUtils.C
@@ -15,7 +15,8 @@
 
 #include <stk_util/parallel/Parallel.hpp>
 #include <stk_mesh/base/FieldParallel.hpp>
-#include <stk_ngp/NgpFieldManager.hpp>
+#include <stk_mesh/base/NgpMesh.hpp>
+#include <ngp_utils/NgpFieldManager.h>
 
 #include <cmath>
 
@@ -608,9 +609,9 @@ void calc_mass_flow_rate_scs(
   const VectorFieldType& velocity,
   const GenericFieldType& massFlowRate)
 {
-  using Traits = sierra::nalu::nalu_ngp::NGPMeshTraits<ngp::Mesh>;
+  using Traits = sierra::nalu::nalu_ngp::NGPMeshTraits<stk::mesh::NgpMesh>;
   using Hex8Traits = sierra::nalu::AlgTraitsHex8;
-  using ElemSimdData = sierra::nalu::nalu_ngp::ElemSimdData<ngp::Mesh>;
+  using ElemSimdData = sierra::nalu::nalu_ngp::ElemSimdData<stk::mesh::NgpMesh>;
 
   const auto& meta = bulk.mesh_meta_data();
   const int ndim = meta.spatial_dimension();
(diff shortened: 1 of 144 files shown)


unit_tests/kernels/UnitTestKernelUtils.h
```diff
@@ -547,12 +547,32 @@ class MomentumABLKernelHex8Mesh : public MomentumKernelHex8Mesh
     const double bcVel[3] = {0.0, 0.0, 0.0};
     MomentumKernelHex8Mesh::fill_mesh_and_init_fields(doPerturb, generateSidesets);
     stk::mesh::field_fill_component(vel, *velocity_);
+    velocity_->modify_on_host();
+    velocity_->sync_to_device();
+
```
sayerhs, line 552: Question regarding new NGP field behavior: does it automatically pull the field to device on first creation of NgpField? Or do we have to explicitly do this from now on?

unit_tests/kernels/UnitTestKernelUtils.C
```diff
@@ -1376,7 +1377,7 @@ void calc_projected_nodal_gradient_boundary(
       auto v_dnv = preReqData.get_scratch_view_1D(dnv);
       auto v_vector = preReqData.get_scratch_view_2D(vectorField);
       auto v_scs_areav = preReqData.get_me_views(sierra::nalu::CURRENT_COORDINATES).scs_areav;
-      const ngp::Mesh::ConnectedNodes node_rels = preReqData.elemNodes;
+      const stk::mesh::NgpMesh::ConnectedNodes node_rels = preReqData.elemNodes;
```
sayerhs, line 1380: I am really liking `auto` right now :)

unit_tests/kernels/UnitTestKernelUtils.h
djglaze: That behavior is unchanged.  The first time an NgpField is requested, it will be created on device and synced against the current stk Field data.

There is a behavior change from the old ngp::FieldManager to the new nalu-specific NgpFieldManager, however.  The old ngp::FieldManager would tear down the NgpFields when it was destructed, so that they would be re-created (and implicitly synced) when requested again.  The new NgpFieldManager lets the fields persist past its own lifetime.

Some of the unit tests were implicitly relying on multiple creations/destructions of the old ngp::FieldManager, which gave an implicit sync to the device before it was needed in the kernel being tested.  Now that the NgpFieldManager lets NgpFields persist on Device, we needed to explicitly sync a few things.

unit_tests/kernels/UnitTestKernelUtils.C
djglaze: Normally I'm against auto, but you might be right on this one.  :-)

PR #569, opened 2020-04-17
Move to GPU versions of gradient kernels.

overfelt:
There was a problem with mixing GPU kernels
and CPU gradient contributions since the
sum into is then in different memory spaces.

This seems to fix much of the difference
between the CPU and GPU results for the
channel flow test we have been using to
debug this.

There are still differences in the results.
A deep dive on the contributions for a single
entry in the right hand side of the momentum
solve shows the order of operations can have
an impact since there are values summed in
of O(10^3) that almost cancel out in the final
result of O(10^-5).

**Pull-request type:**
- [ X] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
- [ X] Builds successfully (*must test on at least one system/compiler combination*)
  - Operating systems 
    - [X ] Linux
    - [ ] MacOS
  - Compilers 
    - [X ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ X] NVIDIA CUDA
- [ ] Passes all unit tests
There are some diffs in the tests on the Sandia ASCICGPU platforms.  But this seems to
be due to order of operations differences.

*For new code updates*

### DIFF
--- a/src/EnthalpyEquationSystem.C
+++ b/src/EnthalpyEquationSystem.C
@@ -377,7 +377,7 @@ EnthalpyEquationSystem::register_interior_algorithm(
       nodalGradAlgDriver_.register_edge_algorithm<ScalarNodalGradEdgeAlg>(
         algType, part, "enthalpy_nodal_grad", &enthalpyNp1, &dhdxNone);
     else
-      nodalGradAlgDriver_.register_legacy_algorithm<AssembleNodalGradElemAlgorithm>(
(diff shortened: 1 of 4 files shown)


overfelt:
I've only run this on channel0.yaml, the base case.  I'll move on up the channel[1234].yaml today.

jro

alanw0:
This is good debugging James. Last night I was putting in print statements trying to find out whether fields were being consistently synced before being accessed etc., but I was a long way from finding this.

PR #575, opened 2020-04-30
Performance optimization of sample plane output

lawrenceccheung:
**Pull-request type:**
- [ ] Bug fix 
- [X] Documentation update
- [X] Feature enhancement

Based on discussions with @psakievich, @alhs6577, and @rcknaus, it was found that the current sample plane output incurs a significant performance penalty when writing out very large planes at very high frequency.  I've made some changes which should increase the performance of the plane outputs and reduce the size of the output files.

Here is a quick summary of the feature updates:
-	The plane outputs can now be written out in a compressed gzip format, which shrinks the files and decreases the amount of data that needs to be written.
-	The coordinates and indices of the plane point can be written out in a separate file, so they don’t need to be included with every write.
-	You can have a separate plane file for each variable.  This allows the work to be spread out over more processors.

Based on speed tests from @alhs6577, sample plane output time dropped from 1.23 sec to 0.092 sec per iteration due to these changes.

Code changes are mostly contained within `DataProbePostProcessing.C`, with a minor addition to `CMakeLists.txt` for adding `boost::iostreams` libraries.  The corresponding documentation has also been updated.  Some changes to people's postprocessing scripts may also be required.

Lawrence

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [X] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [X] Intel compilers
    - [ ] NVIDIA CUDA
- [X] Passes all unit tests

*For new code updates*
- [X] Documentation updates - additions to user, theory, and verification manuals

### DIFF
--- a/src/DataProbePostProcessing.C
+++ b/src/DataProbePostProcessing.C
@@ -164,7 +164,10 @@ DataProbePostProcessing::load(
       }
       NaluEnv::self().naluOutputP0() << "DataProbePostProcessing::Adding "<<formatName<<" output format..." << std::endl;
     }
+    // Enable performance timings of output
+    get_if_present(y_dataProbe, "time_performance", enablePerfTiming_, enablePerfTiming_);
 
+    // Optional speed-up parameters
     get_if_present(y_dataProbe, "write_coords", writeCoords_, writeCoords_);
     get_if_present(y_dataProbe, "gzip_level",   gzLevel_,     gzLevel_);
 
@@ -404,8 +407,8 @@ DataProbePostProcessing::load(
 	    else
 	      probeInfo->offsetSpacings_[iplane+offset].push_back(0.0);
 
-            // string: onlyOutputField
-            const YAML::Node onlyOutputField = y_planenode["only_output_field"];
+	    // string: onlyOutputField
+	    const YAML::Node onlyOutputField = y_planenode["only_output_field"];
 	    if (onlyOutputField)
 	      probeInfo->onlyOutputField_[iplane+offset] = onlyOutputField.as<std::string>() + "_probe";
 	    else
@@ -882,24 +885,24 @@ DataProbePostProcessing::execute()
   }
 
   if ( isOutput ) {
-    const double t1 = NaluEnv::self().nalu_time();  
+    const double t1 = enablePerfTiming_? NaluEnv::self().nalu_time() : 0.0;  
     // execute and provide results...
     transfers_->execute();
-    const double t2 = NaluEnv::self().nalu_time();  
+    const double t2 = enablePerfTiming_? NaluEnv::self().nalu_time() : 0.0; 
     if (useExo_) {
       provide_output_exodus(currentTime);
     }
     if (useText_) {
       provide_output_txt(currentTime);
     }
-    const double t3 = NaluEnv::self().nalu_time();  
-    // Uncomment below to provide timing statistics
-    // NaluEnv::self().naluOutputP0() << "DataProbePostProcessing::execute " 
-    // 				   << " transfer_time: "<<t2-t1
-    // 				   << " output_time: "<<t3-t2      
-    // 				   << " total_time: "<<t3-t1
-    // 				   << std::endl;
-  
+    const double t3 = enablePerfTiming_? NaluEnv::self().nalu_time() : 0.0; 
+    if (enablePerfTiming_) 
+      NaluEnv::self().naluOutputP0() << "DataProbePostProcessing::execute " 
+				     << " transfer_time: "<<t2-t1
+				     << " output_time: "<<t3-t2      
+				     << " total_time: "<<t3-t1
+				     << std::endl;
+    
   }
 }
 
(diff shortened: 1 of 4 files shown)


sayerhs:
@lawrenceccheung Given Trilinos is actively trying to remove dependence on boost, it seems unfortunate that we are bringing in more boost dependency into `nalu-wind`. 

Why do we keep cramming stuff into DataProbePostProcessing? We have already bolted on surface into that class and now we are pushing forward on a non-scalable choice of output format as well? Why not adopt a saner format, i.e., NetCDF, or for that matter just exodus and support multiple exo outputs in nalu-wind? Were these options considered? If so, why were they rejected?

lawrenceccheung:
Hi @sayerhs, a lot the suggested changes were driven by the demands of postprocessing wake planes for certain projects here.  I don't have the full details on everything that was tried, but I think exodus files and paraview catalyst options were considered as inputs for the postprocessing.  High frequency exodus output might have produced files which are too large to handle, and there might have been other issues with catalyst.  Outputting all the planes as line probes scale well, but are unwieldy to specify in the input or generated too many small files to assemble, so planes might have been the compromise solution.  Alan (@alhs6577) or @psakievich may be able to say more regarding the workflow or postprocessing decisions.

I agree with the comment on `boost` libraries, it would be nice to have things be more self-contained in the future.  There's already `boost` dependencies in DataProbePostProcessing, but I wouldn't mind if we created an independent set of utilities (or use C++17 calls) that Nalu-wind can rely on for I/O without linking to `boost`.

Lawrence

src/DataProbePostProcessing.C
```diff
@@ -394,6 +404,13 @@ DataProbePostProcessing::load(
 	    else
 	      probeInfo->offsetSpacings_[iplane+offset].push_back(0.0);
 
+            // string: onlyOutputField
+            const YAML::Node onlyOutputField = y_planenode["only_output_field"];
+	    if (onlyOutputField)
```
rcknaus, line 409: whitespace

src/DataProbePostProcessing.C
```diff
@@ -865,14 +882,24 @@ DataProbePostProcessing::execute()
...
     transfers_->execute();
+    const double t2 = NaluEnv::self().nalu_time();  
     if (useExo_) {
       provide_output_exodus(currentTime);
     }
     if (useText_) {
       provide_output_txt(currentTime);
     }
+    const double t3 = NaluEnv::self().nalu_time();  
+    // Uncomment below to provide timing statistics
```
rcknaus, line 896: Delete or add an input parameter to turn this on

src/DataProbePostProcessing.C
```diff
@@ -992,18 +1019,28 @@ DataProbePostProcessing::provide_output_txt(
...
 	      const int N1 = probeInfo->edge1NumPoints_[inp];
 	      const int N2 = probeInfo->edge2NumPoints_[inp];
 	      const int pointsPerPlane = N1*N2;
 
+	      // Use gzip compression when writing
+	      boost::iostreams::filtering_streambuf<boost::iostreams::output> outbuf;
+	      if ((0<gzlevel)&&(gzlevel<10)) {
+		fileName = fileName+".gz";
+		outbuf.push(boost::iostreams::gzip_compressor(
+		//boost::iostreams::gzip_params(gzlevel)));
```
rcknaus, line 1040: remove

lawrenceccheung:
Thanks @rcknaus.  Yeah, definitely in the future I'd like to have a separate `write_compressed_file` function, along with some other file utilities, which could be used in other places.  That will help remove the dependence on boost utilities

sayerhs:
@lawrenceccheung I am assuming that you are not planning to implement @rcknaus 's suggestion as part of this pull-request? If so, are you ready for this to be merged? 

Also, can you please add a comment regarding how you plan to implement some of the things that was discussed in May 1 meeting that @michaelasprague organized?

src/DataProbePostProcessing.C
lawrenceccheung: Not sure what happened with the whitespace there, but it should look better now, at least on my branch.

lawrenceccheung:
Adding a comment to recap the meeting on May 1 with @michaelasprague, @sayerhs, @rcknaus, @alhs6577, and @dcmania.

Based on the discussion, we agreed to
* Merge in the current pull request 
* Plan to revamp the current DataProbePostProcessing code to have better functionality and cleaner structures. The revamped code would
    1. Remove dependencies on boost libraries -- create separate utility functions as necessary
    2. Retain text file output (for debugging purposes), but introduce better standardized formats (like netcdf)
    3. Allow for parallel I/O output to increase performance
    4. Align this part of code with other NGP sections being written (e.g. actuator line code)

Also, after discussions with AMR-wind team, I feel that the plane output format and code should be synced between the two codes.  When running overset simulations, the sample planes should be able to seamlessly cut through both domains and provide output in an integrated fashion.  This will require coordination with the AMR-wind development.

lawrenceccheung:
Hi @sayerhs,
Yes, the code should be ready to merge now.  Our clusters are back online now, and I implemented and tested Rob's suggested changes in the PR.  I'm planning to implement the write compressed file function as a separate task, probably along with some other file utilities.

I added a comment to the PR above which documents what we discussed during the meeting.  Let me coordinate with @marchdf and others on the new sampling plane code.  We can probably use the same libraries and approach to keep everything in sync.  Hopefully we can present a unified plan soon that will work for both codes.

Lawrence

PR #587, opened 2020-05-14
modify mesh motion module terminology 

ashesh2512:
**Pull-request type:**
- [ ] Bug fix 
- [x] Documentation update
- [x] Feature enhancement

This pull request modifies the mesh motion terminology. It gets rid of the `non-inertial` and `inertial` terminology. It has been brought to attention pointed on numerous occasions that the `inertial` frame isn't really inertial (constant velocity) and the `non-inertial` frame isn't really just non-inertial (e.g. one could use translation at constant velocity). What they really are, is `reference` frame  which should only be a one time motion and `moving` frame . If only reference frame is selected, the mesh motion flag (as seen by the rest of the code) is set to false to avoid re-initialization of linear system, overset, etc.. The reference frame will essentially be a one time displacement which updates the `coordinates` and not the `current_coordinates`. 
Note: This change in terminology changes the input files, which is worth the change in choice of terminology as it has caused a lot of confusion to people who have had to deal with it. I have also updated the documentation to reflect the new terminology. 

EDIT:  Following some discussion, this pull request introduces the concept of `mesh_transformation` which is essentially a one-time mesh motion. `meshMotionAlg` now handles only time-dependent mesh motions. The concept of frames has been removed from the input files, and the older input files that were not using `inertial` frames still work. The regression test input files, unit tests, and documentation have been updated too.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [x] MacOS
  - Compilers 
    - [x] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Compiles without warnings - This pull request did not lead to additional warnings.
- [x] Passes all unit tests


sayerhs:
Regarding the input file, would it make more sense to move the one-time mesh transform operation out of `mesh_motion` section and adding a `mesh_transformations:` section that performs the operations?

ashesh2512:
@sayerhs Your suggestion makes sense. I am thinking implementation wise what do you have in mind ? 

1) I could create a class ‘meshTransformAlg’ that mimics ‘meshMotionAlg’ but acts only on reference frames. The section ‘mesh_transformations’ would activate ‘meshTransformAlg’.

2) In addition to 1), do you think it would be a good idea to get rid of the concept of frames in the input file, such that everything under ‘mesh_motion’ is a moving frame and everything under everything under ‘mesh_transformations’ is a reference frame ?

ashesh2512:
@sayerhs I have made some more changes along the lines of the comment -
https://github.com/Exawind/nalu-wind/pull/587#issuecomment-629598432

I have also updated the description of the pull request. The older input files should still work since the concept of frames has been eliminated and the current tests should be unaffected. However I still have to run the regression tests as a sanity check. I will update once I am able to do so.

src/mesh_motion/FrameBase.C
```diff
@@ -67,9 +65,11 @@ void FrameBase::load(const YAML::Node& node)
...
 void FrameBase::populate_part_vec(const YAML::Node& node)
 {
-  // if nor parts specified and frame is inertial, return
-  if (!node["mesh_parts"] && isInertial_)
-    return;
+  // if no parts specified and frame is reference, return
+  if (!node["mesh_parts"]) {
+    throw std::runtime_error(
+      "FrameBase: No mesh parts found.");
+  }
```
sayerhs, line 72: Does it make sense to allow user to transform the entire mesh if this is a one-time transformation action?

ashesh2512: @sayerhs This part of the code is common to both `mesh_transformation` and `mesh_motion`. My approach was to make sure that the user was more "aware" of what they are trying to move, but I can see how it might be a nuisance if there are a lot of parts involved and someone just wants to move all of them a certain way. I will change this, so let's not merge until I push the commit to allow this.

sayerhs: @ashesh2512 Take a look at what @rcknaus had done for interior parts that allows user to specify all the parts with a single entry in the input file. May be something like that can be done here too that strikes the right balance between user convenience vs. error checking? 

https://github.com/Exawind/nalu-wind/blob/f539ccd12b4ce76e19382f34a53f9f2e3aa19a77/include/Realm.h#L691

ashesh2512: @sayerhs Thank you for the suggestions. I have added a commit to introduce this flexibility - 
https://github.com/Exawind/nalu-wind/pull/587/commits/2bf4c949c5c30a52b0daf02204377415fdde69f6

PR #591, opened 2020-05-16
Prepare overset interface for NGP 

sayerhs:
This pull-request prepares the overset interface for NGP transition. Overset connectivity is still performed via TIOGA, but on the nalu-wind side it now performs necessary field sync (that will go away as more pieces are moved to GPU), and has the ability to perform linear system updates for overset rows on the device. Due to lack of ghosting on device, this capability is restricted to only _decoupled overset solves_ for NGP builds. 

- Introduce NGP versions of overset field interpolation/updates. Perform a sync to host and then back again to device. This will in future be removed as we gradually transition parts of overset search onto device.
- Introduce Kokkos Views for storing entities where linear system rows must be reset for decoupled overset algorithm. 
- Use device coeff applier to populate linear system for the overset rows.
- Add necessary throw statements to prevent non-decoupled overset algorithms being run on GPU builds.
- Add new timers to track overset connectivity and field updates separately. This will be used to keep track of the field interpolation overhead due to sync between host and device.

## Checklist

  - Operating systems 
    - [x] Linux
    - [x] MacOS
  - Compilers 
    - [x] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

### DIFF
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -3606,7 +3606,7 @@ Realm::dump_simulation_time()
   }
 
   // nonconformal or overset
-  if ( has_non_matching_boundary_face_alg() ) {
+  if ( hasNonConformal_ ) {
     double g_totalNonconformal = 0.0, g_minNonconformal= 0.0, g_maxNonconformal = 0.0;
     stk::all_reduce_min(NaluEnv::self().parallel_comm(), &timerNonconformal_, &g_minNonconformal, 1);
     stk::all_reduce_max(NaluEnv::self().parallel_comm(), &timerNonconformal_, &g_maxNonconformal, 1);
@@ -3615,6 +3615,20 @@ Realm::dump_simulation_time()
     NaluEnv::self().naluOutputP0() << "Timing for Nonconformal: " << std::endl;
     NaluEnv::self().naluOutputP0() << "  nonconformal bc --  " << " \tavg: " << g_totalNonconformal/double(nprocs)
                                    << " \tmin: " << g_minNonconformal << " \tmax: " << g_maxNonconformal << std::endl;
+  }
+
+  if (hasOverset_) {
+    double connTime[2] = {oversetManager_->timerConnectivity_, oversetManager_->timerFieldUpdate_};
+    double totTime[2], minTime[2], maxTime[2];
+    stk::all_reduce_sum(NaluEnv::self().parallel_comm(), connTime, totTime, 2);
+    stk::all_reduce_min(NaluEnv::self().parallel_comm(), connTime, minTime, 2);
+    stk::all_reduce_max(NaluEnv::self().parallel_comm(), connTime, maxTime, 2);
+    NaluEnv::self().naluOutputP0()
+      << "Timing for Overset:" << std::endl
+      << "     connectivity --  \tavg: " << totTime[0] / double(nprocs)
+      << " \tmin: " << minTime[0] << " \tmax: " << maxTime[0] << std::endl
+      << "     field update --  \tavg: " << totTime[1] / double(nprocs)
+      << " \tmin: " << minTime[1] << " \tmax: " << maxTime[1] << std::endl;
   }
 
   // transfer
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -104,6 +104,7 @@
 #include "ngp_utils/NgpLoopUtils.h"
 #include "ngp_utils/NgpFieldBLAS.h"
 #include "ngp_utils/NgpFieldManager.h"
+#include "ngp_utils/NgpFieldUtils.h"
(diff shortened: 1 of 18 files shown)


src/overset/AssembleOversetDecoupledAlgorithm.C
```diff
@@ -32,40 +32,13 @@ void AssembleOversetDecoupledAlgorithm::execute()
...
-    // Don't use apply_coeff here as it checks for overset logic
-    eqSystem_->linsys_->sumInto(connected_nodes, scratchIds, scratchVals, rhs, lhs, __FILE__);
-  }
-#endif
+  const auto& fringeNodes = realm_.oversetManager_->ngpFringeNodes_;
+  auto* coeffApplier = eqSystem_->linsys_->get_coeff_applier();
+  Kokkos::parallel_for(
+    fringeNodes.size(), KOKKOS_LAMBDA(const size_t& i) {
+      coeffApplier->resetRows(1, &fringeNodes(i), 0, numDof, 1.0, 0.0);
+    });
```
sayerhs, line 41: @alanw0 I wasn't sure what the best approach here would be:

1. Launch a `parallel_for` loop for all the available entities and then call `resetRows` for each entity, or
2. Launch `parallel_for` with 1 and then let `resetRows` loop over the entities 

I opted for the first one, as I felt that it would reset rows in parallel whereas looping from within `resetRows` will have 1 thread updating all the rows.

src/overset/UpdateOversetFringeAlgorithmDriver.C
```diff
@@ -36,13 +37,28 @@ void UpdateOversetFringeAlgorithmDriver::execute()
...
+  }
+
+  const auto& fieldMgr = realm_.mesh_info().ngp_field_manager();
+  for (auto& finfo: fields_) {
+    auto& ngpField = fieldMgr.get_field<double>(finfo.field_->mesh_meta_data_ordinal());
+    ngpField.sync_to_host();
   }
   oversetManager->overset_update_fields(fields_);
+  for (auto& finfo: fields_) {
+    auto& ngpField = fieldMgr.get_field<double>(finfo.field_->mesh_meta_data_ordinal());
```
alanw0, line 60: Why not do these syncs and modify calls inside overset_update_fields?

src/overset/AssembleOversetDecoupledAlgorithm.C
alanw0: Yes this is probably fine. I think the coeffApplier code assumes it is called on device, so it can't launch a parallel-for in there. So doing the parallel-for outside is probably the only option.

src/Realm.C
```diff
@@ -2954,7 +2955,29 @@ Realm::overset_orphan_node_field_update(
...
+{
+  const auto& fieldMgr = ngp_field_manager();
+  auto& ngpField =
+    fieldMgr.get_field<double>(field->mesh_meta_data_ordinal());
+  ngpField.sync_to_host();
+  oversetManager_->overset_update_field(field, nRows, nCols);
+  ngpField.modify_on_host();
+
+  if (doFinalSyncToDevice)
+    ngpField.sync_to_device();
```
alanw0, line 2983: I think a guiding convention i would like to advocate is keeping the modify_on_host/modify_on_device calls as close as possible to the code that is doing the modifying, and keeping the sync_to_host/sync_to_device calls as close as possible to the code that needs the sync'd data.

src/overset/UpdateOversetFringeAlgorithmDriver.C
sayerhs: Because TIOGA classes don't know about Realm or MeshInfo.

alanw0: Note that the modify_on_* and sync_to_* calls are available on stk::mesh::FieldBase now, so you don't need an ngpField to call those. I guess if the TIOGA scope has no knowledge that there could be a GPU device, then it doesn't make sense for that scope to mark fields as modified nor to sync them.

sayerhs: Oh! I didn't realize that.. that would make this code a lot concise.

sayerhs: Thanks @alanw0. With the ability to use `modify/sync` methods on `FieldBase`, I was able to move the synchronization calls into the Tioga-STK interface class. This should address your _guiding convention_ comment in `Realm.C` also.

src/Realm.C
sayerhs: See https://github.com/Exawind/nalu-wind/pull/591#discussion_r426326776

PR #596, opened 2020-05-19
NGP regression test for mesh motion + overset

sayerhs:
This pull-request adds an NGP version of the rotating cylinder regression test that runs both on host and device. 

The mesh-motion and overset connectivity are preformed on the host currently. This pull request just adds the necessary device synchronizations on the GPU. This is intended to provide a way to check the transition of overset and mesh-motion algorithms to the device as those parts of nalu-wind are updated. 

A new regression test is created because of the following reasons: 1. NGP algorithms are only available for edge based scheme, and 2. NGP + overset only available for decoupled solves. 

## Checklist

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [x] MacOS
  - Compilers 
    - [x] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [x] NVIDIA CUDA
- [x] Passes all unit tests


src/mesh_motion/MeshMotionAlg.C
```diff
@@ -84,6 +85,23 @@ void MeshMotionAlg::initialize( const double time )
...
+  {
+    const auto& meta = bulk_.mesh_meta_data();
+    std::vector<std::string> fnames{
+      "current_coordinates",
+      "mesh_displacement",
+      "mesh_velocity",
+    };
+
+    for (const auto& ff: fnames) {
+      auto* fld = meta.get_field(stk::topology::NODE_RANK, ff);
```
alanw0, line 99: I always have an internal alarm bell go off when I see get-field-by-name. Perhaps this isn't done often enough to be a performance issue. But don't we have the actual fields somewhere nearby?

sayerhs: This class doesn't hold the fields, this is more like the AlgDrivers we have been using. It is holds the sub-algorithms that hold references to the field.

PR #598, opened 2020-05-20
ABL precursor NGP fixes and regression test

sayerhs:
This pull-request fixes field synchronization issues with Enthalpy equation system when running on GPUs, and introduces a new regression test for comparing CPU/GPU runs with tighter tolerances. A new regression test is created for the tighter tolerances on the linear solves. 

Currently, I am using the `clipValue` to determine a minimum TKE value, if I don't do that the turbulent viscosity values end up with some error as it is `sqrt(TKE)`. This happens in the first few timesteps as TKE is initialized to zero and it takes a few timesteps for it to develop. I am not sure what to do about that.


src/AuxFunctionAlgorithm.C
```diff
@@ -78,6 +79,9 @@ AuxFunctionAlgorithm::execute()
 
     auxFunction_->evaluate(coords, time, nDim, length, fieldData, fieldSize);
   }
+
+  field_->modify_on_host();
+  field_->sync_to_device();
```
alanw0, line 84: Are you sure we want the sync_to_device here? It seems like it should be located in/near the code that needs the field on device.

sayerhs: The only thing is performance penalty right? I feel that we should get it correct first and then profile and fix performance. 

Also we call a lot of the ngp algorithms through things like `Assemble*SolverAlgorithm`. As written, it is not possible to know which fields are going to be accessed. If we change the interfaces such that we can call `sync_to_device` on ElemDataRequests (but this doesn't exist for edge and node algorithms) then we can do what you are suggesting.

sayerhs:
[log.cpu.txt](https://github.com/Exawind/nalu-wind/files/4659190/log.cpu.txt)
[log.gpu.txt](https://github.com/Exawind/nalu-wind/files/4659191/log.gpu.txt)

```

   *****************************************************************
              EXODIFF	(Version: 3.01) Modified: 2020-01-06
              Authors:  Richard Drake, rrdrake@sandia.gov
                        Greg Sjaardema, gdsjaar@sandia.gov
              Run on    2020/05/20   13:04:35 MDT
   *****************************************************************

Reading first file ...
Reading second file ...
  FILE 1: /lustre/eaglefs/scratch/sanantha/exwsims/abl_neutral/out_cpu/abl.e
   Title: IOSS Default Output Title
          Dim = 3, Blocks = 1, Nodes = 428652, Elements = 390625, Nodesets = 1, Sidesets = 6
          Vars: Global = 0, Nodal = 8, Element = 0, Nodeset = 0, Sideset = 0, Times = 11

  FILE 2: /lustre/eaglefs/scratch/sanantha/exwsims/abl_neutral/out_gpu/abl.e
   Title: IOSS Default Output Title
          Dim = 3, Blocks = 1, Nodes = 428652, Elements = 390625, Nodesets = 1, Sidesets = 6
          Vars: Global = 0, Nodal = 8, Element = 0, Nodeset = 0, Sideset = 0, Times = 11


Nodal coordinates will be compared .. tol:    1e-06 (absolute), floor:        0
Time step values will be compared  .. tol:    1e-10 (absolute), floor:        0
No Global variables on either file.
Nodal variables to be compared:
enthalpy                         tol:    1e-10 (absolute), floor:        0
pressure                                 1e-10 (absolute),               0
temperature                              1e-10 (absolute),               0
turbulent_ke                             1e-10 (absolute),               0
turbulent_viscosity                      1e-10 (absolute),               0
velocity_x                               1e-10 (absolute),               0
velocity_y                               1e-10 (absolute),               0
velocity_z                               1e-10 (absolute),               0
No Element variables on either file.
No Element Attribute variables on either file.
No Nodeset variables on either file.
No Sideset variables on either file.
Sideset Distribution Factors will be compared .. tol:    1e-06 (relative), floor:        0

  ==============================================================
   NOTE: All node and element ids are reported as global ids.

Sideset Distribution Factors:
Comparing only the final step (step 11 on first, step 11 on second) on each database.

  --------- Time step 11, 5.0000000e+00 ~ 5.0000000e+00, abs diff:  0.00000e+00 ---------
Nodal variables:

exodiff: Files are the same
```

Issue #588, opened 2020-05-15
CUDA warnings for ActuatorNGP 


sayerhs:
NGP code introduced for `Actuator` generates warnings when compiled for CUDA

```
/.../nalu-wind/include/actuator/ActuatorNGP.h(63): warning: calling a __host__ function from a __host__ __device__ function is not allowed
```

More details [CDash](https://my.cdash.org/viewBuildError.php?type=1&buildid=1825358)

psakievich:
I'll take a look

PR #607, opened 2020-06-03
Fix warnings from compiling with Nvidia

psakievich:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

Remove warnings when compiling with Nvidia in response to #588.  Also a little clean up and making it so Cuda space is now used when compiling for device.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [x] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*


include/actuator/ActuatorTypes.h
```diff
@@ -16,9 +16,9 @@
 namespace sierra {
 namespace nalu {
 
-#ifdef ACTUATOR_ON_DEVICE
+#ifdef __CUDACC__
```
sayerhs, line 19: We have been using `KOKKOS_ENABLE_CUDA` in other places, can you switch to that? Will be easier to replace it with whatever Kokkos suggests for `KOKKOS_ENABLE_<ACCELERATOR>`

psakievich: Sure thing.

psakievich:
@sayerhs also it looks like the `ActuatorSourceKernal` currently dies in the unittests so I don't think we're ready for a regression test yet.

include/actuator/ActuatorTypes.h
```diff
@@ -16,15 +16,14 @@
 namespace sierra {
 namespace nalu {
 
-#ifdef ACTUATOR_ON_DEVICE
+#ifdef KOKKOS_ENABLED_CUDA
```
sayerhs, line 19: I think there is a typo here. It is `KOKKOS_ENABLE_CUDA` not `ENABLED`

psakievich: ah thanks

psakievich:
@sayerhs right now it won't work. That's beyond the scope of what I was shooting for in this PR. With the last push I made all the non-fast actuator unit-tests are passing with a cuda build.  There should just be some small clean up to the kernels that are enabled when using fast to get it working though. I'd say let's merge this PR and open a new issue to address that goal.

PR #623, opened 2020-06-15
Fix more build warnings

psakievich:
Fix actuator related build warnings

### DIFF
--- a/include/actuator/ActuatorFunctorsSimple.h
+++ b/include/actuator/ActuatorFunctorsSimple.h
@@ -101,11 +101,11 @@ struct ActSimpleComputeForce
     
   // Dual view polar tables and blade definitions
   ActScalarIntDv   polartable_size_;
-  const int        Npolartable;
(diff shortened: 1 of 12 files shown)


include/actuator/ActuatorFunctorsSimple.h
```diff
@@ -101,11 +101,11 @@ struct ActSimpleComputeForce
     
   // Dual view polar tables and blade definitions
   ActScalarIntDv   polartable_size_;
-  const int        Npolartable;
+  const unsigned   Npolartable;
```
sayerhs, line 104: @lawrenceccheung @psakievich While we are at it, can we also make this consistent with Nalu naming convention?

```
Npolartable -> nPolarTable_
Npts -> nPts_
twist_tableDv_ -> twistTableDv_
elem_areaDv_ -> elemAreaDv_
```

psakievich: yeah I can update those

PR #627, opened 2020-06-21
Github CI workflow using docker images

sayerhs:
This pull-request adds a build/test CI workflow for Nalu-Wind repository using docker containers to manage all the necessary dependencies. You can see it [in action](https://github.com/Exawind/nalu-wind/runs/792833352?check_suite_focus=true).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,8 @@
 .git
 examples
 reg_tests
-**/build*
-build
+build*
 wind-utils
+python
+**/tests
+.*
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.git
+examples
+reg_tests
+**/build*
+build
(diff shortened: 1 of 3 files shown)


jrood-nrel:
Cool. Some of the regression tests are fairly short. Would it make sense to run a small handful?

sayerhs:
@jrood-nrel I think we should enable a few regression tests too. I didn't want to do that in this pull-request as I wanted the bare-minimum changes to the core nalu-wind codebase to make sure this docker pathway works. 

To get regression tests working we need to figure out the best pathway to getting all the files into the docker image. Right now as you can see I am doing a `COPY` which is not ideal if you've to copy the meshes. There are two alternatives volumes and bind mounts. Bind mounts are easier, but for some reason that is discouraged. So I want to investigate more.

PR #636, opened 2020-06-24
Add CMake installation logic to Nalu-Wind

sayerhs:
This pull-request makes Nalu-Wind installable and findable by CMake, so that we can use Nalu-Wind as a dependency for the hybrid Nalu/AMR coupling strategy.

- Don't leak compiler warning/error flags to downstream packages
- Add CMake configuration to allow `find_package(Nalu-Wind)`

*All pull requests*
  - Operating systems 
    - [X] Linux
    - [X] MacOS
  - Compilers 
    - [X] GCC
    - [X] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA


sayerhs:
@lucbv @psakievich @lawrenceccheung Is this going to break Nalu-Wind builds at SNL

lucbv:
I'll give this a try on my workstation and let you know what I observe

lucbv:
@sayerhs I had no problem building as usual, do you want me to try to do a make install or something like that?

sayerhs:
Thanks @lucbv. No install isn't necessary. I wanted to make sure that SNL's CMake version can handle the commands I added in the CMakeLists.txt. If it went through configure, then we are good.

lucbv:
As mentioned in another issue or PR, we are up to CMake 3.17.1, we should be able to handle modern CMake a lot better now. : )

PR #637, opened 2020-06-24
Refactor overset implementation in preparation for multi-Realm/multi-solver overset coupling

sayerhs:
This pull-request is the first stage of refactor of the overset connectivity algorithm implementation within Nalu-Wind to extract the overset connectivity logic from inside Realm to a higher-level so that we can do multi-Realm overset and AMR-Wind/Nalu-Wind overset simulations. 

## Checklist

  - Operating systems 
    - [x] Linux
    - [x] MacOS
  - Compilers 
    - [x] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

### DIFF
--- a/src/AssembleNodalGradAlgorithmDriver.C
+++ b/src/AssembleNodalGradAlgorithmDriver.C
@@ -120,7 +120,7 @@ AssembleNodalGradAlgorithmDriver::post_work()
   if ( realm_.hasOverset_ ) {
     // this is a vector
     const unsigned nDim = meta_data.spatial_dimension();
-    realm_.overset_orphan_node_field_update(dqdx, 1, nDim);
+    realm_.overset_field_update(dqdx, 1, nDim);
   }
 }
 
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -546,10 +546,10 @@ Realm::initialize()
   if ( hasNonConformal_ )
     initialize_non_conformal();
 
-  if ( hasOverset_ ) {
-    oversetManager_->initialize();
-    initialize_overset();
-  }
+  // if ( hasOverset_ ) {
+  //   oversetManager_->initialize();
+  //   initialize_overset();
+  // }
 
   initialize_post_processing_algorithms();
 
@@ -2350,7 +2350,7 @@ Realm::initialize_non_conformal()
 void
 Realm::initialize_overset()
 {
-  if (!isExternalOverset_)
+  if (hasOverset_ && !isExternalOverset_)
     oversetManager_->execute(equationSystems_.all_systems_decoupled());
 }
 
@@ -3003,7 +3003,7 @@ Realm::overset_field_update(
   const unsigned nCols,
   const bool doFinalSyncToDevice)
 {
-  if (!hasOverset_) return;
+  if (!hasOverset_ || isExternalOverset_) return;
 
   const double timeA = NaluEnv::self().nalu_time();
   oversetManager_->overset_update_field(
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -546,8 +546,10 @@ Realm::initialize()
   if ( hasNonConformal_ )
     initialize_non_conformal();
 
-  if ( hasOverset_ )
+  if ( hasOverset_ ) {
+    oversetManager_->initialize();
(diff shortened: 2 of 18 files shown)


sayerhs:
The following tests diff due to this refactor.

```
FAIL: oversetRotCylinder......................    60.0899s 3.8629e-13 7.0971e-09
FAIL: oversetRotCylNGPTrilinos................    80.9438s 3.1971e-09 2.2611e-06
FAIL: oversetRotCylinderHypre.................    81.6712s 3.6088e-11 1.3252e-07
```

src/overset/TiogaSTKIface.C
```diff
@@ -350,7 +350,7 @@ TiogaSTKIface::populate_overset_info()
   ThrowAssert(osetInfo.size() == 0);
 
   VectorFieldType *coords = meta_.get_field<VectorFieldType>
-    (stk::topology::NODE_RANK, realm.get_coordinates_name());
+    (stk::topology::NODE_RANK, coordsName_);
```
alanw0, line 370: There is a function meta_.coordinate_field(); stk-io calls meta_.set_coordinate_field();

src/Realm.C
```diff
@@ -2520,8 +2520,11 @@ Realm::compute_vrtm(const std::string& velName)
   auto vrtm = fieldMgr.get_field<double>(
     get_field_ordinal(*metaData_, velName + "_rtm"));
 
+  auto* vrtm_field = metaData_->get_field<VectorFieldType>(
+    stk::topology::NODE_RANK, velName + "_rtm");
```
alanw0, line 2524: Does this potentially need a 'sync_to_host' or anything like that?

src/Realm.C
```diff
@@ -1774,8 +1774,8 @@ Realm::pre_timestep_work()
       initialize_non_conformal();
 
     // and overset algorithm
-    if ( hasOverset_ )
-      initialize_overset();
+    // if ( hasOverset_ )
```
alanw0: Did you mean to comment this out?

src/AssembleNodalGradAlgorithmDriver.C
```diff
@@ -120,7 +120,7 @@ AssembleNodalGradAlgorithmDriver::post_work()
   if ( realm_.hasOverset_ ) {
     // this is a vector
     const unsigned nDim = meta_data.spatial_dimension();
-    realm_.overset_orphan_node_field_update(dqdx, 1, nDim);
+    realm_.overset_field_update(dqdx, 1, nDim);
```
ashesh2512, line 123: Really appreciate this change in terminology. The previous terminology was quite confusing.

src/overset/ExtOverset.C
```diff
@@ -0,0 +1,135 @@
...
+    }
+  }
+
+  if (multiSolverMode_ && (noverset == 0)) {
+    throw std::runtime_error(
+      "Multi-solver mode is active but no realm has overset");
+  }
+
+  if (!isDecoupled_ && (noverset > 1)) {
+    throw std::runtime_error("External overset requires decoupled solves");
```
ashesh2512, line 57: Above I was wondering why are we getting the decoupled flag, but this check resolves that question. This is good!

src/Realm.C
sayerhs: @alanw0 I don't think so. `vrtm_field` is only used for `stk::mesh::selectField` here and not really modified.

src/overset/TiogaSTKIface.C
sayerhs: @alanw0 In Nalu, we use `coordinates` (the field name read from the Exodus file) when there is no mesh motion, but use `current_coordinates` when there is mesh motion. Previously, this field name was retrieved from `realm` within the TIOGA-STK interface. Now that I've removed Realm object in this interface, it is now being passed through the constructor. using `meta_.coordinate_field()` won't be appropriate as it would still return `coordinates` in case of mesh motion.

src/Realm.C
sayerhs: I've no idea why this showed up.

PR #639, opened 2020-06-25
Refactor of overset logic to allow multi-solver mode (2)

sayerhs:
Builds on #637 and does the following:

- Clean up Realm initialization to perform pre/post actions properly and let TimeIntegrator handle overset
- Remove realm specific code calls that were in TimeIntegrator and create realm pre/post work methods to handle it cleanly within Realm
- Add additional logic in overset manager to restore original code behavior for fully-coupled overset solves, #637 introduced small diffs that this pull-request restores original behavior. 

## Checklist
*All pull requests*
  - Operating systems 
    - [X] Linux
    - [ ] MacOS
  - Compilers 
    - [X] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [X] Passes all unit tests


include/overset/ExtOverset.h
```diff
@@ -44,15 +44,26 @@ class ExtOverset
   void set_multi_solver_mode(const bool flag)
   { multiSolverMode_ = flag; }
 
+  bool is_external_overset() const
+  { return hasOverset_ && isExtOverset_; }
```
ashesh2512, line 48: Should `hasOverset_` be `true` if `isExtOverset_` is `true` ?

sayerhs: Thanks, that is indeed redundant. I had a more complicated logic that I replaced with `isExtOverset_` but overlooked an additional simplification.

sayerhs: Fixed in 81c7d132cbc4

sayerhs:
@ashesh2512 Since @alanw0 hasn't reviewed it yet... I added one more commit. Can you please take a look at that: 
5eeb1c6

PR #643, opened 2020-06-30
add an option to use Trilinos' s-step GMRES

iyamazaki:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

This PR adds an option to use s-step GMRES from Trilinos. 
- In the input file, we can specify to use s-step GMRES by "method: sstep_gmres" in the input file (with an optional step size "krylov_step_size: 5")   
- Currently, there is no unit-test for this option, but I have tested the option with a couple of inputs on Summit.

## Checklist

*All pull requests*
- [x] Builds successfully on Summit
- [ ] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/LinearSolver.C
+++ b/src/LinearSolver.C
@@ -26,7 +26,6 @@
 #include <BelosConfigDefs.hpp>
 #include <BelosLinearProblem.hpp>
 #include <BelosTpetraAdapter.hpp>
-#include <Belos_Tpetra_GmresSstep.hpp>
 
 #include <Ifpack2_Factory.hpp>
 #include <Kokkos_DefaultNode.hpp>
@@ -113,10 +112,6 @@ void TpetraLinearSolver::setupLinearSolver(
 
     // create the solver, e.g., gmres, cg, tfqmr, bicgstab
     LinSys::SolverFactory sFactory;
-    if (config_->get_method() == "TPETRA GMRES S-STEP") {
-      bool verbose = (params_->get<int>("Output Frequency", 0) > 0);
-      BelosTpetra::Impl::register_GmresSstep (verbose);
-    }
     solver_ = sFactory.create(config_->get_method(), params_);
     solver_->setProblem(problem_);
   }
--- a/src/LinearSolver.C
+++ b/src/LinearSolver.C
@@ -26,6 +26,7 @@
 #include <BelosConfigDefs.hpp>
 #include <BelosLinearProblem.hpp>
 #include <BelosTpetraAdapter.hpp>
+#include <Belos_Tpetra_GmresSstep.hpp>
 
 #include <Ifpack2_Factory.hpp>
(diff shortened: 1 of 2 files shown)


src/LinearSolver.C
```diff
@@ -112,6 +113,10 @@ void TpetraLinearSolver::setupLinearSolver(
 
     // create the solver, e.g., gmres, cg, tfqmr, bicgstab
     LinSys::SolverFactory sFactory;
+    if (config_->get_method() == "TPETRA GMRES S-STEP") {
+      bool verbose = (params_->get<int>("Output Frequency", 0) > 0);
+      BelosTpetra::Impl::register_GmresSstep (verbose);
```
alanw0, line 118: We have to call a function with the option encoded into its name? Is that a static method? Is it putting something into a singleton? Why isn't the option just passed through params_ or however belos gets most of its option choices?

iyamazaki: This is not one of the Belos solvers. It uses Tpetra's MultiVector (instead of MvTraits). Currently, to use it through BelosSolverFactory, we need to register each solver. But, it will be nice to improve the interface. If you have any suggestions, please let us know.. Thank you.

sayerhs: @iyamazaki Are you suggesting that there is a way to improve the interface on the Nalu-Wind side? If so, I'd recommend we take this opportunity to modify the solver registration interface so that we don't have to keep changing the solver interface and recompiling the code, as @jhux2 suggests.

iyamazaki: No.. I believe the changes need to happen in Belos side. We are discussing how to proceed..

iyamazaki: With [PR#7623](https://github.com/trilinos/Trilinos/pull/7623), we don't need to register GmresSstep. The code would compile with an earlier version of Trilinos, but will throw std::invalid_argument if we try to use sstep_gmres. Thank you so much for the review, and let us know if you have more comments!!

sayerhs:
Thanks @iyamazaki . I'll wait for the CI checks to complete before merging.

iyamazaki:
I could not figure out how to get the details of the failed check. Did it pass the check?

sayerhs:
@iyamazaki It is a spurious error. I think the first build attempt just failed because of machine issues... and the second attempt succeeded. 

If you look on the top of this pull request there are tabs `Conversations, Commits, Checks, Files changed` clicking on the "Checks" tab will show you the CI output.

iyamazaki:
Thank you for the reviews, everyone!! I clicked on the failed build, and I only saw "This check failed" and was not sure.

sayerhs:
@iyamazaki Here is the link to the successful build report: https://github.com/Exawind/nalu-wind/pull/643/checks?check_run_id=842388606

PR #658, opened 2020-07-16
Implementation of Hypre GPU/Kokkos matrix assembly. Kokkos is used fo…

PaulMullowney:
…r the assembly process.

Then, the data is moved back to host data structures and matrices are injected into the
Hypre API calls HYPRE_IJMatrixSet/AddToValues (same for the vector). This branch does
NOT use/integrate HYPRE GPU solving capabilities.

The Kokkos implementation support both Host and Device execution. Host execution is 2-4x
faster than the original implementation. Device execution is about 10x faster than host
execution. Device assembly is working on up to 32 GPUs on eagle. THe original host path
for assembly has been removed.

Can do momentum as either scalar or vector (segregated) solver.

Also, added an option for Hypre assembly, ensure_reproducible, that when enabled, sorts
the matrix/rhs values that are summed for each entry, by absolute value floating point
ascending order. This ensures an ordering and a matrix/rhs pair that is guaranteed to be
the same given the same input (lhs/rhs). In spite of this, runs still aren't deterministic.
Sigh. This is achieved via Thrust::stable_sort, stable_sort_by_key on Cuda devices only.
One must also use use_native_cuda_sort:yes in order to ensure reproducibility. Later, we
plan on adding a Kokkos version of this sort. For unsupported architectures, a runtime
execption is thrown indicate as much.

Finally, we have added support for overset on the CPU via Kokkos interface.

**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*


src/HypreLinearSystem.C
```diff
@@ -140,83 +126,543 @@ HypreLinearSystem::beginLinearSystemConstruction()
...
   beginLinearSystemConstruction();
+  stk::mesh::MetaData & metaData = realm_.meta_data();
+  const stk::mesh::Selector s_owned = metaData.locally_owned_part()
+                                      & stk::mesh::selectUnion(parts)
+                                      & !(realm_.get_inactive_selector());
+  stk::mesh::BucketVector const& buckets = realm_.get_buckets(realm_.meta_data().side_rank(), s_owned);
+
+  if (numDof_==1) {
+    std::vector<HypreIntType> hids(0);
+    if (buckets.size()) {
```
alanw0, line 269: The sizing of this 'hids' array needs to be inside the outer loop over buckets (at line 275). The buckets contain sides, which can be either quads or tris if the mesh has a mixture of element types. Within each bucket the sides are all the same, but different buckets can have different topologies. So after line 275 you can say 'numNodes = b.topology().num_nodes();' and do the hids.resize there, and then hids will always be the correct size in the inner loop at line 284.

src/HypreLinearSystem.C
```diff
@@ -140,83 +126,543 @@ HypreLinearSystem::beginLinearSystemConstruction()
...
+  const stk::mesh::Selector s_owned = metaData.locally_owned_part()
+                                      & stk::mesh::selectUnion(parts)
+                                      & !(realm_.get_inactive_selector());
+  stk::mesh::BucketVector const& buckets = realm_.get_buckets(stk::topology::EDGE_RANK, s_owned);
+
+  if (numDof_==1) {
+    std::vector<HypreIntType> hids(0);
+    if (buckets.size()) {
+      const unsigned numNodes = (unsigned) (*buckets[0]).num_nodes(0);
+      hids.resize(numNodes);
```
alanw0, line 347: This one should technically be after line 351 also. Edges are less likely to have different node counts than sides, but if we do high-order elements you could get some 3-node edges...

src/HypreLinearSystem.C
```diff
@@ -140,83 +126,543 @@ HypreLinearSystem::beginLinearSystemConstruction()
...
+    }
+
+    for(size_t ib=0; ib<buckets.size(); ++ib) {
+      const stk::mesh::Bucket & b = *buckets[ib];
+      for ( stk::mesh::Bucket::size_type k = 0 ; k < b.size() ; ++k ) {
+
+	const unsigned numNodes = (unsigned) b.num_nodes(k);
```
alanw0, line 354: This numNodes can be obtained outside the loop (between lines 351 and 352). 'numNodes = b.topology().num_nodes();'. All entities within a bucket have the same topology, thus the same numNodes.

src/HypreLinearSystem.C
```diff
@@ -140,83 +126,543 @@ HypreLinearSystem::beginLinearSystemConstruction()
...
+
```
alanw0, line 355: Also the return type of Bucket::num_nodes is unsigned, so the cast to unsigned is not needed.

src/HypreLinearSystem.C
```diff
@@ -295,41 +826,416 @@ HypreLinearSystem::finalizeSolver()
...
+  entityToLIDHost_ = HypreIntTypeViewHost("entityToRowLIDHost",bulk.get_size_of_entity_index_space());
+  entityToLID_ = HypreIntTypeView("entityToRowLID",bulk.get_size_of_entity_index_space());
+
+  const stk::mesh::BucketVector& nodeBuckets = realm_.get_buckets(stk::topology::NODE_RANK, selector);
+  for(const stk::mesh::Bucket* bptr : nodeBuckets) {
+    const stk::mesh::Bucket& b = *bptr;
+    for(size_t i=0; i<b.size(); ++i) {
+      stk::mesh::Entity node = b[i];
+      const auto naluId = *stk::mesh::field_data(*realm_.naluGlobalId_, node);
+      const auto mnode = bulk.get_entity(stk::topology::NODE_RANK, naluId);
```
alanw0, line 843: BulkData::get_entity is expensive (map find). I think you're getting the master node for periodic, meaning naluId==stkId. You could say something like 'mnode = (naluId == bulk.identifier(node)) ? node : bulk.get_entity...'. That way you only use get_entity when you really need to. BulkData::identifier is a lot faster than get_entity.

PaulMullowney:
> It's hard to do a good review of a large change-set like this. I don't see any obvious problems, but left a couple comments regarding stk usage.

I understand that concern. However with this change set, I didn't see a clean way to do it stages. Hypre's assembly was completely at odds with a Kokkos model. Implementing a new version required essentially starting over.

src/HypreLinearSystem.C
PaulMullowney: I sized the arrays outside of the big loops due to speed concerns. I hadn't encountered situations where the node count is different for a particular build*Graph algorithm?

Should I go in and fix these issues in my branch?

alanw0: I usually rank correctness above speed. If you put the resize inside the outer loop it's not much of a speed hit. If all your cases have homogeneous sides then it's no speed hit at all because a vector resize is very cheap if the size is staying the same.

PaulMullowney: Agreed. I think I assumed that for methods like buildEdgeToNodeGraph, each element in the bucket would have the same number of nodes.

src/HypreLinearSystem.C
PaulMullowney: Does this apply buildFaceToNodeGraph and buildElemToNodeGraph also?

alanw0: Yes the general rule is that within a bucket entities (faces, elements) all have the same number of nodes due to all having the same topology. But when an outer loop goes over multiple buckets, things can change from one bucket to the next. (Unless the selector that selected the buckets ensures that only homogeneous things are selected.)

PaulMullowney: Ok. I've fixed buildEdgeToNodeGraph, buildFaceToNodeGraph and buildElemToNodeGraph to be:
  if (numDof_==1) {
    std::vector<HypreIntType> hids(0);

    for(size_t ib=0; ib<buckets.size(); ++ib) {
      const stk::mesh::Bucket & b = *buckets[ib];

      auto numNodes = b.topology().num_nodes();
      hids.resize(numNodes);

      for ( stk::mesh::Bucket::size_type k = 0 ; k < b.size() ; ++k ) {

        stk::mesh::Entity const * nodes = b.begin_nodes(k);

        /* save the hypre ids */
        for (unsigned i=0; i<numNodes; ++i) hids[i] = get_entity_hypre_id(nodes[i]);

        /* fill owned/shared 1 Dof */
        fill_owned_shared_data_structures_1DoF(numNodes, hids);
      }
    }
  } else {
    std::vector<HypreIntType> hids(0);
    std::vector<HypreIntType> columns(0);

    for(size_t ib=0; ib<buckets.size(); ++ib) {
      const stk::mesh::Bucket & b = *buckets[ib];

      auto numNodes = b.topology().num_nodes();
      hids.resize(numNodes);
      columns.resize(numNodes*numDof_);

      for ( stk::mesh::Bucket::size_type k = 0 ; k < b.size() ; ++k ) {

        stk::mesh::Entity const * nodes = b.begin_nodes(k);

        /* save the hids and columns */
        fill_hids_columns(numNodes, nodes, hids, columns);

        /* fill owned/shared for more than 1 Dof */
        fill_owned_shared_data_structures(numNodes, hids, columns);
      }
    }
  }

include/LinearSystem.h
```diff
@@ -66,6 +66,16 @@ class CoeffApplier
 
   virtual void free_device_pointer() = 0;
   virtual CoeffApplier* device_pointer() = 0;
+  virtual void resetInternalData() {};
+  virtual void applyDirichletBCs(Realm &,
+				 stk::mesh::FieldBase *,
+				 stk::mesh::FieldBase *,
+				 const stk::mesh::PartVector&) {};
+  virtual void finishAssembly(void *, std::vector<void *>) {};
```
alanw0, line 74: I prefer not to see function signatures with un-named arguments. What are these arguments? And if I did know what they were, I would probably object to using void.

include/LinearSystem.h
```diff
@@ -66,6 +66,16 @@ class CoeffApplier
 
   virtual void free_device_pointer() = 0;
   virtual CoeffApplier* device_pointer() = 0;
+  virtual void resetInternalData() {};
+  virtual void applyDirichletBCs(Realm &,
+				 stk::mesh::FieldBase *,
+				 stk::mesh::FieldBase *,
+				 const stk::mesh::PartVector&) {};
```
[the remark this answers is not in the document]
PaulMullowney: I don't think this needs to be implemented in Tpetra. In HypreLinearSystem.C, I use methods like this as well as sum_into_nonNGP to write in the Kokkos data structures held in HyperLinSysCoeffApplier.

include/LinearSystem.h
PaulMullowney: unnamed gets rid of compiler warnings for derived classes that don't implement these methods like Tpetra. void * is nasty, I would agree. The challenge here is that I need to get my Kokkos data structures for the CRS matrix into a Hypre data structure. Now, I can't pass a Hypre specific type into this method as would break a non Hypre build. The alternative to passing Hypre structures in is to pull Kokkos CRS data structures out.

alanw0: ok, I see the challenge, I agree that's hard. Maybe you need to stick with void. As for the names, maybe you could put commented-out names, like finishAssembly(void * /\*my-arg\*/)

src/HypreLinearSystem.C
PaulMullowney: This code is legacy from the original implementation though I'm still relying on it. Changing this line (and one other instance of the same call) to:
const auto mnode = (naluId == bulk.identifier(node)) ? node : bulk.get_entity(stk::topology::NODE_RANK, naluId);
seems to work though I can't tell if things are faster or not.

include/LinearSystem.h
PaulMullowney: yes. That's good. I'll try to stick to that convention.

src/HypreLinearSystem.C
PaulMullowney: fixed

include/LinearSystem.h
sayerhs: @PaulMullowney I don't see any reason why `FinishAssembly` needs to be part of the public API and therefore `LinearSystem`. In that case `finish_assembly` can just be a local method with its proper type for each LinearSystem that uses it.

PaulMullowney: I may have tried to be local to HypreLinSysCoeffApplier and failed to get it to compile. I can try again.

PaulMullowney: Actually this pertains to all the new methods of LinearSystem: resetInternalData, finishAssembly, sum_into_nonNGP, and applyDirichletBCs.

PaulMullowney: I've tried:

HypreLinSysCoeffApplier* hcApplier = reinterpret_cast<HypreLinSysCoeffApplier*>(hostCoeffApplier);
hcApplier->finishAssembly(...);

This gives:../src/HypreLinearSystem.C:1194:99: error: invalid cast from type ‘std::unique_ptr<sierra::nalu::CoeffApplier>’ to type ‘sierra::nalu::HypreLinearSystem::HypreLinSysCoeffApplier*’
   HypreLinSysCoeffApplier* hcApplier = reinterpret_cast<HypreLinSysCoeffApplier*>(hostCoeffApplier);

What's the right way to do this?

alanw0: You need to extract the raw pointer from the unique_ptr, then use dynamic_cast. Very rarely use reinterpret_cast for anything. Maybe something like dynamic_cast<HypreLinSysCoeffApplier*>(hostCoeffApplier.get())

sayerhs: You can also do 

```C++
auto coeff_applier = dynamic_cast<HypreLinSysCoeffApplier&>(*hostCoeffApplier);
```

PaulMullowney: Ok. I think I've fixed it. I removed the methods from LinearSystem and used dynamic cast to call them in HypreLinearSystem. I'm gonna push shortly.

include/LinearSystem.h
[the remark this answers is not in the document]
PaulMullowney: method is gone from the base class.

include/LinearSystem.h
PaulMullowney: methods are gone from the base class

src/HypreLinearSystem.C
PaulMullowney: fixed

src/HypreLinearSystem.C
PaulMullowney: Hopefully this is what you had in mind.

src/HypreLinearSystem.C
PaulMullowney: fixed

src/HypreLinearSystem.C
alanw0: Looks good.

PR #668, opened 2020-07-31
Fix regressions introduced by HYPRE assembly updates in #658

PaulMullowney:
This branch contains fixes for the crashing/unstable reg tests. 1) For oversetRotCylMultiRealm, a check was added to ensure that the owned part of the Matrix/Rhs has entries before calling fillCSRMatrix and SetValues. 2) For ablNeutralNGPHypre(Segregated), we run the test on only 2 processors. For 3 or more MPI ranks, the order of the AddToValues calls can change from run to run thus leading to instability. Finally, I have fixed build*NodeGraph code in HypreUVWLinearSystem to use the same APIs as HypreLinearSystem. Gold results for the tests were produced on rhodes.

Commenting out these tests for the time being.

**Pull-request type:**
- [x] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*


reg_tests/CTestList.cmake
```diff
@@ -238,7 +238,7 @@ else(NOT ENABLE_CUDA)
   #=============================================================================
   add_test_r(ablNeutralNGPTrilinos 2)
   add_test_r(conduction_p4 2)
-  add_test_r(airfoilRANSEdgeTrilinos 1)
+  #add_test_r(airfoilRANSEdgeTrilinos 1)
```
sayerhs, line 241: This removes capability introduced in dbe7646292e0d4bb54950bf12799ed4f3ed801d8 Please undo this change

PaulMullowney: fixed

PaulMullowney:
The capabilities are regression tested under the various Hypre reg tests in including ablNeutralNGPHypre, ablNeutralNGPHypreSegregated, and oversetRotCylMultiRealm. The core issue is that Hypre APIs appear to lead to non reproducible results for sims with 3 or more procs. The solution for the testing is to only run them on 2 procs. This is not satisfying however the Hypre team has confirmed this is possible.

Issue #673, opened 2020-08-10
Nacelle Model Issues


psakievich:
## Detailed description of the bug 

The nacelle affects don't appear to be showing up in the ALM runs.  The following image shows the average velocity 0.5D downstream from the turbine.  A large deficit can be seen from the tower, but almost no deficit at all is occurring from the hub.  This issue is being submitted on behalf of Ken Brown and Alan Hsieh.

<img width="791" alt="image" src="https://user-images.githubusercontent.com/12102506/89809063-e9a27b00-daef-11ea-8daa-a157ce8f14df.png">

Drag Coefficient = 1.0
Nacelle Area = 4.0 m^2
Free stream velocity ~= 8m/s
Mesh size = 0.625m

psakievich:
I've done an initial dive into this and it looks like the code paths are being executed correctly.  The funny thing is the forces are dramatically smaller for the hub.  I'm seeing:
```
blade: O(10^4)
tower: O(10^3)
hub: O(10^2)
```
in the `nrel5MWactuatorLine` regression test.  
I'm going to dive further into the implementation tomorrow to see if the units are off in the implemented equation or something.  
@tonyinme or @gantech could you also point me to where the tower computation is done in openfast? I'd like to compare to the hub calculation we're doing.  My fortran keeps getting rustier and rustier as time goes on...

psakievich:
There is theory derived by @tonyinme around the nacelle model in we use that the  integral impact should be consistent in terms of momentum deficit.  He can elaborate better than me.  In practice we see that stagnation impacts from the nacelle are not fully captured so higher speed jet flows through the gaps in the blades rather than a wake profile that intuition suggests.

I am going to close this since it does not appear to be a code deficiency but a model deficiency.  It is worth noting for OpenTurbine, AMR-Wind, OpenFAST and other projects focused on actuator based turbine modeling. @michaelasprague @pscrozi

It would be worth exploring this further when resolved nacelles are the issue at hand for the development and research community in the ExaWind program.

tonyinme:
Hi @psakievich , yes, I agree with you. The nacelle actuator model is meant to create a wake that will match the momentum thickness of the wake based on the drag coefficient and area. A resolved nacelle model will be able to provide more details and correct flow physics. There is some work in the literature exploring the differences.

PR #675, opened 2020-08-12
Allow users to fix hub epsilon

psakievich:
The current nacelle model determines epsilon automatically
based on the paramaters that are supplied.
One side affect of this model is as cD and area are increased so is
epsilon, making the nacellel forcing more diffuse.
This change allows users to fix the hub epsilon to concentrate the
actuator force from the hub instead of having it be calculated
automatically for them.

An intermediate step to give analysts a little more freedom while we work on #673 

**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/actuator/ActuatorParsingFAST.C
+++ b/src/actuator/ActuatorParsingFAST.C
@@ -11,8 +11,6 @@
 #include <NaluParsing.h>
 #include <actuator/ActuatorParsingFAST.h>
 #include <NaluEnv.h>
-#include <yaml-cpp/node/node.h>
-#include <yaml-cpp/node/type.h>
 
 namespace sierra {
 namespace nalu {
--- a/src/actuator/ActuatorParsingFAST.C
+++ b/src/actuator/ActuatorParsingFAST.C
@@ -11,6 +11,8 @@
 #include <NaluParsing.h>
 #include <actuator/ActuatorParsingFAST.h>
 #include <NaluEnv.h>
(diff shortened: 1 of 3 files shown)


src/actuator/ActuatorParsingFAST.C
```diff
@@ -11,6 +11,8 @@
 #include <NaluParsing.h>
 #include <actuator/ActuatorParsingFAST.h>
 #include <NaluEnv.h>
+#include <yaml-cpp/node/node.h>
+#include <yaml-cpp/node/type.h>
```
sayerhs, line 15: Why not 

```C++
#include <yaml-cpp/yaml.h>
```

psakievich: This was an IDE plugin's fault. Didn't notice it until now. These aren't even necessary.

psakievich: I'll clean up the unnecessary includes

psakievich:
Minor update. Merging despite CI failure because CI failure is just a misalignment with the code base and the container version of Trilinos

PR #674, opened 2020-08-12
Fix compile error

alanw0:
stk's EntitySorterBase base class has moved into its own header.
It used to be declared in EntityLess.hpp

**Pull-request type:**
- [x] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

## Description of the pull-request

*Provide a description of your proposed changes. If there is a related issue, please specify.**

## Checklist

**NOTE** The checklist below is a suggestion and not mandatory. You don't have
to _check all boxes_ to submit a pull request. However, please add a brief
explanation in your pull request summary explaining the omission for the benefit
of reviewers and developers.

*All pull requests*
- [ ] Builds successfully (*must test on at least one system/compiler combination*)
  - Operating systems 
    - [ ] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Compiles without warnings
- [ ] Passes all unit tests
- [ ] Passes all regression tests
- [ ] Documentation builds without errors

*For new code updates*
- [ ] Documentation updates - additions to user, theory, and verification manuals
- [ ] New unit tests providing coverage for new code, bug fixes etc.
- [ ] New regression tests exercising the new code

PR #676, opened 2020-08-12
Fix build errors from latest trilinos updates

sayerhs:
See #674 for details

### DIFF
--- a/include/EntityExposedFaceSorter.h
+++ b/include/EntityExposedFaceSorter.h
@@ -14,6 +14,7 @@
 
 #include <stk_mesh/base/BulkData.hpp>
 #include <stk_mesh/base/MetaData.hpp>
+#include <stk_mesh/base/EntitySorterBase.hpp>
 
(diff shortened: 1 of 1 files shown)


sayerhs:
@alanw0 found it easier to just cherry-pick your commit instead of fiddling with the modified submodule state.

alanw0:
I got a build error with my PR this morning also. Is it a chicken-and-egg issue where this change is necessary to work with the latest trilinos but trilinos hasn't been updated because it wouldn't build without this change?

sayerhs:
@alanw0 Yes, I need to update the trilinos docker image that is used for CI tests. I'll trigger this docker update so that subsequent CI updates work properly

alanw0:
ok. once this is in, I'll be interested to see how the nalu-wind tests run. In particular on a mac/clang platform. Stefan is seeing errors beneath stk-io, but only on mac clang, not on his gcc and intel builds.

PR #681, opened 2020-08-21
Refactor generic loop functor to void/lambda

psakievich:
@jvo1012 let me know if this PR resolves the kokkos error we were seeing yesterday. Unit tests and regression tests pass on my laptop build.


sayerhs:
Superseded by #687

PR #684, opened 2020-08-27
Shared import

rcknaus:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

More for discussion here.

Performs import from owned to a shared-not-owned solution before copying the tpetra vector(s) to the stk field.

Generally, this combines the periodic and shared communication step since we add the periodic dofs that don't have a primary on the process to the shared-not-owned map along with the regular shared-not-owned dofs.  I'm not really sure this is a good idea, but should make it so we can reduce the number of syncs (and the size of the communication a little) we make.  

This approach has some draw-backs---we use some more storage and we write to the primary nodes twice if they have a replica on the process.  It doesn't really improve anything if there's no periodic dofs.  I'm not really sure it handles all of wind's use cases.   Just thought I'd put it up as an idea.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [x] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [x] NVIDIA CUDA
- [x] Passes all unit tests
- [x] Passes all regression tests (don't have hypre hooked up, so some were skipped in my local run)

*For new code updates*

### DIFF
--- a/include/HypreLinearSystem.h
+++ b/include/HypreLinearSystem.h
@@ -233,6 +233,9 @@ public:
   virtual void buildDirichletNodeGraph(const std::vector<stk::mesh::Entity>&);
   virtual void buildDirichletNodeGraph(const stk::mesh::NgpMesh::ConnectedNodes);
 
+  virtual bool needs_communication_of_delta_on_periodic_nodes() const { return true; }
+
+
   sierra::nalu::CoeffApplier* get_coeff_applier();
 
   /***************************************************************************************************/
--- a/include/LinearSystem.h
+++ b/include/LinearSystem.h
@@ -113,7 +113,10 @@ public:
   virtual void buildDirichletNodeGraph(const stk::mesh::NgpMesh::ConnectedNodes) {}
 
   // Matrix Assembly
-  virtual void zeroSystem()=0;
+  virtual void zeroSystem() = 0;
+
+  virtual bool needs_communication_of_delta_on_periodic_nodes() const  { return false; };
+
 
 
   virtual CoeffApplier* get_coeff_applier()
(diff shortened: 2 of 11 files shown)


psakievich:
Interesting idea. I'm surprised to see just how many parallel communications you avoid. Seems like it could pay off depending on the communication cost. I haven't been following the gpu to gpu communication progress, but if we still have to copy to host to do our communication the it seems like it'd be a big win. Have you run any tests to compare the two yet?

sayerhs:
@rcknaus This looks like a good idea that could benefit ABL precursor simulations. As @psakievich suggests, it would be great if we can document gains for the standard 5x5x1 km domain at 10m resolution. 

@psakievich For STK field parallel operations, the STK team has found it almost always faster to do the `device - host - mpi - host - device` than doing CUDA-aware transfers. I don't think that has changed still.

rcknaus:
I can set up a 10m ABL but I probably won't get to it for a little while.  I did try it on a 384^3 taylor green vortex when I was working on this and saw about a 20-25% speed up on 4x5 GPUs.

rcknaus:
I ran the 10m 5x5x1 ABL with 5 nodes with 4 volta gpus each.  Before applying this change, the average time-per-timestep (neglecting the first and last steps) was 6.78 seconds.  After applying, it was 5.42 seconds for a speedup of 25%.  The change causes a diff since things are ordered differently, but it's 4.69e-11 after 100 timesteps.

sayerhs:
@rcknaus Do you want open this up from Draft to review?

rcknaus:
This probably needs some more thought / testing to flesh out for use in the main code base.   I think I'll leave this on the backburner for a bit, but I might return to it later.

PR #687, opened 2020-09-03
Add calls to modify and sync to get ActLineSimpleNGP test working on GPU

jvo1012:
Also update tolerances for linear solver and solution options to use edge scheme
in input file, and rebless the test

**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*


psakievich:
@jvo1012 so with these changes we do not need #681?

reg_tests/test_files/ActLineSimpleNGP/ActLineSimpleNGP.i
```diff
@@ -40,17 +40,17 @@ linear_solvers:
   type: tpetra
   method: gmres
   preconditioner: sgs
-  tolerance: 1e-5
-  max_iterations: 50
+  tolerance: 1e-12
```
psakievich, line 43: What is the timing difference with these solver tolerances?  I'm not sure we want to modify the regression test with these tight of tolerances since it will affect nightly/CI times.

psakievich (commented):
Should we also add this test to the nightly GPU runs in the CTestLists.cmake as well?

reg_tests/test_files/ActLineSimpleNGP/ActLineSimpleNGP.i
```diff
@@ -40,17 +40,17 @@ linear_solvers:
   type: tpetra
   method: gmres
   preconditioner: sgs
```
sayerhs, line 42:
```suggestion
  preconditioner: mt_sgs
```

@jvo1012 we want to use `mt_sgs` on GPUs. Can you please check and rebless the gold files with this setting?

reg_tests/test_files/ActLineSimpleNGP/ActLineSimpleNGP.i
sayerhs: @psakievich For NGP tests that run on both CPU and GPU, we use tight tolerances in linear solvers so that we can still use CPU golds to check the GPU runs to a tolerance tighter than `1e-8`. So this change is appropriate and is being done for all other GPU tests.

sayerhs:
@jvo1012 We should add this to the set of GPU nightly tests. Please add a line here https://github.com/Exawind/nalu-wind/blob/293b28afd993c36fe2343cd8d5b55f42f47f0637/reg_tests/CTestList.cmake#L238-L241

reg_tests/test_files/ActLineSimpleNGP/ActLineSimpleNGP.i
psakievich: @sayerhs okay thanks for the clarification

reg_tests/test_files/ActLineSimpleNGP/ActLineSimpleNGP.i
jvo1012: Done

jvo1012:
> @jvo1012 We should add this to the set of GPU nightly tests. Please add a line here
> 
> https://github.com/Exawind/nalu-wind/blob/293b28afd993c36fe2343cd8d5b55f42f47f0637/reg_tests/CTestList.cmake#L238-L241

Done

jvo1012:
> @jvo1012 so with these changes we do not need #681?

That's right, #681 is not necessary to get the test working on the GPU

psakievich (approved):

reg_tests/CTestList.cmake
```diff
@@ -238,6 +238,7 @@ else(NOT ENABLE_CUDA)
   add_test_r(ablNeutralNGPTrilinos 2)
   add_test_r(conduction_p4 2)
   add_test_r(airfoilRANSEdgeNGPTrilinos.rst 1)
+  add_test_r(ActLineSimpleNGP 4)
```
sayerhs, line 241: Please use only 2 GPUs, and convert the CPU test to be also 2 MPI ranks. We are not currently setup to assign multiple MPI ranks per GPU with our test setup.

sayerhs (changes requested):

reg_tests/CTestList.cmake
jvo1012: Done

alanw0 (approved):
Looks good to me.

PR #688, opened 2020-09-03
Make nalu-wind c++14 compatible

psakievich:
I wanted to open the discussion on moving to c++14.  Not sure if we're ready to move to standards yet, but it seems like this is all it would take.


alanw0 (commented):
Seems like a good change to me.
I would like to see a review from Shreyas or Jon Rood in case they know of any platforms that we support where c++14 isn't available yet.

sayerhs (approved):

sayerhs:
@alanw0 We are using at least GCC 7.4.0 on all the platforms we work on, and C++14 standard is supported. We still retain GCC 4.8.5 on CDash, but I believe the changes @psakievich are also supported in C++14.

overfelt (approved):

jrood-nrel:
The lowest we go is GCC 4.9.4.

sayerhs:
@psakievich I am going to revert this commit as it is breaking nightly tests and builds on NREL Eagle ([CDash](https://my.cdash.org/index.php?project=Nalu-Wind&date=2020-09-07)). I'll reinstate this commit once @jrood-nrel and I can figure out what is going on with NREL builds. Sorry about this.

PR #689, opened 2020-09-03
Hypre linear system assembly fixes for dirichlet rows

PaulMullowney:
This branch fixes bugs observed in many core Hypre runs. Dirichlet nodes were in the ghost cells were being accumulated multiple times to the matrix and rhs and the skipped rows logic was not catching this. The fix is to ensure that in the shared rows of the matrix, that row actually exists and is a dirichlet row i.e. not accumulated through the normal Assemble*Algorithm mechanism.

**Pull-request type:**
- [x] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/HypreLinearSystem.C
+++ b/src/HypreLinearSystem.C
@@ -1506,6 +1506,8 @@ HypreLinearSystem::HypreLinSysCoeffApplier::sum_into(
 	const double* cur_lhs = &lhs(ir, 0);
 	hid = localIds[ir];
 	
+	if (!map_shared_.exists(hid)) continue;
+
 	/* Find the index of the row */
 	unsigned index = map_shared_.value_at(map_shared_.find(hid));
 	unsigned lower = mat_row_start_shared_(index);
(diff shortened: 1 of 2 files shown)


src/HypreLinearSystem.C
```diff
@@ -1583,7 +1586,7 @@ HypreLinearSystem::HypreLinSysCoeffApplier::sum_into_1DoF(
       for (unsigned k=0; k<numEntities; ++k) {
 	/* binary search subrange rather than a map.find */
 	HypreIntType col = localIds[k];
-	unsigned matIndex=globalNumRows_+1;
+	unsigned matIndex;
```
alanw0, line 1589: I'm puzzled by this change. Why not initialize matIndex?
Not a big deal, I guess it would be a non-issue if matIndex was a return-value from binarySearchShared, rather than an output argument.

alanw0 (approved):

src/HypreLinearSystem.C
PaulMullowney: If binarySearchShared does not return successfully with a "reasonable value", then there are much bigger problems. I'm reworking this code in another branch. I didn't want to conflate the bug fix which is addressed by the map_shared_.exists() additions with other refactoring.

PaulMullowney: For now, I'm trying to keep the code consistent, i.e. uninitialized. This occurs a couple of other places in HypreLinearSystem and HypreUVWLinearSystem.

PR #690, opened 2020-09-09
Update openfast input files

psakievich:
Update openfast input files to work off dev branch.  Will require updating openfast builds.


jrood-nrel (approved):

psakievich:
As a side note, I think we should come up with some plan for mitigating the sensitivity of these tests to OpenFAST changes.  The fact that these tests are not backwards compatible and are so sensitive to feature changes in OpenFAST is a little troubling to me. Since OpenFAST's release cycle is so infrequent just building a release version doesn't seem tractable to me.  At the very least I think we should maintain a reference sha for openfast to know what we are testing against.

tonyinme:
Looks good.

The inconsistent input files is a problem in openfast with the way the input files are handled.
This is a known problem and there has been some talk to change this.
However, I don't think there is a plan to change it for now.

tonyinme (approved):

PR #699, opened 2020-09-22
Add in WALE and smagorinsky for matrix-free

rcknaus:
**Pull-request type:**
- [x] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

Adds in a couple of turbulence models for matrix free.  Generally, there are a couple options for computing an accurate gradient for these models:
* Compute the gradient locally on the element
* Solve a projection equation to get a continuous gradient at the nodes

This is implements the first.  The second would allow reuse of turbulence models but the inversion is fairly expensive and behaves similarly to the first option in testing.  A third option would be lump the projection equation, so it's just diagonal.  This loses high-order accuracy but is justifiable from a turbulence modeling perspective.  The lumped projection is equivalent in CVFEM to explicitly passing a discrete box-filter over the high-order accurate gradient.  This would allow reuse of turbulence models with the main nalu, is cheap, and does well in practice but requires some extra explanation. 

No regression tests for this yet.  There was a bug for for rho != 1 as well.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [x] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [x] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*


alanw0 (approved):
Nice change, I like seeing a named variable rather than a hard-coded false.

alanw0 (approved):

psakievich (approved):
Nice work!

Issue #701, opened 2020-09-22
Clang issues


jrood-nrel:
Looks like two problems. I'm not sure how to describe them except viewing them here: https://my.cdash.org/viewTest.php?onlyfailed&buildid=1888153

alanw0:
@jrood-nrel It looks like the unit-test is a matrix-free issue for Robert @rcknaus to look at.
The other test is throwing a std::bad_cast but I can't see any stack-trace information even though it appears to be a debug build. Is there any way to get more information about that?

jrood-nrel:
I just ran this test manually and got this stack trace. I wonder why it doesn't show me this on the dashboard.
```
60:     #2 0x15b1d6d in sierra::nalu::SolutionOptions::load(YAML::Node const&) /home/jrood/exawind/nalu-wind/src/SolutionOptions.C:377:50
60:     #2 0x15b1d6d in sierra::nalu::SolutionOptions::load(YAML::Node const&) /home/jrood/exawind/nalu-wind/src/SolutionOptions.C:377:50
60:     #3 0x14b258f in sierra::nalu::Realm::load(YAML::Node const&) /home/jrood/exawind/nalu-wind/src/Realm.C:806:21
60:     #4 0x155cec8 in sierra::nalu::Realms::load(YAML::Node const&) /home/jrood/exawind/nalu-wind/src/Realms.C:58:14
60:     #5 0xed1d73 in sierra::nalu::Simulation::load(YAML::Node const&) /home/jrood/exawind/nalu-wind/src/Simulation.C:105:12
60:     #6 0xe558e8 in main /home/jrood/exawind/nalu-wind/nalu.C:167:7
60:     #7 0x7fb550a1b554 in __libc_start_main (/lib64/libc.so.6+0x22554)
60:     #8 0xd27723 in _start (/home/jrood/exawind/nalu-wind/build/naluX+0xd27723)
```

jrood-nrel:
Ok, I tried a little harder to get the backtrace this time:
```
error: naluX :: Class '__vector_base<std::__1::basic_string<char>, std::__1::allocator<std::__1::basic_string<char> > >' has a base class 'std::__1::__vector_base_common<true>' which does not have a complete definition.
error: naluX :: Try compiling the source file with -fstandalone-debug.
* thread #1, name = 'naluX', stop reason = signal SIGABRT
  * frame #0: 0x00007fffe428f387 libc.so.6`__GI_raise + 55
[...]

PR #702, opened 2020-09-22
Fix some issues caught by clang

rcknaus:
**Pull-request type:**
- [x] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

Fixes https://github.com/Exawind/nalu-wind/issues/701
Missed guarding the local ids in one of the continuity operator tests.  Fix a bad cast to VectorFieldType.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [x] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/MatrixFreeLowMachEquationSystem.C
+++ b/src/MatrixFreeLowMachEquationSystem.C
@@ -121,17 +121,15 @@ void
 MatrixFreeLowMachEquationSystem::register_copy_state_algorithm(
   std::string name, int length, stk::mesh::Part& part)
 {
-  auto* field =
-    meta_.get_field<VectorFieldType>(stk::topology::NODE_RANK, name);
+  auto* field = meta_.get_field(stk::topology::NODE_RANK, name);
   auto copy = new CopyFieldAlgorithm(
-    realm_, &part, &field->field_of_state(stk::mesh::StateNP1),
-    &field->field_of_state(stk::mesh::StateN), 0, length,
-    stk::topology::NODE_RANK);
+    realm_, &part, field->field_state(stk::mesh::StateNP1),
+    field->field_state(stk::mesh::StateN), 0, length, stk::topology::NODE_RANK);
   copyStateAlg_.push_back(copy);
 
   auto copyn = new CopyFieldAlgorithm(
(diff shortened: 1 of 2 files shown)


src/MatrixFreeLowMachEquationSystem.C
```diff
@@ -121,17 +121,15 @@ void
...
 {
-  auto* field =
-    meta_.get_field<VectorFieldType>(stk::topology::NODE_RANK, name);
+  auto* field = meta_.get_field(stk::topology::NODE_RANK, name);
   auto copy = new CopyFieldAlgorithm(
-    realm_, &part, &field->field_of_state(stk::mesh::StateNP1),
-    &field->field_of_state(stk::mesh::StateN), 0, length,
-    stk::topology::NODE_RANK);
+    realm_, &part, field->field_state(stk::mesh::StateNP1),
+    field->field_state(stk::mesh::StateN), 0, length, stk::topology::NODE_RANK);
```
alanw0, line 127: So here you are switching from the Field<T> derived type to the FieldBase base class? Is this to fix the std::bad_cast error? I'm puzzled by that.

rcknaus: Yeah, just using the base class here (`CopyFieldAlgorithm` just takes the base class anyway).  I was passing both fields of different derived types (`VectorFieldType` and `ScalarFieldType`) to this algorithm but casting to `VectorFieldType`.

alanw0: Interesting. So in this case clang was actually catching a bad cast that other compilers were ignoring?

alanw0 (approved):

src/MatrixFreeLowMachEquationSystem.C
rcknaus: I think it's probably the address sanitizer being active that allowed it to be caught (I build with clang, though not clang-7).  Not sure why it wasn't a sanitizer error directly, though.

jrood-nrel (approved):

PR #717, opened 2020-10-06
Implement FLLC for newer Actuator Models

psakievich:
This draft PR is for implementing the FLLC for the NGP actuator models.  My goal is to get this working with ActLineSimpleNGP first so we can reproduce the results from [Martinez-Tossas and Meneveau, 2019](https://www.cambridge.org/core/journals/journal-of-fluid-mechanics/article/filtered-lifting-line-theory-and-application-to-the-actuator-line-model/EA90C34B7D48030D725B3AFDC06501DF) for verification.  Then adapting to OpenFAST based models should be straight forward.

This work is to support the Sandia V&V UQ team's efforts.

Planned implementation process:

- [x]  Create unit tests for each of the implementation equations from section 5 of the reference paper
- [x]  Create a regression test for ActLineSimpleNGP to reproduce the paper results
- [x]  Run verification/grid convergence
- [ ]  Create unit tests for OpenFAST Actutator Models
- [ ]  Convert OpenFAST FLLC regression test to use the new model


psakievich:
I think this is ready to merge.  The next step of refactoring for the OpenFAST actuators is a little more invasive than I intended because it seems like it will be best to refactor the way the new actuators are constructed in Realm.C as part of the process.  I've started working on it, but I think we should merge this and open a second PR to keep this one from getting too huge. Thoughts @sayerhs @tonyinme @lawrenceccheung ?

tonyinme (approved):
This is working well.
Tested on a fixed wing case and results match well with theory.

sayerhs:
Please don't merge this until we have stabilized the nightly test dashboard. we are still trying to fix issues introduced by #725

PR #729, opened 2020-10-20
Transition overset interface to NGP

sayerhs:
This PR is a first in the series of updates to overset interface so that we can access the NGP API of TIOGA when running on GPUs. Changes

- Update data structures that were using `std::vector` and `std::map` in Tioga interface classes to Kokkos::Views
- Update the code to consistently manage host/device views of the data
- Register the host views using the current interface
- Fix code in TimeIntegrator so that tioga instance is no longer created for non-overset runs

I decided to keep the code simple by explicitly handling View+HostMirror instead of using `Kokkos::DualView`, but it should be fairly easy to switch these out in future if the need arises. 

## Checklist

*All pull requests*
  - Operating systems 
    - [X] Linux
    - [ ] MacOS
  - Compilers 
    - [X] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [X] NVIDIA CUDA
- [X] Passes all unit tests

### DIFF
--- a/src/overset/TiogaOptions.C
+++ b/src/overset/TiogaOptions.C
@@ -34,6 +34,13 @@ void TiogaOptions::load(const YAML::Node& node)
     hasMexclude_ = true;
     mExclude_= node["num_exclude"].as<int>();
   }
+
+  if (node["cell_resolution_multiplier"]) {
+    cellResMult_ = node["cell_resolution_multiplier"].as<double>();
+  }
+  if (node["node_resolution_multiplier"]) {
+    nodeResMult_ = node["node_resolution_multiplier"].as<double>();
+  }
 }
 
 void TiogaOptions::set_options(TIOGA::tioga& tg)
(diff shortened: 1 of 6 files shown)


src/overset/TiogaOptions.C
```diff
@@ -34,6 +34,13 @@ void TiogaOptions::load(const YAML::Node& node)
...
     mExclude_= node["num_exclude"].as<int>();
   }
+
+  if (node["cell_resolution_multiplier"]) {
+    cellResMult_ = node["cell_resolution_multiplier"].as<double>();
+  }
+  if (node["node_resolution_multiplier"]) {
+    nodeResMult_ = node["node_resolution_multiplier"].as<double>();
+  }
 }
```
ashesh2512, line 44: This is very helpful for giving the user the flexibility to hard-code node and cell resolution through the input file. Helps a great deal when debugging!

ashesh2512 (approved):

alanw0 (approved):
I don't have a solid understanding of this code, but I don't see any obvious problems.

PR #730, opened 2020-10-21
Overset: Adjust resolutions for mandatory fringes

sayerhs:
Adds an option to TIOGA overset connectivity logic that is used to provide 
hints to TIOGA regarding nodes/cells that are adjacent to the overset boundaries
and must be ignored when considering potential donors. This is done by setting
a high value of nodal/cell resolutions, the criteria that TIOGA uses to determine
potential donors.

## Checklist

*All pull requests*
  - Operating systems 
    - [X] Linux
    - [X] MacOS
  - Compilers 
    - [X] GCC
    - [X] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [X] Passes all unit tests
- [X] Passes all **overset** regression tests

### DIFF
--- a/include/overset/TiogaBlock.h
+++ b/include/overset/TiogaBlock.h
@@ -30,24 +30,78 @@ typedef stk::mesh::Field<int> ScalarIntFieldType;
  */
 struct NgpTiogaBlock
 {
+  /** Number of cell types supported by TIOGA
+   *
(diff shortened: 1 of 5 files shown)


ashesh2512 (approved):
The comments are really helpful, thanks!

PR #731, opened 2020-10-21
Add user option to reuse linear systems for decoupled overset runs

sayerhs:
Introduces an option in linear systems section to control whether the equation
systems should attempt to reuse linear system graph structures (avoid init costs)
when performing decoupled overset solves with mesh motion.

For the `oversetRotCylinderNGPTrilinos` unit test

With linsys reinit:    74.9 s
Without linsys reinit: 42.3 s

## Checklist

*All pull requests*
  - Operating systems 
    - [X] Linux
    - [X] MacOS
  - Compilers 
    - [X] GCC
    - [X] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [X] Passes all unit tests
- [X] Passes all **overset** regression tests

### DIFF
--- a/include/LinearSolverConfig.h
+++ b/include/LinearSolverConfig.h
@@ -65,6 +65,16 @@ public:
   inline bool useSegregatedSolver() const
   { return useSegregatedSolver_; }
 
+  /** User flag indicating whether equation systems must attempt to reuse linear
+   *  system data structures even for cases with mesh motion.
+   *
+   *  This option only affects decoupled overset system solves where the matrix
+   *  graph doesn't change, only the entries within the graph. This can be
+   *  controlled on a per-solver basis.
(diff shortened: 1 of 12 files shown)


alanw0 (approved):
I think this looks fine.
I recall we also had some logic about reusing the preconditioner. Is this tied to that as well?

sayerhs:
@alanw0 Yes, there are `recompute_preconditioner` and `reuse_preconditioner` options (that have been there since beginning of ExaWind project, if I remember correctly). Those options are unaffected. This allows the user to bypass the `delete linsys_` calls and therefore not have to revisit all the `build*NodeGraph` and `finalizeLinearSystem()` methods in `LinearSystem`.

PR #735, opened 2020-10-27
Alternate approach to fixing node resolutions for TIOGA connectivity

sayerhs:
This PR addresses some of the issues found with fringe/field mismatch in overset connectivity when using TIOGA with RCB/RIB type decompositions

## Checklist

*All pull requests*
  - Operating systems 
    - [X] Linux
    - [X] MacOS
  - Compilers 
    - [X] GCC
    - [X] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [X] Passes all unit tests
- [X] Passes all regression tests (**overset** tests and matches CDash status)

### DIFF
--- a/src/overset/TiogaBlock.C
+++ b/src/overset/TiogaBlock.C
@@ -88,9 +88,13 @@ void TiogaBlock::setup(stk::mesh::PartVector& bcPartVec)
   ScalarIntFieldType& ibcell = meta_.declare_field<ScalarIntFieldType>(
     stk::topology::ELEM_RANK, "iblank_cell");
 
+  ScalarFieldType& nVol = meta_.declare_field<ScalarFieldType>(
+    stk::topology::NODE_RANK, "tioga_nodal_volume");
+
   for (auto p: blkParts_) {
     stk::mesh::put_field_on_mesh(ibf, *p, nullptr);
     stk::mesh::put_field_on_mesh(ibcell, *p, nullptr);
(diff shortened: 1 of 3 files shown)


ashesh2512 (approved):

src/overset/TiogaBlock.C
```diff
@@ -274,35 +278,57 @@ void TiogaBlock::adjust_resolutions()
...
   auto& eidmap = bdata_.eid_map_.h_view;
   auto& cellres = bdata_.cell_res_.h_view;
-  auto& noderes = bdata_.node_res_.h_view;
+
   for (auto b: mbkts) {
-    for (size_t fi=0; fi < b->size(); ++fi) {
-      const auto face = (*b)[fi];
-      const auto* elems = bulk_.begin_elements(face);
-      const auto num_elems = bulk_.num_elements(face);
+    for (size_t in =0; in < b->size(); ++in) {
```
alanw0, line 289: You could do this here:
for(auto node : *b) {
if you don't need the index 'in' for anything else.

alanw0 (approved):

src/overset/TiogaBlock.C
sayerhs: Hold the presses! @alanw0 recommending `auto`??? Addressed in 1ae8ce3

alanw0: Arrggh! This would be better: 'for(stk::mesh::Entity node : *b) {'

PR #744, opened 2020-11-09
CUDA enhancements and fixes for warnings

sayerhs:
- Use `enable_language` instead of `find_package` for CUDA

- For CUDA versions 11.0.2 and above enable `-Wext-lambda-captures-this` flag which will warn users when a local copy of class instance variables are not created

- Fix long standing warnings with fallthrough in linear interpolation switch statement

**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [X] Feature enhancement

## Checklist

*All pull requests*
  - Operating systems 
    - [X] Linux
    - [ ] MacOS
  - Compilers 
    - [X] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [X] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*


alanw0 (approved):
Aren't you off this week?

PR #745, opened 2020-11-09
Update CMake logic for hypre 

sayerhs:
This PR updates the CMake logic to search for hypre library and simplifies the logic related to init/finalize in CUDA builds

- Remove check for COGMRES. Instead require at least v2.18.2 for HYPRE from now on
- Use CMake named target to include HYPRE libraries and include directories
- Support building all combinations of host-only/gpu versions of hypre and nalu-wind


jrood-nrel (approved):

PaulMullowney:
This seems reasonable. The cogmres interface may not really be that useful. I've never seen any impact from that algorithm.

PaulMullowney (approved):

PR #758, opened 2020-11-12
Overset: update hybrid solver solution exchange interface

sayerhs:
This PR updates the hybrid overset interface in Nalu-Wind to explicitly take a list of fields to exchange to allow more flexibility when coupling with AMR-Wind. 

Currently, the field exchange is only performed for fields registered by the equation systems and their order is determined by the order in which these equations have been specified in the input file. This would lead to inconsistencies if AMR-Wind has equations in a different order. Another reason is that AMR-Wind has cell-centered fields and one node-centered field (pressure). So the exchange of solutions has to be treated specially to deal with the differences between the solver. 

This modification doesn't affect any of the regression tests in nalu-wind as they do not use the external interface. 

## Checklist

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [X] MacOS
  - Compilers 
    - [ ] GCC
    - [X] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [X] Passes all unit tests

### DIFF
--- a/include/overset/ExtOverset.h
+++ b/include/overset/ExtOverset.h
@@ -53,7 +53,7 @@ public:
   void post_overset_conn_work();
 
   //! Register solution fields to TIOGA before interpolation step
-  int register_solution();
+  int register_solution(const std::vector<std::string>& fnames);
 
   //! Update solution fields after TIOGA has performed interpolations
   void update_solution();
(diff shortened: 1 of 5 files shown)


alanw0 (approved):
Yes I think this looks reasonable.

PR #760, opened 2020-11-12
Resolve warnings in Apple-clang

psakievich:

I got tired of the seeing all the warnings in about deleted constructors. This compiles with `-Werror` on my Mac at least for the base code and OpenFAST turned on.  I'll do a follow up PR with other TPL's turned on.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [x] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [x] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/NaluParsing.C
+++ b/src/NaluParsing.C
@@ -171,7 +171,7 @@ namespace sierra
       {
         case OversetBoundaryConditionData::TPL_TIOGA:
 #ifdef NALU_USES_TIOGA
-          oversetBC.userData_.oversetBlocks_ = oversetUserData;
+          oversetBC.userData_.oversetBlocks_ = node["overset_user_data"];
           break;
 #else
           throw std::runtime_error(
--- a/src/NaluParsing.C
+++ b/src/NaluParsing.C
@@ -167,8 +167,6 @@ namespace sierra
         }
       }
 
-      const YAML::Node& oversetUserData = node["overset_user_data"];
-
       switch (oversetBC.oversetConnectivityType_)
       {
         case OversetBoundaryConditionData::TPL_TIOGA:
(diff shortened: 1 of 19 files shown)


src/NaluParsing.C
```diff
@@ -167,8 +167,6 @@ namespace sierra
         }
       }
 
-      const YAML::Node& oversetUserData = node["overset_user_data"];
```
sayerhs, line 170: This breaks builds with TIOGA enabled

sayerhs (changes requested):
Breaks build with TIOGA overset enabled.

psakievich:
Anyone know how `KOKKOS_FUNCTION constuctor() =delete;` will behave?  I'll test it out.  I'm guessing I should probably just remove the kokkos macros too though.

src/NaluParsing.C
psakievich: Ah oh yeah.  Okay I'll move this inside the `#ifdef`

psakievich: K pushed a fix for this.  I've added the WIP tag to prevent merging until I finish testing the cuda build.

alanw0:
@psakievich I know there is a KOKKOS_DEFAULTED_FUNCTION, perhaps there is a KOKKOS_DELETED_FUNCTION but I'm not sure. If the function is deleted then it doesn't seem like it should need to be decorated...

psakievich:
Well it looks like the cuda build doesn't care if you leave the macro there, but I think I will remove them all anyways because I think it's a better practice.

sayerhs (approved):

PR #763, opened 2020-11-15
test: Add relative tolerance for regression tests

sayerhs:
## Changes

- Introduce `TEST_REL_TOL` variable to control relative test tolerance for regression tests via CMake.

- Set default relative tolerance to 1.0e-13 for the GOLD system (Linux/GCC v.7.4.0) instead of the relaxed 1.0e-7 tolerance in the pass/fail python script.

- Fix unnecessary piping in pass/fail python script.

## Additional background 

With the new `pass-fail.py` script, the default relative tolerance is set to `1.0e-7`, this is too lax for the Linux/GCC v7.4.0 build and lets tests pass even if the norms do not match as show below [CDash link](https://my.cdash.org/test/15151477)

```
PASS: airfoilRANSEdgeNGPHypre.................   119.1670s 6.2707e-04 9.7410e-10
```

In the past, this would have indicated that the gold files should be updated after looking at the commit that caused this change. The changes in this PR is an attempt to catch such issues. 

## Checklist

*All pull requests*
  - Operating systems 
    - [X] Linux
    - [X] MacOS
  - Compilers 
    - [X] GCC
    - [X] LLVM/Clang
    - [X] Intel compilers
    - [ ] NVIDIA CUDA
- [X] Passes all unit tests

### DIFF
--- a/reg_tests/CMakeLists.txt
+++ b/reg_tests/CMakeLists.txt
@@ -17,7 +17,8 @@ set(NALU_GOLD_NORMS_DIR "${CMAKE_CURRENT_SOURCE_DIR}" CACHE PATH
 
 if (NOT ${NALU_GOLD_NORMS_DIR} STREQUAL ${CMAKE_CURRENT_SOURCE_DIR})
   message(STATUS "Setting custom gold norms path: ${NALU_GOLD_NORMS_DIR}")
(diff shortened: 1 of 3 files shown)


alanw0 (approved):
Great, this should help a lot.

PR #766, opened 2020-11-16
Add CUDA build to Github Actions

sayerhs:
This PR adds CUDA build line for GitHub CI. It uses CUDA v11.1. I originally wanted to use `--Werror ext-lambda-captures-this` but it currently fails for the master branch. We can turn that on once this issue has been addressed. Meanwhile, this should capture build issues on CUDA for pull requests and development branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,7 @@ on:
       - 'CMakeLists.txt'
       - 'CTestConfig.cmake'
       - 'Dockerfile'
+      - 'docker/**'
       - 'nalu.C'
       - 'unit_tests.C'
   pull_request:
@@ -25,13 +26,22 @@ on:
       - 'CMakeLists.txt'
       - 'CTestConfig.cmake'
       - 'Dockerfile'
+      - 'docker/**'
       - 'nalu.C'
       - 'unit_tests.C'
 
 jobs:
   build:
+    name: GCC-8.3.0
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - name: Build and test
         run: docker build -t nalu-wind .
+  cuda-build:
+    name: CUDA-11.1/GCC-8.3.0
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build and test
+        run: docker build -f ./docker/Dockerfile.cuda -t nalu-wind .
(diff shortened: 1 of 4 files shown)


sayerhs:
@PaulMullowney FYI, when using `--Werror ext-lambda-captures-this` `nvcc` doesn't like this line https://github.com/Exawind/nalu-wind/blob/master/src/HypreLinearSystem.C#L1682

alanw0 (approved):
I don't know anything about docker stuff, but I like this change, getting cuda in the mix will be great.

sayerhs:
@alanw0 So it seems that the linker step is failing on Github CI with the following error in `libamesos2.a` library 

```
error adding symbols: archive has no index; run ranlib to add one
```

I'll keep digging into this, but if there is some insight from trilinos team in why this could happen that would be great.

alanw0:
@jhux2 or @lucbv do you guys have any insights about the error Shreyas describes above?

sayerhs:
@jhux2 @lucbv Just note that I am using CUDA v11.1 for these builds because I want to be able to use `--Werror ext-lambda-captures-this` as part of the CI build so that we catch these silly errors during PR review phase.

lucbv:
@sayerhs @alanw0 
I am also not a docker expert but it looks like a pre-built version of Trilinos is used here?
At least I cannot find easily where the build info for Trilinos is located when I click on the `Details` for `Nalu-Wind-CI/CUDA-11.1/GCC-8.3.0`, can someone point me to the build configuration used for Trilinos?
This way I could try to reproduce on a testbed or Summit and diagnose the problem.
Also it would be interesting to see how that build configuration compares with the CI builds that Trilinos does for its PRs, the closer we can be the safest and easiest the process will be.

sayerhs:
@lucbv Here are the links: 

- [CMake configure](https://github.com/sayerhs/exawind-docker/blob/master/exw-trilinos-cuda/cmake-configure.sh) 
- The complete [trilinos build log](https://github.com/sayerhs/exawind-docker/runs/1408441816?check_suite_focus=true) 
- Trilinos CUDA [Dockerfile](https://github.com/sayerhs/exawind-docker/blob/master/exw-trilinos-cuda/Dockerfile)
- [Docker image](https://hub.docker.com/layers/exawind/exw-dev-cuda/latest/images/sha256-83545547df0a8ab8eb27b186830316b45f53052ff7a91df3c4a1932a9167797e?context=explore) on top of which Trilinos/Nalu

One thing I had to do to get Docker builds to work for Trilinos is that I had to set `-DTPL_ENABLE_CUSPARSE=OFF` for it to build. Not sure if this messes up amesos2.

lucbv:
Thanks, I will look into this and setup a reproducer, probably on Summit, so I can diagnose the problem.
I do not think that disabling cusparse should lead to trouble in Amesos2 but I can double check.

PR #775, opened 2020-11-19
Iddes les indicator

psakievich:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [s] Feature enhancement

Provide a field for checking LES/RANS regions in the iddes model.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [x] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/node_kernels/TKESSTIDDESNodeKernel.C
+++ b/src/node_kernels/TKESSTIDDESNodeKernel.C
@@ -16,9 +16,8 @@
 namespace sierra {
 namespace nalu {
 
-TKESSTIDDESNodeKernel::TKESSTIDDESNodeKernel(
-  const stk::mesh::MetaData& meta
-) : NGPNodeKernel<TKESSTIDDESNodeKernel>(),
+TKESSTIDDESNodeKernel::TKESSTIDDESNodeKernel(const stk::mesh::MetaData& meta)
+  : NGPNodeKernel<TKESSTIDDESNodeKernel>(),
     tkeID_(get_field_ordinal(meta, "turbulent_ke")),
     sdrID_(get_field_ordinal(meta, "specific_dissipation_rate")),
     densityID_(get_field_ordinal(meta, "density")),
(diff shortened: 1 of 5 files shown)


src/node_kernels/TKESSTIDDESNodeKernel.C
```diff
@@ -47,6 +47,7 @@ TKESSTIDDESNodeKernel::setup(Realm& realm)
   dualNodalVolume_ = fieldMgr.get_field<double>(dualNodalVolumeID_);
   maxLenScale_     = fieldMgr.get_field<double>(maxLenScaleID_);
   fOneBlend_       = fieldMgr.get_field<double>(fOneBlendID_);
+  lengthScaleRatio_ = fieldMgr.get_field<double>(lengthScaleRatioID_);
```
sayerhs, line 50:
```suggestion
  lengthScaleRatio_ = fieldMgr.get_field<double>(lengthScaleRatioID_);
  // call modify before this field gets modified in kernel execute phase
  lengthScaleRatio_.modify_on_device();
```

src/node_kernels/TKESSTIDDESNodeKernel.C
```diff
@@ -122,11 +123,14 @@ void TKESSTIDDESNodeKernel::execute(
...
   const DblType lSST = sqrtTke / betaStar_ / sdr;
+  const DblType lLES = cDES * delta;
 
   // Find minimum length scale, limit minimum value to 1.0e-16 to prevent
   // division by zero later on
   const DblType lIDDES =
-      stk::math::max(1.0e-16, fdHat * (1.0  + fe) * lSST + (1.0 - fdHat) * cDES * delta);
+    stk::math::max(1.0e-16, fdHat * (1.0 + fe) * lSST + (1.0 - fdHat) * lLES);
+
+  lengthScaleRatio_.get(node, 0) = lLES / lIDDES;
```
sayerhs, line 133:
```suggestion
  // modify on device called in setup
  lengthScaleRatio_.get(node, 0) = lLES / lIDDES;
```

sayerhs (commented):
The field registration and population logic looks fine for CPU runs. However, there is an issue when running on GPUs... we never call `modify_on_device()` so the field might not get synced back to host before being written to exodus files. 

I added a few comments where we could sneak in `modify_on_device()`, but happens in setup away from where the field is actually modified... which is sub-optimal. You can either do that, or merge as is and open an issue so that we don't forget about this when executing on GPUs.

psakievich:
Thanks @sayerhs good point.  I'll update the PR.  I like the idea of keeping it device compatible at all times.

Issue #743, opened 2020-11-09
Segfaults observed in GPU builds with  CUDA 10.2.89 


sayerhs:
GPU builds throw segfaults with latest Nalu-Wind/Trilinos builds:

# Build information

| | |
|---|---|
| Host compiler | GCC 8.4.0 |
| GPU compiler | CUDA 10.2.89 |
| Nalu-Wind Git SHA | 1ebacaebff8db6 |
| Trilinos Git SHA | 13.1-gc23783ec3ea | 

# Error with unit tests

This error is erratic... using `--gtest_shuffle` will sometimes allow regression tests to run successfully. Also running `LowMachKernelHex8Mesh.NGP_nodal_grad_popen` in standalone mode will succeed. 

```
[ RUN      ] LowMachKernelHex8Mesh.NGP_nodal_grad_popen
terminate called after throwing an instance of 'std::runtime_error'
  what():  cudaStreamSynchronize(m_stream) error( cudaErrorIllegalAddress): an illegal memory access was encountered ../packages/kokkos/core/src/Cuda/Kokkos_Cuda_Instance.cpp:312
Traceback functionality not available
```

```
MPT: #13 0x00002b73d502f615 in _Unwind_Resume ()
MPT:     at /scratch/jrood/.spack/stage/spack-stage-gcc-8.4.0-2a3vha6hlw4xc5ja3jyhr7huzaxuw2kt/spack-src/libgcc/unwind.inc:241
MPT: #14 0x0000000000ea8d2c in Kokkos::Impl::ViewValueFunctor<Kokkos::Cuda, sierra::nalu::FieldInfoNGP, false>::execute(bool) ()
MPT: #15 0x0000000000ea8eb2 in void Kokkos::Impl::(anonymous namespace)::deallocate<Kokkos::CudaSpace, Kokkos::Impl::ViewValueFunctor<Kokkos::Cuda, sierra::nalu::FieldInfoNGP, false> >(Kokkos::Impl::SharedAllocationRecord<void, void>*) ()
MPT: #16 0x00000000065c4cc7 in Kokkos::Impl::SharedAllocationRecord<void, void>::decrement(Kokkos::Impl::SharedAllocationRecord<void, void>*) ()
MPT: #17 0x000000000082db25 in sierra::nalu::ElemDataRequestsGPU::~ElemDataRequestsGPU() ()
[...]

Issue #778, opened 2020-11-22
STK warnings in CUDA 11 build


sayerhs:
```
/opt/exawind/include/stk_mesh/base/NgpFieldParallel.hpp(150): warning: __device__ annotation is ignored on a function("NgpFieldInfo") that is explicitly defaulted on its first declaration

/opt/exawind/include/stk_mesh/base/NgpFieldParallel.hpp(150): warning: __host__ annotation is ignored on a function("NgpFieldInfo") that is explicitly defaulted on its first declaration

/opt/exawind/include/stk_mesh/base/NgpFieldParallel.hpp(153): warning: __device__ annotation is ignored on a function("NgpFieldInfo") that is explicitly defaulted on its first declaration

/opt/exawind/include/stk_mesh/base/NgpFieldParallel.hpp(153): warning: __host__ annotation is ignored on a function("NgpFieldInfo") that is explicitly defaulted on its first declaration

/opt/exawind/include/stk_mesh/base/NgpFieldParallel.hpp(156): warning: __device__ annotation is ignored on a function("~NgpFieldInfo") that is explicitly defaulted on its first declaration

/opt/exawind/include/stk_mesh/base/NgpFieldParallel.hpp(156): warning: __host__ annotation is ignored on a function("~NgpFieldInfo") that is explicitly defaulted on its first declaration
```

tasmith4:
I think we fixed this a while ago.  @alanw0 please reopen if I'm wrong.

alanw0:
Yes I think this was fixed.

PR #780, opened 2020-11-24
Fix memory leak reported in CDash

sayerhs:
This commit fixes the memory leak reported in CDash while simultaneously ensuring that the MasterElement creation on device works in both CUDA 9.2.x and 10.2.x

See [CDash](https://my.cdash.org/test/16023552) for details of the memory leak. See #743 for discussion regarding issues with CUDA builds

### DIFF
--- a/include/master_element/Edge22DCVFEM.h
+++ b/include/master_element/Edge22DCVFEM.h
@@ -35,8 +35,7 @@ class Edge2DSCS : public MasterElement
 public:
   KOKKOS_FUNCTION
   Edge2DSCS();
-  KOKKOS_DEFAULTED_FUNCTION
-  virtual ~Edge2DSCS() = default;
+  KOKKOS_FUNCTION virtual ~Edge2DSCS() {}
   using AlgTraits = AlgTraitsEdge_2D;
   using MasterElement::shape_fcn;
   using MasterElement::determinant;
--- a/include/master_element/Edge32DCVFEM.h
+++ b/include/master_element/Edge32DCVFEM.h
@@ -41,8 +41,7 @@ public:
 
   KOKKOS_FUNCTION
   Edge32DSCS();
-  KOKKOS_DEFAULTED_FUNCTION
-  virtual ~Edge32DSCS() = default;
+  KOKKOS_FUNCTION virtual ~Edge32DSCS() {}
   using MasterElement::determinant;
   using MasterElement::shape_fcn;
   using MasterElement::shifted_shape_fcn;
--- a/include/master_element/Hex27CVFEM.h
+++ b/include/master_element/Hex27CVFEM.h
@@ -40,8 +40,7 @@ public:
 
   KOKKOS_FUNCTION
   HexahedralP2Element();
-  KOKKOS_DEFAULTED_FUNCTION
-  virtual ~HexahedralP2Element() = default;
+  KOKKOS_FUNCTION virtual ~HexahedralP2Element() {}
 
   void shape_fcn(double *shpfc);
   void shifted_shape_fcn(double *shpfc);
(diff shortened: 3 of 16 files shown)


psakievich:
This is interesting. @sayerhs I'm trying to wrap my head around this.  Could you explain a little more why the destructor change and manual calling fixes the leak? Is it some strange Cuda voodoo?

overfelt:
So, the root of the problem was the use of KOKKOS_DEFAULTED_FUNCTION?

overfelt (approved):

PaulMullowney (approved):
Looks good to me.

sayerhs:
@psakievich @overfelt 

In the following explanation it might be useful to refer to abc5449bba2f563 as the state of our original code where we observed #743. At this point, @overfelt had actually inserted an explicit destructor call when clearing out master elements

https://github.com/Exawind/nalu-wind/blob/abc5449bba2f563d1fea992949a5cf2c8edf23b4/src/master_element/MasterElementFactory.C#L187-L190

This is the line that caused the error in CUDA 10.2.89 that I posted in in #743. Removing this cleared the error in CUDA 10.2.89. This was because we were mixing `virtual ~MESubClass() = default;` and `virtual ~MESubClass() {}` and so whenever it hit the `= default` destructor (e.g., `Quad4SCS` it would crash).

This led me to think may be we should be consistent across all master elements and so I converted everything to `= default()` instead of empty definitions. The thinking was, better to be explicit state using default instead of providing empty body definitions. However, then calling the destructor on device caused crashes  (I only have observations on NREL Eagle that calling `default` virtual destructors on device crashes) So I thought hmmm... there is nothing to be cleaned up so why even bother to call the destructor because the destructor shouldn't be doing anything special. So I removed the explicit all to destructor in 04a72728b. This seemed to fix issues on CUDA 9.2 and 10.2.89. 

However, the next day LLVM address sanitizer started complaining about memory leaks... because the higher order elements (e.g., Hex27) use Kokkos::Views as class members, so when using placement new we have to explicitly call the destructors before releasing the memory as C++ doesn't automatically call the destructors on objects that were created using placement new. If we don't do this, the Kokkos::View destructors are never called. So the situation is that we do need those destructors like James originally intended. However, if it is defaulted CUDA 10.2 will promptly crash there.. so this pull request switched back to using `virtual ~MESubClass() {}` consistently for the entire ME class hierarchy and now CUDA doesn't barf when calling the explicit destructor. 

Regarding `KOKKOS_DEFAULTED_FUNCTION` – In CUDA 11.1, we see warnings if you do `KOKKOS_FUNCTION Class() = default` (same for destructor) -- see #778. So we need to use `KOKKOS_DEFAULTED_FUNCTION` when using `= default;`

However, the entire thing is still unsatisfactory, even though it works. We still have two mysteries that we don't have good answer to. 

1. Why is the change to `MasterElementFactory.h` in 04a72728b7fe required? Without this, `naluX` and `unittestX` will fail in `Kokkos::parallel_for` within `create_device_expression` 

2. @PaulMullowney reports that the change to `NaluParsing.h` in 5497fe0c66db introduces a severe performance regression for NREL5MW GPU runs. Restoring that header in `NaluParsing.h` recovers performance. 

Both these things lead me to suspect that we might have some other memory issue that we are working around with these bandaids. I don't think we have hit upon the root cause yet. 

@PaulMullowney any thing to add here?

psakievich:
Nice.  @sayerhs thanks for the thorough explanation.  That clears things up for me.

PaulMullowney:
With the regards to the performance regression that was mentioned, it was the removal of 1 header file that caused the slowdowns. This was observed in both Hypre and Trilinos solver stacks. The regression is truly mind boggling--I agree with @sayerhs that it suggests another issue that we're just not seeing. In any event, I'm glad I reran the scaling studies on summit yesterday. It was only 7 step bisection rather than something much more time consuming.

PR #781, opened 2020-11-25
Add some Iddes fields and bug fixes

psakievich:
**Pull-request type:**
- [x] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

Fix some bugs in IDDES, add a RANS region indicator, and `max_peclet_factor`. to indicate the level of upwinding.
I ran the `IDDESPeriodicHillEdge` regression test, but not the entire suite.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [x] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/edge_kernels/MomentumEdgePecletAlg.C
+++ b/src/edge_kernels/MomentumEdgePecletAlg.C
@@ -110,7 +110,6 @@ void
 DetermineMaxPecletFactor(
   stk::mesh::BulkData& bulk, const stk::mesh::MetaData& meta)
 {
-  // fields
   ScalarFieldType* maxPecFac = meta.get_field<ScalarFieldType>(
     stk::topology::NODE_RANK, "max_peclet_factor");
   ScalarFieldType* pecletFactor =
@@ -118,25 +117,22 @@ DetermineMaxPecletFactor(
 
   stk::mesh::field_fill(0.0, *maxPecFac);
 
-  // selector
   const stk::mesh::Selector sel =
     stk::mesh::selectField(*pecletFactor) & meta.locally_owned_part();
 
-  // buckets
   for (const auto* ib : bulk.get_buckets(stk::topology::EDGE_RANK, sel)) {
     const auto& b = *ib;
     const size_t length = b.size();
     for (size_t k = 0; k < length; ++k) {
       stk::mesh::Entity edge = b[k];
-      const double pecFac = *(stk::mesh::field_data(*pecletFactor, edge));
+      const double* pecFac = stk::mesh::field_data(*pecletFactor, edge);
       const auto* nodes = bulk.begin_nodes(edge);
       double* maxPecL = stk::mesh::field_data(*maxPecFac, nodes[0]);
       double* maxPecR = stk::mesh::field_data(*maxPecFac, nodes[1]);
-      *maxPecL = std::max(*maxPecL, pecFac);
-      *maxPecR = std::max(*maxPecR, pecFac);
+      *maxPecL = std::max(*maxPecL, *pecFac);
+      *maxPecR = std::max(*maxPecR, *pecFac);
     }
   }
-  // loop
   stk::mesh::copy_owned_to_shared(bulk, {maxPecFac});
 }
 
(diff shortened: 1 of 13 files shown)


psakievich:
Putting a WIP tag on this while I try to resolve the parallel inconsistencies I'm seeing across processor boundaries for the `iddes_rans_indicator` field

psakievich:
Okay I think it's ready for review

src/edge_kernels/MomentumEdgeSolverAlg.C
```diff
@@ -126,6 +135,8 @@ MomentumEdgeSolverAlg::execute()
 
       const DblType pecfac = pecletFactor.get(edge, 0);
       const DblType om_pecfac = 1.0 - pecfac;
+      maxPeclet.get(nodeR, 0) = stk::math::max(maxPeclet.get(nodeR, 0), pecfac);
+      maxPeclet.get(nodeL, 0) = stk::math::max(maxPeclet.get(nodeL, 0), pecfac);
```
sayerhs, line 139: This most likely won't do what you expect it to do on GPUs.

src/node_kernels/TKESSTIDDESNodeKernel.C
```diff
@@ -106,15 +106,15 @@ void TKESSTIDDESNodeKernel::execute(
   omegaSq *= 2.0;
   Pk *= tvisc;
 
-  DblType rdl = visc/(density * kappa_ * kappa_ * dw * dw * stk::math::sqrt(0.5 * (sijSq + omegaSq) ));
+  DblType rdl = visc/(density * kappa_ * kappa_ * dw * dw * stk::math::sqrt(0.5 * (sijSq + omegaSq) ) + 1e-10);
```
tonyinme, line 109: I am not entirely sure why this 1e-10 is here. It would be good to double-chek with @gantech if this is the right thing to do?

tonyinme (commented):
The main bugs have been addressed. 
I was wondering if it would be worth changing the variables used here to be consistent with the paper notation (CDES1 and CDES2): 
https://github.com/Exawind/nalu-wind/blob/a474b8b295a168831c513f77966ff6ba0d242692/src/node_kernels/TKESSTIDDESNodeKernel.C#L123

src/node_kernels/TKESSTIDDESNodeKernel.C
psakievich: It's an epsilon to prevent a divide by zero in the case that `omegaSq` and `sijSq` are both zero.

sayerhs: @psakievich you should have the `epsilon` factor outside of `sqrt` now you are effectively adding `sqrt(1e-10)` which is `1e-5` probably not what you want

sayerhs: Sorry ignore my previous comment... parsed the parenthesis wrong.

src/edge_kernels/MomentumEdgeSolverAlg.C
psakievich: :+1: will refactor

psakievich:
Just checked and I'm getting weird values for the `iddes_rans_indicator`. Need to find where I introduced the problem

src/edge_kernels/MomentumEdgePecletAlg.C
```diff
@@ -103,5 +105,36 @@ MomentumEdgePecletAlg::execute()
...
+  stk::mesh::field_fill(0.0, *maxPecFac);
+
+  const stk::mesh::Selector sel =
+    stk::mesh::selectField(*pecletFactor) & meta.locally_owned_part();
+
+  for (const auto* ib : bulk.get_buckets(stk::topology::EDGE_RANK, sel)) {
+    const auto& b = *ib;
+    const size_t length = b.size();
+    for (size_t k = 0; k < length; ++k) {
+      stk::mesh::Entity edge = b[k];
```
sayerhs, line 127:
```suggestion
  for (const auto* ib : bulk.get_buckets(stk::topology::EDGE_RANK, sel)) {
    for (const auto edge: *ib) {
```

PR #783, opened 2020-11-26
Add FLLC to NGP actuator models

psakievich:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

This pull request is to extend the filtered lifting line correction (FLLC) capabilities to the newer OpenFAST based actuator models that use Kokkos data structures. Sorry this is kind of a big PR.  I was working on this while I waited for the first round of FLLC changes to merge and so it is a bigger chunk of code than is ideal.

This involves changing the FLLC looping to not loop over an entire actuator entity (this was a blade for the ActLineSimpleNGP case but becomes an entire turbine for the OpenFAST variants), but rather a subsection of the actuator entity. 
This also allows for greater parallelization of this computation because blades can be split across more ranks i.e. if there are more ranks than actuator models the individual blades can be distributed to the larger number of ranks.
In terms of the OpenFAST based models this is really the only change that needed to be made since the FLLC equations remain the same and they are all designed to work with the top of the inheritance hierarchy (`ActuatorBulk`).

The additional feature of this pull request that changess a lot of code is to add higher level encapsulation of the actuator model suite to remove some nasty logic from `Realm.C` and streamline the interface.  It's not all gone yet, but as soon as we're comfortable removing the legacy models that all can be deleted.

There is also a change to reduce the really computational taxing part of this by reducing the correlation of every point on a blade with every other point to just correlating with `N` neighbors.  @tonyinme and I did a little testing with it and it didn't really make much of a difference.  Seems like the OpenFAST computation is the bottle neck of the actuator analysis. Especially with the finer resolution along the blades.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [x] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [x] Intel compilers
    - [x] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*


tonyinme (commented):
This looks good.
Just one question, where is "nNeighbors" defined and what is the default value?

src/actuator/ActuatorBladeDistributor.C
```diff
@@ -0,0 +1,120 @@
...
+  case (ActuatorType::ActLineSimpleNGP): {
+    auto actMetaSimp = dynamic_cast<const ActuatorMetaSimple&>(actMeta);
+    // one blade per processor for this case, but we could change this
+    if (!actMeta.entityFLLC_(actBulk.localTurbineId_))
+      break;
+    if (rank == actBulk.localTurbineId_) {
+      const int iBlade = actBulk.localTurbineId_;
+      const int offset = actBulk.turbIdOffset_.h_view(iBlade);
+      const int nPoints = actMetaSimp.num_force_pts_blade_.h_view(iBlade);
+      const int nNeighbor = actMetaSimp.numNearestPointsFllcInt_.h_view(iBlade);
```
psakievich, line 49: @tonyinme it's defined here. This is zero unless users specify it in the input deck. So the default is to integrate the full blade.

tonyinme (approved):
I tested the correction and it works well on an nrel 5mw case.

PR #787, opened 2020-12-10
Ngp transition of mesh motion

ashesh2512:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [X] Feature enhancement

This pull request transitions the mesh motion classes to run on GPUs. The design in this pull request follows the structure of node kernels such that every motion subclass forms an individual kernel. Consequently, all standard library calls have been removed from the motion subclasses, and data structures of `TransMatType` and `ThreeDVecType` have been refactored to support this transition.

The relevant unit tests have been converted to run on GPUs too. One of the redundant unit test cases of `UnitTestCompositeMotions` has been removed. Additionally, following the transition of `FrameBase::compute_centroid_on_parts` to GPUs, a new unit test has been introduced to test computation of centroid on GPUs. 

All NGP regression tests seem to run fine with the moving NGP overset test cases showing absolute differences in norms of about 1e-13~1e-14 on Eagle.  

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [X] Linux
    - [X] MacOS
  - Compilers 
    - [X] GCC
    - [X] LLVM/Clang
    - [ ] Intel compilers
    - [X] NVIDIA CUDA
- [X] Passes all unit tests

*For new code updates*
- [X] New unit tests providing coverage for new code, bug fixes etc.


src/mesh_motion/MeshMotionAlg.C
```diff
@@ -61,8 +61,8 @@ void MeshMotionAlg::initialize( const double time )
 
     for (const auto& ff: fnames) {
       auto* fld = meta.get_field(stk::topology::NODE_RANK, ff);
-      fld->modify_on_host();
-      fld->sync_to_device();
+      fld->modify_on_device();
+      fld->sync_to_host();
     }
```
alanw0: Where was the field modified on device, and where is it needed on host?
In general you want the modify call as close as possible to where the actual modification happens, so that you only make the modify call if the field data was actually modified.
Similarly, put the sync_to_host call as close as possible to where the field is actually needed on host, so that you only do the sync if/when you actually need to. 
If these calls aren't placed carefully, it can cause un-necessary data copying between host and device, which is expensive.

ashesh2512: @alanw0 Thanks for pointing these out. I did plan on moving these over closer to where they should be but forgot to move them eventually. I have made the appropriate changes in 4a4789251b4a6edbf471d0671158a8cb3068bfdb

PR #790, opened 2020-12-17
Minor updates to IDDES

psakievich:
I combed through the Gritskevitch and Shur papers to compare.  Nothing really major stood out. Couple of minor updates based on Shur 2008.


psakievich:
Forgot to build with Hypre so I haven't checked the regression test yet.  I don't think it will diff based on this PR though.

PR #797, opened 2021-01-19
Add viscous penalty to the edge scheme symmetry bc

psakievich:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

Add the same penalty factor @rcknaus added for #616 to the edge scheme.  Also create an option to control this penalty factor via solution options

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [x] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*


psakievich:
@rcknaus this change is changing the gold matrix in `UnitTestMomentumSymmetryEdge.C`.  What is the scenario we are using to set up the matrix for this? We can either:

1. regenerate the matrix
2. set the penalty factor to 0 for the test
3. set the default penalty factor to 0

I'm in favor of 1 or 2. 2 is definitely the easiest, but 1 could be considered more proper?  thoughts?

include/SolutionOptions.h
```diff
@@ -150,6 +150,7 @@ class SolutionOptions
   double earthAngularVelocity_;
   double latitude_;
   double raBoussinesqTimeScale_;
+  double symmetryBcPenatlyFactor_;
```
sayerhs, line 153: Typo in `penalty`

sayerhs:
@psakievich I would prefer the original unit test retain `penaltyFactor_ = 0.0` and then add a new unit test that tests the intended penaltyFactor_.

psakievich:
@sayerhs that is a good suggestion.

psakievich:
Going to run the regressions tests this morning.  I expect some level of diffs to crop up in the edge tests using the symmetry bc similar to the diffs created by #616.

sayerhs:
@psakievich I feel that setting the default value to zero would be least surprising to users who reuse their old input files and reproduce runs. Without any change in the input file they might see slight changes in solutions that could lead to confusion if they are not closely tracking nalu-wind commits or catch this change in default value. 

Is there a significant improvement in solution at the boundary if we switch the default to 2.0 for flat boundaries? That would be an argument to break backwards compatibility. In this situation, we might want to print a warning message if the user hasn't explicitly set penalty factor in the input file alerting them that the default behavior has changed.

As for regression tests, I don't see an issue with re-blessing if they are updated to use the new penalty factor of 2.0.

psakievich:
@sayerhs from the initial studies @rcknaus and I have done the flat boundaries behave like strong symmetry is applied with the penalty factor of 2.0.  So yes there is an improvement.  But the more I think about it, the more I think we should just leave it at 0.0 for the default for now.  We have only done a handful of tests so I can't really speak to full ramifications at this point.  Adding the penalty should also increase the error to some degree in the interior so my vote is to make this a beta capability for now and keep the current default.

sayerhs:
@psakievich Leaving the default to 0.0 is good with me. I suppose you will add or modify a regression test to handle exercising `penaltyFactor  = 2.0` case?

psakievich:
@sayerhs sure I can do that as well.

psakievich:
@sayerhs @rcknaus I've removed the WIP flag and I believe this is ready for merging. I ran a subset of the regression tests and the unit tests and didn't see any failures on my mac.  I believe some diffs will occur in a few element scheme tests since @rcknaus turned on the penalty factor for that scheme by default but they should be small. I wasn't able to detect them in my local testing. When we identify them we can set the penalty factor for them in their input decks or just rebless.

sayerhs (approved):

Issue #817, opened 2021-03-04
CPU build not compiling on summit (Kokkos simd)


PaulMullowney:
Here are the compilation errors ...

[3/645] Building CXX object CMakeFiles/nalu.dir/src/AssembleNodalGradBoundaryAlgorithm.C.o
FAILED: CMakeFiles/nalu.dir/src/AssembleNodalGradBoundaryAlgorithm.C.o
[...]

PR #819, opened 2021-03-09
Fix Summit Power9 build issues

sayerhs:
This PR adds two changes that address issues observed in Summit CPU-only builds:

- Disable STK SIMD when building for CPUs on Summit with GCC v7.5.0. This is a temporary workaround to address #817 
- Change default value for `NALU_ALIGNED` on Summit to remove warnings generated during CPU-only builds.


alanw0 (approved):

alanw0:
FYI @tasmith4 @jvo1012 @ldh4

PR #825, opened 2021-03-14
Fix CUDA warnings in UnitTestCreateOnDevice

sayerhs:
CDash: https://my.cdash.org/viewBuildError.php?type=1&buildid=1981182

### DIFF
--- a/unit_tests/UnitTestCreateOnDevice.C
+++ b/unit_tests/UnitTestCreateOnDevice.C
@@ -10,26 +10,10 @@
 #include <KokkosInterface.h>
 #include "utils/CreateDeviceExpression.h"
 
-class Deviceable {
-protected :
-  Deviceable *deviceCopy_;
+class Shape
+{
 public :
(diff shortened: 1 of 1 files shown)


unit_tests/UnitTestCreateOnDevice.C
```diff
@@ -10,39 +10,23 @@
 #include <KokkosInterface.h>
 #include "utils/CreateDeviceExpression.h"
```
overfelt, line 11: Is this header file still needed since the test has changed to not keep a class device pointer.

overfelt (approved):

unit_tests/UnitTestCreateOnDevice.C
sayerhs: @overfelt The class device pointer was never used in the test. Instead a new device pointer was created in the test function. This section requires that header file

https://github.com/Exawind/nalu-wind/blob/b70bdfee832da7ce82611be19f2d4ccbace11f45/unit_tests/UnitTestCreateOnDevice.C#L69-L76

Issue #829, opened 2021-03-21
Remove unused element kernels from codebase


sayerhs:
CVFEM kernels and supplemental algorithms are no longer used in wind simulations. This issue will track the progress of identifying and removing unused element kernels from the codebase. 

- [x] Tag current working version of Nalu-Wind before removing code
- [x] Remove deprecated SuppAlg code
- [x] Change default in Realm to `use_edges: yes` and warn users for deprecation
- [ ] Identify key functionality that is currently being covered by element schemes that must have an edge equivalent test, if any
- [x] Remove `SuppAlg` and `SuppAlgDep` kernels and their corresponding unit tests
- [x] Remove element kernels and their corresponding unit tests

sayerhs:
Tagged current version of the codebase: [v1.3.0](https://github.com/Exawind/nalu-wind/releases/tag/v1.3.0)

Corresponding Trilinos commit: trilinos/Trilinos@32db0da
CDash report: https://my.cdash.org/index.php?project=Nalu-Wind&date=2021-03-21

sayerhs:
# List of algorithms/kernels to be removed 

## Supplemental element algorithms 
- [ ] `ContinuityMassElemSuppAlgDep`
- [ ] `MomentumBuoyancySrcElemSuppAlgDep`
- [ ] `MomentumMassElemSuppAlgDep`
- [ ] `ScalarMassElemSuppAlgDep`
- [ ] `TurbKineticEnergyKsgsBuoyantElemSuppAlg`
- [ ] `nso/MomentumNSOElemSuppAlgDep`
- [ ] `nso/MomentumNSOGradElemSuppAlg`
- [ ] `nso/MomentumNSOKeElemSuppAlgDep`
- [ ] `nso/ScalarNSOElemSuppAlgDep`
- [ ] `nso/ScalarNSOKeElemSuppAlg`
- [ ] `user_functions/SteadyTaylorVortexContinuitySrcElemSuppAlg`
- [ ] `user_functions/SteadyTaylorVortexMomentumSrcElemSuppAlg`
- [ ] `user_functions/SteadyThermal3dContactSrcElemSuppAlgDep`
- [ ] `user_functions/SteadyThermalContactSrcElemSuppAlg`
- [ ] `user_functions/VariableDensityContinuitySrcElemSuppAlg`
- [ ] `user_functions/VariableDensityMixFracSrcElemSuppAlg`
[...]

PR #832, opened 2021-03-22
Remove element-scheme support for incompressible solver

sayerhs:
See #829 for deprecation details. 

Additional changes:

- Switch default scheme to `edges` in Realm. Throw error if the scheme is not `edges` or `matrix-free`
- Convert all overset and OpenFAST actuator line tests to `use_edges: true`
- Switch `ActLineSimple` to edge-based scheme (superseded by `ActLineSimpleNGP` anyway)
- Remove regression tests that used element-based schemes (restore if needed on a case-by-case basis)

Other details/issues:

- Currently, checks are performed within each `EquationSystem`. Clean-up of `SolutionOptions` still pending.
- Some element kernels and legacy, non-NGP element algorithms are still retained because they are used by `HCEQS` and `PNGEQS`. Should either convert to NGP versions or remove these EQS to allow for removal of these kernels. 
- DG interface retained with edge-based scheme for sliding mesh


sayerhs:
@lawrenceccheung @tonyinme I added you both so that you can check the changes NREL5MW actuator tests and the `ActLineSimple` tests.

tonyinme (approved):
Hey @sayerhs, this looks good to me. There was no real reason for us not to use edges before. As long as the tests pass, this should all be good.

psakievich:
I ran this on one of the compute servers using GCC 7.2 and saw the following related failures:
```
         25 - ActLineSimple (Failed)
         39 - nrel5MWactuatorLineAnisoGauss (Failed)
         40 - nrel5MWactuatorLineFllc (Failed)
         41 - nrel5MWactuatorDisk (Failed)
         42 - nrel5MWadvActLine (Failed)
         43 - oversetSphereTIOGA (Failed)
         44 - oversetRotCylinder (Failed)
```
These were all elem tests that @sayerhs noted were converted to the edge scheme so no surprises to see the diffs.

reg_tests/CTestList.cmake
```diff
@@ -121,63 +121,22 @@ if(NOT ENABLE_CUDA)
...
   add_test_r(edgeHybridFluids 8)
   add_test_r(ekmanSpiral 4)
-  add_test_r(ekmanSpiralConsolidated 4)
-  add_test_r_inp(elemBackStepLRSST 4)
-  add_test_r(elemClosedDomain 2)
-  add_test_r(elemHybridFluids 8)
-  add_test_r(elemHybridFluidsShift 8)
-  add_test_r(femHC 2)
-  add_test_r(femHCGL 2)
-  add_test_r(heatedBackStep 4)
```
psakievich, line 153: I'm wondering if we should convert a few more of these tests to be edge scheme. Browsing through, I'm thinking about:

- `closedDomain`
- at least one of the `backStep`'s
- `milestoneRun`
- `variableDensNonIso` (this might be better to swap the with the NonUniform version of the test which uses mixture fraction. IIRC we're going to remove mixture fraction at some point as well)

reg_tests/CTestList.cmake
```diff
@@ -242,8 +199,6 @@ if(NOT ENABLE_CUDA)
   #=============================================================================
   # Performance tests
   #=============================================================================
-  add_test_p(uqSlidingMeshDG 8)
```
psakievich, line 245: We can also remove the performance test macros from the top of this file unless we want to replace them with more targeted wind energy edge variants. I'm open to either option.

psakievich (changes requested):
@sayerhs thanks for doing this. Looks really good.  I noticed the `ductWedge` test is an edge test that I don't think should be removed. That's the only change I'm requesting.  I think it is worth discussing converting a few more tests from elem to edge to make sure we have adequate coverage.  But I don't think those conversions should hold up this PR.  We can certainly address them in a follow on PR. Looking forward to trimming the mesh repo once this goes through :)

reg_tests/CTestList.cmake
sayerhs: @psakievich I was hoping to add wind-energy specific performance tests here. The empty section header was meant as reminder for whoever that ends up adding these tests later.

reg_tests/CTestList.cmake
sayerhs: When we reinstate these tests, we should rename tests like `milestoneRun` with something more relatable

psakievich: Agreed

reg_tests/CTestList.cmake
psakievich: Sounds good to me

marchdf (approved):
three thumbs-up!

psakievich (approved):
After speaking with @sayerhs we agreed there isn't anything special in the `ductWedge` case to make it a critical case to preserve in the regression test suite and hold back this PR.

PR #838, opened 2021-03-27
add convecting taylor vortex regression test using edges

ashesh2512:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

Add regression test for convecting taylor vortex using edges.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [x] MacOS
  - Compilers 
    - [x] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [x] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*


sayerhs:
Thanks @ashesh2512. It looks good overall. I had a couple of questions:

- Since you're outputting solution norms, can implement a check based on solution norms instead of `mean system norms`? That would be preferable because we can set a tighter tolerance for all compilers.
- Would it be possible to run this test on GPUs also? If we do that we should tighten the linear solver tolerances.

ashesh2512:
@sayerhs I have pushed an additional commit to modify the test to check against solution norms - `845196205b4808870f7938e55014b030bd11fc9d`

src/SolutionNormPostProcessing.C
```diff
@@ -339,6 +339,12 @@ SolutionNormPostProcessing::execute()
     int offSet = 0;
     for ( size_t j = 0; j < fieldPairVec_.size(); ++j ) {
 
+      // sync fields
+      fieldPairVec_[j].first->modify_on_device();
```
sayerhs, line 343: @ashesh2512 These should be marked as modified where they were updated. Only need `sync_to_host` here.

src/SolutionNormPostProcessing.C
```diff
@@ -339,6 +339,12 @@ SolutionNormPostProcessing::execute()
...
+      fieldPairVec_[j].first->sync_to_host();
+      fieldPairVec_[j].second->modify_on_device();
+      fieldPairVec_[j].second->sync_to_host();
```
sayerhs, line 346: Exact DOF field is `user_functions` and don't have NGP implementations. I suppose we are computing exact fields again somewhere after this?

src/SolutionNormPostProcessing.C
ashesh2512: @sayerhs Thanks for pointing that out. I was trying to be safe, but I have pushed a commit removing the redundant calls - https://github.com/Exawind/nalu-wind/pull/838/commits/4911b48139317aaee25402b59baa3fbb4a35a4e6

src/SolutionNormPostProcessing.C
ashesh2512: @sayerhs Thanks for catching this. You are right, the exact fields should not be modified anywhere on device. I have removed these calls in https://github.com/Exawind/nalu-wind/commit/4911b48139317aaee25402b59baa3fbb4a35a4e6

sayerhs (approved):

PR #842, opened 2021-03-29
Fix bug in ActLineSimple introduced by #832

sayerhs:
The change from elem to edge schemes missed adding the node-based lumped actuator source terms to the momentum equation.


jrood-nrel (approved):

psakievich (approved):
This works too :).  But this test will be removed in another week or two.

PR #847, opened 2021-04-14
Simpler HYPRE IJMatrix Assemble:

PaulMullowney:
Regressions test would pass on Rhodes except for the changes in reg_tests/hypre_settings/hypre_blade_resolved.yaml. I've changed these to reflect new smoothers/preconditioners and better parameters to use for BoomerAMG in the ABL tests. 

Now to the actual changes.

In this branch, the data structures for the assembled matrix (i,j,values) are combined into one large
monolithic buffer for owned and shared. The owned part comes first, then the shared. In addition,
the number of contributions (nnz, including repeats) on rank i from all other ranks is computed. This
allows us to resize the monolithic buffers to be nnz_owned + max(nnz_shared, nnz_shared_recv). Then,
this buffer can be reused inside IJMatrixAssemble thus saving a substantial amount of memory. Since this
optimized hypre assembly algorithnm is not yet in Hypre master, the API methods that indicate to
Hypre to use this new algorithm are disabled (though the code is left in for future use). The current
version of this new Nalu data structure works with Hypre master (without the disabled code!). As a part of this
development, the row_indices/counts has been replaced by rows which is the same size of as cols/values.
This allows us to use simpler IJMatrixSet/AddToValues2 APIs which use less memory and have fewer kernel
calls. We have added the option "simple_hypre_matrix_assemble", which will us to toggle between the
optimized and more general assembly algorithms, and will be useful once the optimized version is in
Hypre master.

Next, for all calls to sumInto* (HypreLinearSystem and HypreUVWLinearSystem), we have replaced the
binarySearch method for finding matrix write location with a sort and a constrained linear search. Because
the matrix locations defined by the column are sorted, one can use the old linear search as the starting
point for the next matrix write location, defined by the next column, which is sorted in ascending order.
These optimizations are facilitated by moving all read-only data structures in the numeric assembly to
RandomAccess Kokkos views. That is, we attempt to use fast Texture memory where possible.

We have added the ability to dump matrix statistics (rows, nnz per rank). This is useful for measuring
the effect of load balancing strategies and partitioning algorithms like parmetis. This is turned off
by default. Moreover, if toggled, we only dump matrix stats once per equation system construction after
the global matrix assembly. The option is "dump_hypre_matrix_stats".

**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [x] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [x] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*

### DIFF
--- a/include/HypreLinearSystem.h
+++ b/include/HypreLinearSystem.h
@@ -20,6 +20,9 @@
 #endif // HYPRE_LINEAR_SYSTEM_DEBUG
 #undef HYPRE_LINEAR_SYSTEM_DEBUG
 
+#include <iostream>
+#include <fstream>
+
 #include "LinearSystem.h"
 #include "HypreDirectSolver.h"
 
@@ -49,6 +52,10 @@ using DoubleViewHost = DoubleView::HostMirror;
 using HypreIntTypeView = Kokkos::View<HypreIntType*, sierra::nalu::MemSpace>;
 using HypreIntTypeViewHost = HypreIntTypeView::HostMirror;
 
+// const random access views for read only, noncoalesced (texture) memory fetches
+using UnsignedViewRA = Kokkos::View<const unsigned*, sierra::nalu::MemSpace, Kokkos::MemoryTraits<Kokkos::RandomAccess> >;
+using HypreIntTypeViewRA = Kokkos::View<const HypreIntType*, sierra::nalu::MemSpace, Kokkos::MemoryTraits<Kokkos::RandomAccess> >;
+
 // This 2D view needs to be LayoutLeft. Do NOT change
 using DoubleView2D =
   Kokkos::View<double**, Kokkos::LayoutLeft, sierra::nalu::MemSpace>;
@@ -76,6 +83,8 @@ using DoubleViewUVM = Kokkos::View<double*, sierra::nalu::UVMSpace>;
 using DoubleView2DUVM =
   Kokkos::View<double**, Kokkos::LayoutLeft, sierra::nalu::UVMSpace>;
 using HypreIntTypeViewUVM = Kokkos::View<HypreIntType*, sierra::nalu::UVMSpace>;
+using HypreIntTypeView2DUVM =
+  Kokkos::View<HypreIntType**, Kokkos::LayoutLeft, sierra::nalu::UVMSpace>;
 
 // Periodic Node Map
 using PeriodicNodeMap =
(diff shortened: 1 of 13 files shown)


include/HypreLinearSystem.h
```diff
@@ -20,6 +20,9 @@
 #endif // HYPRE_LINEAR_SYSTEM_DEBUG
 #undef HYPRE_LINEAR_SYSTEM_DEBUG
 
+#include <iostream>
+#include <fstream>
```
sayerhs, line 24: Do we need these headers here? Move them to `C++`?

include/LinearSolverConfig.h
```diff
@@ -59,6 +59,12 @@ class LinearSolverConfig
   inline bool useSegregatedSolver() const
   { return useSegregatedSolver_; }
 
+  inline bool simpleHypreMatrixAssemble() const
+  { return simpleHypreMatrixAssemble_; }
+
+  inline bool dumpHypreMatrixStats() const
+  { return dumpHypreMatrixStats_; }
+
```
sayerhs, line 67: Seems like inheritance is breaking down here. Exposing subclass logic in base class?

src/HypreLinearSystem.C
```diff
@@ -1463,18 +1548,124 @@ HypreLinearSystem::hypreIJVectorSetAddToValues()
   auto num_rows_owned = hcApplier->num_rows_owned_;
   auto num_rows_shared = hcApplier->num_rows_shared_;
 
+  HypreDirectSolver* solver = reinterpret_cast<HypreDirectSolver*>(linearSolver_);
+  if (solver->getConfig()->simpleHypreMatrixAssemble()) {
+#if 0
```
sayerhs, line 1553: Is this the code that requires updates to hypre to work?

sayerhs (approved):

include/LinearSolverConfig.h
PaulMullowney: You're right. I should try to fix this.

include/HypreLinearSystem.h
PaulMullowney: I was using these for some IO defined by dumpHypreMatrixStats. What do you suggest?

src/HypreLinearSystem.C
PaulMullowney: Code wrapped in "#if 0" is code in a branch of Hypre that I have not get merged to Hypre master. I'm keeping it in there to prepare for the time when I do get that into master.

In this PR, simple_hypre_matrix_assemble does nothing. We just use the default Hypre assembly. Once, the Hypre branch is merged, it will allow me to either toggle between the default assembly and the new, fast assembly. 

I'm trying to keep my code from diverging from master too much and not have to rebranch every so often. My intent is definitely to clean this up in the long run while reducing my short term work.

include/HypreLinearSystem.h
sayerhs: Can you show me the specific code in this file? I am assuming the functions/classes from these headers are used in the C++ file?

PaulMullowney: The function that uses this HypreLinearSystem.C::dumpRowNNZStats(). Line 1577

PaulMullowney: https://github.com/PaulMullowney/nalu-wind/blob/hypre_rows_counts/src/HypreLinearSystem.C#L1577

sayerhs: Thanks. Any reason then that these headers are not in `HypreLinearSystem.C`

PaulMullowney:
None. I’ll fix.

From: Shreyas Ananthan ***@***.***>
Reply-To: Exawind/nalu-wind ***@***.***>
Date: Wednesday, April 14, 2021 at 1:03 PM
To: Exawind/nalu-wind ***@***.***>
Cc: "Mullowney, Paul" ***@***.***>, Author ***@***.***>
Subject: Re: [Exawind/nalu-wind] Simpler HYPRE IJMatrix Assemble: (#847)


CAUTION: This email originated from outside of NREL. Do not click links or open attachments unless you recognize the sender and know the content is safe.

@sayerhs commented on this pull request.

________________________________

In include/HypreLinearSystem.h<https://gcc02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FExawind%2Fnalu-wind%2Fpull%2F847%23discussion_r613514785&data=04%7C01%7CPaul.Mullowney%40nrel.gov%7C4c6ccc5e4a27491792bb08d8ff77e486%7Ca0f29d7e28cd4f5484427885aee7c080%7C0%7C0%7C637540237909450495%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=7956%2BIiX8%2BjeM7rQTL%2B0lxviPGSNNR2HzTy93flNgKs%3D&reserved=0>:

> +#include <iostream>

+#include <fstream>

Thanks. Any reason then that these headers are not in HypreLinearSystem.C

include/LinearSolverConfig.h
PaulMullowney: Fixed in the most recent commit.

include/HypreLinearSystem.h
PaulMullowney: Fixed in the most recent commit.

PR #848, opened 2021-04-19
Improve FAST timestep ratio tolerance

psakievich:
**Pull-request type:**
- [x] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

Analyst have reported an issue with the computation of the timestep ratio for FAST and nalu-wind.  I've updated the unit test to use their case and tightened the checking tolerance.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [x] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/actuator/ActuatorBulkFAST.C
+++ b/src/actuator/ActuatorBulkFAST.C
@@ -64,10 +64,9 @@ ActuatorBulkFAST::~ActuatorBulkFAST() { openFast_.end(); }
 
 bool
 ActuatorBulkFAST::is_tstep_ratio_admissable(
-  const ActuatorMetaFAST& actMeta, const double naluTimeStep)
+  const double fastTimeStep, const double naluTimeStep)
 {
-  const double stepCheck =
-    std::abs(naluTimeStep / actMeta.fastInputs_.dtFAST - tStepRatio_);
+  const double stepCheck = std::abs(naluTimeStep / fastTimeStep - tStepRatio_);
   return stepCheck < 1e-12;
 }
 
@@ -76,7 +75,7 @@ ActuatorBulkFAST::init_openfast(
   const ActuatorMetaFAST& actMeta, const double naluTimeStep)
 {
   openFast_.setInputs(actMeta.fastInputs_);
-  if (!is_tstep_ratio_admissable(actMeta, naluTimeStep)) {
+  if (!is_tstep_ratio_admissable(actMeta.fastInputs_.dtFAST, naluTimeStep)) {
     throw std::runtime_error("ActuatorFAST: Ratio of Nalu's time step is not "
                              "an integral multiple of FAST time step.");
   } else {
@@ -109,12 +108,15 @@ ActuatorBulkFAST::init_openfast(
   } else {
     squash_fast_output(std::bind(&fast::OpenFAST::init, &openFast_));
   }
-  if (!is_tstep_ratio_admissable(actMeta, naluTimeStep)) {
+  /* TODO update/uncomment this check once openfast adds in a way 
+  to get the actual time step from fast::OpenFAST
+  if (!is_tstep_ratio_admissable(openFast_.dtFAST, naluTimeStep)) {
     throw std::runtime_error("OpenFAST is using a different time step than "
                              "what was specified in the input deck. "
                              "Please check that your workflow is consistent "
                              "(restarts, FAST files, etc.");
   }
+  */
 
   for (int i = 0; i < nTurb; ++i) {
     if (localTurbineId_ == openFast_.get_procNo(i)) {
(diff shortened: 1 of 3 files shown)


src/actuator/ActuatorBulkFAST.C
```diff
@@ -64,20 +64,19 @@ ActuatorBulkFAST::~ActuatorBulkFAST() { openFast_.end(); }
...
-    // arbitrary number
-    // 0.001
+  const double stepCheck =
+    std::abs(naluTimeStep / actMeta.fastInputs_.dtFAST - tStepRatio_);
+  if (stepCheck > 1e-12) {
+    throw std::runtime_error(
+      "ActuatorFAST: Ratio of Nalu's time step is not "
+      "an integral multiple of FAST time step. Round off error is " +
+      std::to_string(stepCheck));
+  } else {
```
sayerhs, line 77: This check is performed too early in the process and only checks if `dt` in the Nalu-Wind YAML input file is an integral multiple of the `dt_fast` specified in the same YAML file. As far as OpenFAST is concerned, it doesn't use/respect this variable. `fast::OpenFAST::init` returns the `dtFAST` that it uses OpenFAST based on FAST input files/restart (see [FAST_OpFM_Init definition](https://github.com/OpenFAST/openfast/blob/1bb0fbc622a374a68cc805cb7d0edcdedb443243/modules/openfast-library/src/FAST_Library.f90#L455-L475)). Thus, the `dtFAST` used by OpenFAST Fortran routines (and stored internally in `fast::OpenFAST` might still be different despite the check performed here. A proper fix would require recomputation of `tStepRatio_` after `dtFAST` is set by OpenFAST C++ interface (i.e., after a call to `fast::OpenFAST::init`).

sayerhs: @psakievich Couple of points:

- 142ab26 still doesn't do the right thing as C++ API doesn't return the `dtFAST` used by OpenFAST back to Nalu-Wind. You'll need to query the `dtFAST` variable from `fast::OpenFAST` class to verify if it has changed.
- There is a misunderstanding of the OpenFAST's OpenFOAM interface. As you can see from the code snippet I shared. `dtFAST` is a variable with `intent(out)`, i.e., it was never the case that CFD code sets the timestep size through the API. CFD code is the _consumer_ of the timestep size returned by OpenFAST after initialization. Basically, `dt_fast` should not be an input in Nalu-Wind side at all, instead it should be whatever is set in OpenFAST input files.

psakievich: Okay I will fix that. On bullet point 2, then what is the point of having this variable in the C++ API?  It sounds like the timestep in the fast inputs is essentially not used for anything?  The analyst issue was just a round off error when type casting to int during initialization. I've verified that it is taken care of. What you're describing sounds like a potential new class of bugs that we have yet to experience.

sayerhs: As I mentioned in my comment `dt_fast` should *never* have been an input on Nalu-Wind side. Instead `fast::OpenFAST` should have had an accessor for `dtFAST` that could have been verified to: 1. either ensure that OpenFAST and Nalu-Wind timestep sizes are integral multiples, or 2. adjust Nalu-Wind timestep to the nearest integral multiple of OpenFAST timestep. 

As for the _analyst issue_, the original error check and the new error check really doesn't protect the user from mistakes if there was a different timestep in OpenFAST input file.

src/actuator/ActuatorBulkFAST.C
```diff
@@ -62,18 +62,23 @@ ActuatorBulkFAST::ActuatorBulkFAST(
 
 ActuatorBulkFAST::~ActuatorBulkFAST() { openFast_.end(); }
 
+bool
+ActuatorBulkFAST::is_tstep_ratio_admissable(
+  const ActuatorMetaFAST& actMeta, const double naluTimeStep)
+{
+  const double stepCheck =
+    std::abs(naluTimeStep / actMeta.fastInputs_.dtFAST - tStepRatio_);
```
sayerhs, line 70: `actMeta.fastInputs_.dtFAST` is out of sync with `fast::OpenFAST::dtFAST`

psakievich: @sayerhs I'm not seeing a way to get `dtFAST` from `fast::OpenFAST`.  Looks like it's a private member with no getter function.  I'm going to just comment this new check out and open an issue with OpenFAST to add that feature in.

lawrenceccheung (approved):
Thanks Phil for taking a look at this.  There's actually a number of consistency checks we need with OpenFAST, e.g., not only checking the timestep, but also the density and a couple of other inputs.  We'll probably need to find a good, general way to handle them so analysts don't run into them in the future.

src/actuator/ActuatorBulkFAST.C
psakievich: I understand what you are saying.  In this particular case the result of the integer division was such that not doing this check would have lead to an error even if we had used the correct openfast value. I will open an issue with the openfast team to add a data accessor and once that is in place we can add the extra check back and/or move the computation of the timestep ratio at that time.  This is good to know. Thanks @sayerhs

PR #851, opened 2021-04-27
Update golds in response to Kokkos update in Trilinos

jrood-nrel:
https://github.com/trilinos/Trilinos/pull/9054 appears to have caused diffs https://my.cdash.org/viewTest.php?onlyfailed&buildid=2003614

After investigating, I think it's fine to update these.


PaulMullowney (commented):
Most of the diffs are small however ablHill3d has large diffs. Not sure what to make of this.

psakievich:
@PaulMullowney I have the same thoughts.  I think the abl hill warrants a closer investigation before reblessing.  I'll try to take a closer look today.

jrood-nrel:
Looking at the gold file, it looks reasonable to me, but I will wait.

psakievich (approved):
I took a closer look at the ablHill norms and I'm fine reblessing them as well.  For some reason I was thinking it was a E-2 absolute diff from when I checked yesterday instead of E-9. I think this is still in the realm of order of operation diffs.  Small diffs are likely amplified by FFT's in the ABL potential flow model used for that test.

PR #855, opened 2021-04-29
Update golds in response to fixes to Kokkos in Trilinos

jrood-nrel:
https://github.com/trilinos/Trilinos/pull/9074


psakievich:
I'm seeing a real diff on the actuator tests.  Not really sure yet what's causing it.

PaulMullowney (commented):
The diffs for Hypre and SST tests seem reasonable. The actuator line tests have significant diffs as Phil noted.

sayerhs:
@psakievich the decomposition changed with the STK generated mesh (it is now bread slices in z direction) that's going to be different from `rcb`). You might want to call balance to redistribute into RCB and see if that improves matters. Either way, I'd suspect a larger change in mean system norms based on different partitioning.

psakievich:
@sayerhs ah I wasn't thinking the partitioning should have had that strong of an effect. I'll try stk rebalance and see if that improves things.  Visual inspection in paraview shows the results still look good.

sayerhs:
@psakievich to be absolutely sure run the test on 1 process with both the old and new mesh. That should match exactly between the two cases.

psakievich (approved):
I've run with `rebalance: yes` and `stk_rebalance_method: rcb` and it shifted the norms again.  This definitely supports @sayerhs observation about how decomposition effects on the global norm.  Given that nothing has changed in the actual actuator code and I can't see a diff via visual inspection with ParaView I'm okay with reblessing them.

FWIW I also tried to do a point to point diff with  exodiff and ParaView, but because the block name changed exodiff chokes on that. And ParaView with a programmable filter is also giving me garbage trying to diff them.  I'm guessing its because the node id's to spatial locations changed 🤷

sayerhs:
@psakievich what option did you use in paraview to transfer field from one mesh to another?

psakievich:
@sayerhs I created a programmable filter using this python code:
```
A = inputs[0]
B = inputs[1]

for k in A.PointData.keys():
  print(k)
  output.PointData.append(A.PointData[k]-B.PointData[k], k)
```
The diffs it is giving me are O(1) compared to the DOF's and I can confirm through visual inspection that diffs aren't that large between the filter's source files.

sayerhs:
@psakievich OK. That might not work because the node IDs might be off. Pointwise meshes usually number the corners first and then the interior nodes. Have you tried `Resample With Dataset` filter in Paraview. I remember that Ganesh has used that successfully when comparing pure nalu-wind simulations with nalu/amr-wind hybrid simulations.

psakievich:
@sayerhs you are correct. Resample with Image/Dataset drives the diffs down to the linear solver tolerance for all fields.

PR #853, opened 2021-04-27
Move all actuator tests to IOSS generated meshes

psakievich:
Use IOSS mesh generation for actuator regression tests that just use box meshes

Regression tests pass locally

PR #862, opened 2021-05-11
Update gold files for nrel5MW tests

jrood-nrel:
I am still trying to find time to verify these. They showed up around #853 and some changes in Trilinos simultaneously. I'm making a PR so it's easy to update them.


jrood-nrel:
I think they're fine but I need to verify them against the original meshes in the submodule.

psakievich (approved):
I think they are fine as well.  I verified that the decomp changes from using the IOSS generated mesh were around `1E-7` and this lines up with a previous study done by @rcknaus.  So the only other thing that changed was the flip flopping around with Trilinos right at the same time.  I did a manual inspection of some of these exodus files and couldn't find anything noteworthy in the diff.

jrood-nrel:
Ok with that. I think I will merge this.

PR #864, opened 2021-05-14
Update gold files in response to transition from SuperLU to KLU

jrood-nrel:


jrood-nrel:
These should be fine. Just tagging you both to be aware that the tests will be running with SuperLU now.

PaulMullowney (approved):

PR #865, opened 2021-05-22
Update OpenFAST files to work with OpenFAST v2.6.0

psakievich:
I have not tested this with nalu-wind yet, but I did test it with AMR-Wind yesterday using the OpenFAST develop branch.  I'm hoping we can use this as a patch to get the nightly tests passing (we'll have to update OpenFAST builds) since these changes won't work with the main branch.  

In general we need to come up with a more robust strategy for testing against openfast.  I was thinking a common repo that can be shared by amr-wind and nalu-wind that will also work with openfast master and develop.  It would be good to test against both like we do with Trilinos since this is a more volatile library like Trilinos.  Shreyas suggested using OpenFAST's regression test suite which could work well provided we get the paths and controller build to work.


psakievich:
Updating this to work with the main branch instead

psakievich:
tested locally and merging to get nightly tests working again

PR #867, opened 2021-06-02
AMS updates

marchdf:
**Pull-request type:**
- [X] Bug fix 
- [ ] Documentation update
- [X] Feature enhancement

This provides an update to AMS. The following changes were made:

- Updated Peclet calculation for AMS (uses the turbulent viscosity as opposed to the previous version which didnt)
- Fix inconsistency with the calculation of CM43scale (it used to be a slightly different calculation in two different parts of the code)
- Added user input for CMdeg
- Added the possibility of outputting the peclet number
- Added an AMS `post-iter` work to correctly output AMS forcing components
- Added an AMS `predict_state` to correctly handle AMS average quantities

This should introduce no diffs in regression tests _except_ AMS ones. We expect diffs there.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [X] Linux
    - [ ] MacOS
  - Compilers 
    - [X] GCC
    - [ ] LLVM/Clang
    - [X] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/AMSAlgDriver.C
+++ b/src/AMSAlgDriver.C
@@ -298,18 +298,18 @@ AMSAlgDriver::compute_metric_tensor()
 void
 AMSAlgDriver::post_iter_work()
 {
-  // const auto& fieldMgr = realm_.ngp_field_manager();
-  // const auto& meta = realm_.meta_data();
-  // auto& bulk = realm_.bulk_data();
-  // auto ngpForcingComp =
-  //   fieldMgr.get_field<double>(forcingComp_->mesh_meta_data_ordinal());
-  // ngpForcingComp.sync_to_host();
-  // VectorFieldType* forcingComp = meta.get_field<VectorFieldType>(
-  //   stk::topology::NODE_RANK, "forcing_components");
-  // stk::mesh::copy_owned_to_shared(bulk, {forcingComp});
-  // if (realm_.hasPeriodic_) {
-  //   realm_.periodic_delta_solution_update(forcingComp, 3);
-  // }
+  const auto& fieldMgr = realm_.ngp_field_manager();
+  const auto& meta = realm_.meta_data();
+  auto& bulk = realm_.bulk_data();
+  auto ngpForcingComp =
+    fieldMgr.get_field<double>(forcingComp_->mesh_meta_data_ordinal());
+  ngpForcingComp.sync_to_host();
+  VectorFieldType* forcingComp = meta.get_field<VectorFieldType>(
+    stk::topology::NODE_RANK, "forcing_components");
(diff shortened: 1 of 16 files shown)


jamelvin:
Note that this also:

- Adds a buffer to the division by wallDist in `src/node_kernels/MomentumSSTAMSForcingNodeKernel.C` to prevent `inf` errors
- Corrects an averaging bug in overset cases where the invalid state was contributing to the running average

src/AMSAlgDriver.C
```diff
@@ -17,6 +17,17 @@
 #include "ngp_utils/NgpFieldManager.h"
 #include "ngp_algorithms/MetricTensorElemAlg.h"
 #include "stk_mesh/base/NgpMesh.hpp"
+#include <stk_mesh/base/BulkData.hpp>
+#include <stk_mesh/base/Field.hpp>
+#include <stk_mesh/base/FieldParallel.hpp>
+#include <stk_mesh/base/MetaData.hpp>
+
+//stk_mesh/base/fem
+#include <stk_mesh/base/BulkData.hpp>
```
jamelvin, line 26: Looks like we duplicated these

marchdf: good catch

src/edge_kernels/AMSMomentumEdgePecletAlg.C
```diff
@@ -0,0 +1,113 @@
...
+      for (int d = 0; d < ndim; ++d) {
+        const DblType dxj =
+          coordinates.get(nodeR, d) - coordinates.get(nodeL, d);
+        asq += av[d] * av[d];
+        axdx += av[d] * dxj;
+        udotx += 0.5 * dxj * (vrtm.get(nodeR, d) + vrtm.get(nodeL, d));
+      }
+
+      const DblType pecnum =
+        rhoIp * stk::math::abs(udotx) / (muIp + pecScale_ * mutIp + eps);
```
psakievich, line 106: What is the purpose of the scaling factor here?

psakievich (approved):
Looks good to me.

src/edge_kernels/AMSMomentumEdgePecletAlg.C
jamelvin: It's pretty adhoc at the moment, basically to allow us to force central differencing in a larger part of the flow and we needed to be able to control it as the turbine case was showing some sensitivity to it.

We are looking into a more theoretically sound formulation to replace this in the future.

psakievich: :+1:

gantech (approved):

Issue #859, opened 2021-05-06
Extract surface data during run


gantech:
Its would be great to be able to extract the surface data and geometry on specified boundaries during a run at much higher frequencies compared to the full field output. This would help calculation of load profiles on blades at much higher frequencies. The geometry information could be optimized for rigid body rotation runs, but would be necessary at the same frequency for FSI runs.

psakievich:
@gantech thanks for putting this on the radar.  I don't think this will be too big of a lift.  Would you be okay with wrapping this into the Q4 milestone as a subtask or do you need it sooner?

psakievich:
Just updating here that we talked about this today and are going to try using a transfer and input-output region.

PR #870, opened 2021-06-22
Allow for multiple output sections in a realm

psakievich:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

This PR is to resolve the request in #859 to allow for multiple outputs at varying time intervals. It's WIP. I'll update as I go.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/OutputManager.C
+++ b/src/OutputManager.C
@@ -165,17 +165,19 @@ OutputManager::perform_outputs(
 {
   const auto& metaData = ioBroker->meta_data();
   const int nOutputs = infoVec_.size();
+
+  const double elapsedWallTime = stk::wall_time() - wallTimeStart;
+  // find the max over all core
+  double g_elapsedWallTime = 0.0;
+  stk::all_reduce_max(
+    NaluEnv::self().parallel_comm(), &elapsedWallTime, &g_elapsedWallTime, 1);
+  // convert to hours
+  g_elapsedWallTime /= 3600.0;
+
   for (int i = 0; i < nOutputs; i++) {
     auto* outputInfo = &(infoVec_[i]);
     const auto resultsFileIndex = indexVec_[i];
     const int modStep = timeStepCount - outputInfo->outputStart_;
-    const double elapsedWallTime = stk::wall_time() - wallTimeStart;
-    // find the max over all core
-    double g_elapsedWallTime = 0.0;
-    stk::all_reduce_max(
-      NaluEnv::self().parallel_comm(), &elapsedWallTime, &g_elapsedWallTime, 1);
-    // convert to hours
-    g_elapsedWallTime /= 3600.0;
 
     // check for elapsed WALL time threshold
     bool forcedOutput = false;
@@ -184,10 +186,10 @@ OutputManager::perform_outputs(
       if (g_elapsedWallTime > outputInfo->userWallTimeResults_.second) {
         forcedOutput = true;
         outputInfo->userWallTimeResults_.first = false;
-        NaluEnv::self().naluOutputP0()
-          << "Realm::provide_output()::Forced Result output will be processed "
-             "at current time: "
-          << currentTime << std::endl;
+        NaluEnv::self().naluOutputP0() << "Realm::provide_output()::Forced "
+                                          "Result output will be processed "
+                                          "at current time: "
+                                       << currentTime << std::endl;
         NaluEnv::self().naluOutputP0()
           << " Elapsed (max) WALL time: " << g_elapsedWallTime << " (hours)"
           << std::endl;
@@ -200,7 +202,8 @@ OutputManager::perform_outputs(
 
     if (isOutput) { /*
        NaluEnv::self().naluOutputP0()
-         << "Realm shall provide output files at : currentTime/timeStepCount: "
+         << "Realm shall provide output files at : currentTime/timeStepCount:
+       "
          << currentTime << "/" << timeStepCount << " (" << name_ << ")"
          << std::endl;*/
 
--- a/src/OutputManager.C
+++ b/src/OutputManager.C
@@ -104,6 +104,28 @@ OutputManager::create_output_mesh(
     } else {
       indexVec_[i] = ioBroker->create_output_mesh(
         oname, stk::io::WRITE_RESULTS, *outputInfo->outputPropertyManager_);
+      // create selector and limit output if necessary
+      if (!outputInfo->targetNames_.empty()) {
+        // extract part
+        auto& partNameList = outputInfo->targetNames_;
+        stk::mesh::PartVector searchParts;
+        for (size_t k = 0; k < partNameList.size(); ++k) {
+          stk::mesh::Part* thePart = metaData->get_part(partNameList[k]);
+          if (NULL != thePart)
+            searchParts.push_back(thePart);
+          else
+            throw std::runtime_error(
+              "OutputManager::create_output_mesh: Part is null" +
+              partNameList[k]);
+        }
+
+        // selector and bucket loop
+        stk::mesh::Selector s_locally_owned =
+          metaData->locally_owned_part() & stk::mesh::selectUnion(searchParts);
+        // restrict output to the selector and entity rank
+        ioBroker->set_output_selector(
+          indexVec_[i], get_entity_rank(outputInfo, metaData), s_locally_owned);
+      }
     }
 
     // Tell stk_io how to output element block nodal fields:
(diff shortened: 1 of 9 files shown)


src/OutputManager.C
```diff
@@ -0,0 +1,229 @@
...
+{
+  const int nOutputs = infoVec_.size();
+  for (int i = 0; i < nOutputs; i++) {
+    auto* outputInfo = &(infoVec_[i]);
+    const auto resultsFileIndex = indexVec_[i];
+    const int modStep = timeStepCount - outputInfo->outputStart_;
+    const double elapsedWallTime = stk::wall_time() - wallTimeStart;
+    // find the max over all core
+    double g_elapsedWallTime = 0.0;
+    stk::all_reduce_max(
```
rcknaus, line 175: Probably should do this check just once for all output blocks

rcknaus (approved):
Looks good.  The realm changes are a bit hard to pin down since it looks like the file got reformatted.

psakievich:
ah man. I didn't check the realm changes.  Ugh.

src/OutputManager.C
```diff
@@ -0,0 +1,160 @@
...
+  indexVec_.resize(nFound);
+
+  for (int i = 0; i < nFound; i++) {
+    infoVec_[i].load(*foundNodes[i]);
+  }
+}
+
+void
+OutputManager::create_output_mesh(
+  stk::io::StkMeshIoBroker* ioBroker, stk::mesh::MetaData* metaData)
```
tokusanya, line 74: I'm assuming that the meta data passed in is associated with the bulk data on the ioBroker. If that is the case, you can grab the bulk data from the ioBroker and then grab the meta data from the bulk data

unit_tests/UnitTestOutputManager.C
```diff
@@ -0,0 +1,33 @@
...
+
+#include "OutputInfo.h"
+#include "gtest/gtest.h"
+
+#include "NaluParsing.h"
+
+namespace sierra {
+namespace nalu {
+namespace {
+TEST(OutputManagerTest, can_read_multiple_output_nodes)
```
tokusanya, line 18: Can you also add a unit test to write multiple output ?

src/OutputManager.C
psakievich: ah nice! thanks @tokusanya

unit_tests/UnitTestOutputManager.C
psakievich: @tokusanya can you point me to a good example of how to do this? is there a way to call the write operation without actually having to generate the files?

src/OutputManager.C
psakievich: ah yeah thanks.  I'll pull it out of the loop

psakievich:
I tried to remove most of the auto-formatting issues in `Realm.C`.

psakievich:
Looks like my strategy of reusing the `output` key is a bust:
https://github.com/jbeder/yaml-cpp/issues/60

Gonna have to rethink this a little bit.  Also I've found the ` NaluParsingHelper::find_nodes_given_key` doesn't do what I thought it did.  It just returns a node containing the key, so if you have multiple sub-nodes matching the key they wont get cached individually. I tested this with the `actuator` node in an input file and it doesn't catch multiple occurrences.

sayerhs:
@psakievich I had prototyped this when I was trying to unify paraview and normal Exodus output. Right now it is an either or situation. You can retain the idea of `output` section handling all the outputs. In the OutputInfo,  I checked the type of node for output. If it is a pure dictionary and not a sequence, then it is the legacy format. Emit a deprecated message and process a single output file. If it is a sequence then create multiple Exodus files and output subsets. 

Although from my conversations with @alanw0 it is my understanding that you can't output just the sidesets in Exodus. You need an element mesh. At that point you end up with a lot of transfers anyway and there's already an interface to do that.

alanw0:
@psakievich @sayerhs For outputting sidesets, I think you need at least some elements because in exodus a side is a pair involving the element and which side of the element. So if you want to output a sideset you at least need to also output the element that each side is attached to. We can get Tolu involved @tokusanya to give more insight to the best way to do it.

psakievich:
@alanw0 or @tokusanya so if we create a selector that catches the elements touching the sideset would that be sufficient? Basically the same selector we use for the surface operations?  @gantech would that be sufficient for your post-processing needs?

tokusanya:
@psakievich creating a selector for just the element blocks touching the sideset is sufficient. For the framework based codes, Greg Sjaardema had actually implemented a skin option in StkIo that would output the skinned mesh only but that is only for the exposed boundary surfaces and not a specific sideset :(

unit_tests/UnitTestOutputManager.C
tokusanya: @psakievich I don't know how nalu represents meshes or if you have some simple fixture that allows you to create a small mesh with multiple data that you would want to output to different output streams but STK unit tests typically create small easily managed meshes that it can manipulate for testing.

sayerhs: Here is the mesh creation function used by unit tests: 

https://github.com/Exawind/nalu-wind/blob/154f167f063c68096bf5f55152fa5b0b4c568993/unit_tests/UnitTestUtils.C#L74-L88

sayerhs:
@tokusanya The use-case being considered here would be almost always exposed boundary surfaces. If those exposed boundary faces belong to a sideset (and are the only faces in the sideset), would it be possible to use Greg's skin option? Would it also output the associated fields with it? 

Also how would one create a selector for the elements touching the sideset?

tokusanya:
@sayerhs That changes things a little ... if you were interested in outputting all the exposed boundary facets without regards as to whether they belong to a specific sideset, then it might be possible to wire in Greg's skin option if we are given a part that represents all the skin blocks.

Creating a selector for the elements touching a sideset is pretty easy. I assume you have a surface part that represents the sideset. In that case, you can use 

std::vector<const stk::mesh::Part*> stk::mesh::MetaData::get_blocks_touching_surface(const stk::mesh::Part* surface);

to get the list of element blocks touching the sideset. You can now construct a selector from that vector list using

stk::mesh::Selector selectUnion( const std::vector<const stk::mesh::Part*>&  );

sayerhs:
@tokusanya the element block touching the sidesets is the entire mesh for Ganesh's use case. We can already output the blade mesh alone. That's still millions of cells.

tokusanya:
@sayerhs Hmm, so what i'm understanding here is that you would need to output the entire element block list because the sideset you are interested in covers the entire model?

sayerhs:
@tokusanya yeah, that's how I am understanding the current situation. For example, consider a mesh with the following specs

```
generated:512x256x64|bbox:0.0,0.0,0.0,200.0,100.0,50.0,|sideset:xXyYzZ|show
```

If I wanted to output the sideset `X` and the Fields associated with `X`, how would I do it without the element block? The only way I can currently see this happening is if I split the mesh into two blocks such that one block is just a couple of layers normal to `X` sideset. That split is easy in this box mesh, sometimes is not simple in a blade mesh.

tokusanya:
@alanw0 @sayerhs Option 2 is pretty much the skin mesh option that i had mentioned earlier. With some work, it should even be possible to limit it to specific sidesets

sayerhs:
@alanw0 @tokusanya Option 2, with specific sidesets, would be the preferred one. I have previously used shell element type to output sidesets in [Nalu wind-utils](https://github.com/Exawind/wind-utils/blob/master/src/preprocessing/BdyIOPlanes.cpp#L75). That works well.

psakievich:
@alanw0 and @tokusanya could we get option 2 on the docket for the stk team to implement?  The ability to just dump out a sideset as a readable exodus file would be really nice. I've wanted this feature on other projects outside Exawind as well.

alanw0:
@tokusanya let's put a little effort into estimating how big a job this would be.
Like @sayerhs mentions, there will be some complication with getting fields written out, especially if converting sides to shell elements.

psakievich:
@alanw0 what's the status on this? I believe we left this open so you could add a story on the stk backlog to add this capability natively into stk.  Are we ready to close this yet?

alanw0:
@psakievich Yes I think we can close this. I have added a story to the stk backlog for this. We'll try to get to it soon.

PR #874, opened 2021-07-23
WIP: Add mesh motion to io realm

psakievich:
This is to allow mesh motion to the input_output realm so we can just output surfaces of the blades for turbine runs.
It is not as robust of a solution as those discussed in #870 but is a temporary stop gap for finishing the runs for the IEA29 simulations.

I'm still working through a couple seg faults, but I won't be able to get to this again until next week.  Posting as a draft in case someone wants to pick it up and finish it and to get comments on things I might be missing.

### DIFF
--- a/src/InputOutputRealm.C
+++ b/src/InputOutputRealm.C
@@ -13,6 +13,7 @@
 #include <NaluParsing.h>
 #include <Realm.h>
 #include <SolutionOptions.h>
+#include <mesh_motion/MeshMotionAlg.h>
 
 // transfer
 #include <xfer/Transfer.h>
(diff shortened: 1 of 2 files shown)


src/InputOutputRealm.C
```diff
@@ -64,15 +65,45 @@ InputOutputRealm::initialize_prolog()
...
+  if (solutionOptions_->meshMotion_)
+    register_mesh_motion_fields();
+
   ioBroker_->populate_mesh();
   ioBroker_->populate_field_data();
   create_output_mesh();
   input_variables_from_mesh();
+  if (solutionOptions_->meshMotion_) {
+    init_current_coordinates();
+    meshMotionAlg_->initialize(get_current_time());
```
psakievich, line 77: Seg faulting here inside the `get_current_time()` call. End of the stack trace is inside the `timeIntegrator_` when it should be returning the `double` for time.

src/InputOutputRealm.C
```diff
@@ -64,15 +65,45 @@ InputOutputRealm::initialize_prolog()
...
+  for (auto&& info : inputOutputFieldInfo_) {
+    std::vector<std::string> targetNames = info->targetNames_;
+    // loop over target parts and declare/put the field
+    for (size_t j = 0; j < targetNames.size(); ++j) {
+      const std::string targetName = targetNames[j];
+      stk::mesh::Part* part = metaData_->get_part(targetName);
+
+      VectorFieldType* displacement =
+        &(metaData_->declare_field<VectorFieldType>(
+          stk::topology::NODE_RANK, "mesh_displacement"));
```
neilmatula, line 97: Could this and the other VectorFieldType be pulled outside the loop?

sayerhs:
Why is this necessary? Given the use-case, the donor/receiver element/node pairs can be determined by the base (undisplaced) mesh in STK::Transfer. All one would need to do is transfer `mesh_displacement` to the IORealm mesh and output it for further post-processing. Wouldn't it be simpler to just replace the coordinates name for the from/to Realms with the one from MetaData? 

https://github.com/Exawind/nalu-wind/blob/8e0c05281b3bb30cdfa4121233e4daa139f36789/src/xfer/Transfer.C#L342

psakievich:
Sure that's another option. I'll take a look at that route.

psakievich:
Trying @sayerhs suggestion, [link](https://github.com/psakievich/nalu-wind/commit/eec7b6eeac61d9cd2c7ad18a30276355f821320b) by manually setting the coordinates, but I'm getting some failures in the linear system assembly. Also the search is still not converging in the test case I've setup so there is still something screwy going on.

psakievich:
closing due to #875

PR #875, opened 2021-07-28
Sideset writer that doesn't require transfers

### LINKED ISSUE #859: Extract surface data during run

psakievich:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

Add a sideset writer option to nalu-wind that doesn't require transfers.  Thanks to @rcknaus for putting the utility together.  I've tested this on the ellipse problem and it works well. Resolves #859.  Removes need for #874 and also #870.  However it would still be best to have the stk team put this functionality into stk for an optimized long term solution as noted in the comments of #870.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [x] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*

### DIFF
--- a/include/Realm.h
+++ b/include/Realm.h
@@ -78,6 +78,7 @@ class MeshMotionAlg;
 class MeshTransformationAlg;
 
 class SolutionNormPostProcessing;
+class SideWriterContainer;
 class TurbulenceAveragingPostProcessing;
 class DataProbePostProcessing;
 struct ActuatorModel;
@@ -417,6 +418,7 @@ class Realm {
   stk::mesh::MetaData *metaData_;
   stk::mesh::BulkData *bulkData_;
   stk::io::StkMeshIoBroker *ioBroker_;
+  std::unique_ptr<SideWriterContainer> sideWriters_;
 
   size_t resultsFileIndex_;
   size_t restartFileIndex_;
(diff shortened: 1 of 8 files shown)


psakievich:
@jrood-nrel looks like the Trilinos version being used by the CI isn't current.

jrood-nrel:
Yeah the docker images needs to be updated.

psakievich:
Anyone want to comment on this? the CI is broken for now.  I'd like to merge this unless anyone has issues.

rcknaus (approved):
Looks good. May want to modify a regression test to use this capability as well.

psakievich:
@rcknaus good point. I've added to a regression test. We can also use this feature to output precursor planes for abl simulations.

PR #889, opened 2021-09-09
Fix ADM turbine yaw

lawrenceccheung:
**Pull-request type:**
- [x] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

The current ADM implementation has a bug where the turbine yaw is fixed to 270 degrees (pointing in the -X direction), regardless of what the ElastoDyn input `NacYaw` says.  I've changed it so that the initialization of the swept points follows more closely the structure of the ALM implementation.  This allows the ADM actuator point orientation to be correctly read from OpenFAST after the first set of calls.

_However_, there is a performance penalty -- somewhere around 10% in the multi-turbine tests I've tried.  This is due to the fact that `initialize_swept_points()` the associated search is called every timestep.  We should definitely find a way to remove this penalty, and I am open to suggestions.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [x] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/actuator/ActuatorBulkDiskFAST.C
+++ b/src/actuator/ActuatorBulkDiskFAST.C
@@ -42,6 +42,15 @@ ActuatorBulkDiskFAST::ActuatorBulkDiskFAST(
     });
   compute_offsets(actMeta);
   init_epsilon(actMeta);
+  // LCC: Debug ADM turbine yaw
+  if (openFast_.isTimeZero()) {
+    if (openFast_.isDebug()) {
+      openFast_.solution0();
+    } else {
+      squash_fast_output(std::bind(&fast::OpenFAST::solution0, &openFast_));
+    }
+  }
+  // -------------------------
   RunActFastUpdatePoints(*this);
   initialize_swept_points(actMeta);
 }
--- a/src/actuator/ActuatorBulkDiskFAST.C
+++ b/src/actuator/ActuatorBulkDiskFAST.C
@@ -42,24 +42,20 @@ ActuatorBulkDiskFAST::ActuatorBulkDiskFAST(
     });
   compute_offsets(actMeta);
   init_epsilon(actMeta);
-  //RunActFastUpdatePoints(*this);
-  //initialize_swept_points(actMeta);
+  RunActFastUpdatePoints(*this);
+  initialize_swept_points(actMeta);
 }
 
 void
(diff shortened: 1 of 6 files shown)


src/actuator/ActuatorBulkDiskFAST.C
```diff
@@ -48,6 +48,17 @@ ActuatorBulkDiskFAST::ActuatorBulkDiskFAST(
...
 }
 
+void
+ActuatorBulkDiskFAST::initialize_ADM_points(const ActuatorMetaFAST& actMeta)
+{
+  // LCC NOTE: Currently this is called every iteration -- incurring a
+  // noticeable performance penalty. Need to fix so that it's only
+  // called in the beginning (for fixed yaw), or called when the
+  // turbine yaw changes.
+  initialize_swept_points(actMeta);
```
psakievich, line 58:
```suggestion
if(!initialized_ADM_points){
  RunActFastUpdatePoints(*this);
  initialize_swept_points(actMeta);
  stk_search_act_pnts(actMeta, stkBulk_);
  initialized_ADM_points = true;
 }    
```

psakievich: I was looking through the commit history and I'm wondering if you tried this when you were investigating this hanging?  You'd need to change the function signature and remove these added calls from `ActuatorExecutorFASTNgp.C`.

psakievich (changes requested):
Let me know if you've tried my suggested changes already. If so then I'll approve.

src/actuator/ActuatorBulkDiskFAST.C
lawrenceccheung: Sorry for the super long delay in response.  Yes, I think I did actually try something similar to that last year, but the combination that I put in there was the one that didn't hang.

Lawrence

rcknaus (approved):

tonyinme (commented):
This is looking good.
Is the function to "initialize swept points" being called every time step? Is there a need for that? Or can it be called only in the first time step?

psakievich:
@tonyinme it is only called on the first time step

PR #890, opened 2021-09-19
Add a sinusoidal rotation mesh motion kernel

gantech:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

Adds a mesh motion kernel corresponding to a sinusoidal mesh motion of the type 

theta = ampl * sin (omega * t)

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [x] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*

### DIFF
--- a/include/mesh_motion/MotionSinRotationKernel.h
+++ b/include/mesh_motion/MotionSinRotationKernel.h
@@ -1,0 +1,62 @@
+#ifndef MOTIONSINROTATIONKERNEL_H
+#define MOTIONSINROTATIONKERNEL_H
+
+#include "NgpMotion.h"
+
+namespace sierra{
+namespace nalu{
+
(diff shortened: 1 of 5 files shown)


ashesh2512:
@gantech So the only thing different than the rotation motion is how the angle changes every time step, is that correct ? If so, why not just introduce a user input option, `is_sinusoidal` in the rotation kernel which if set to true in the input file will update the angle every time step ?

gantech:
That's a great idea @ashesh2512. Let me try reformatting this and putting it in as a new pull request.

PR #880, opened 2021-08-31
GCL Rebase

psakievich:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

## Description of the pull-request

PR to replicate #826.  Inspection of #826 showed some rebase issues in the regression test file and `LowMachEquationSystem` so I decided to try again with a clean slate.  I removed some of the whitespace changes from the formatter, squashed and rebased off the latest master.

This removes some of the git history, but since we are going to squash anyways I don't think it matters. I added the main contributors to this code as Co-Authors to try and preserve some history. 
Initially placing as a draft request, and I will move it to a full PR once I have tested locally. 

## Checklist

**NOTE** The checklist below is a suggestion and not mandatory. You don't have
to _check all boxes_ to submit a pull request. However, please add a brief
explanation in your pull request summary explaining the omission for the benefit
of reviewers and developers.

*All pull requests*
- [ ] Builds successfully (*must test on at least one system/compiler combination*)
  - Operating systems 
    - [ ] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Compiles without warnings
- [ ] Passes all unit tests
- [ ] Passes all regression tests
- [ ] Documentation builds without errors

*For new code updates*
- [ ] Documentation updates - additions to user, theory, and verification manuals
- [ ] New unit tests providing coverage for new code, bug fixes etc.
- [ ] New regression tests exercising the new code

PR #896, opened 2021-10-21
Restore accidently removed lines GCL

psakievich:
These lines were accidentally removed as part of #880 introducing a bug.


psakievich:
sounds good. I tested locally and it removed the diffs there. Hopefully this is it 🤞

PR #900, opened 2021-10-29
Adding dynamic forcing for momentum

marchdf:
**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [X] Feature enhancement

This PR adds a dynamic forcing option for momentum forcing in a box. In this input file, it can be enabled as such (the new section is just the `dynamic_body_force_box_parameters` section):

```
        - source_terms:
            momentum: body_force_box

        - source_term_parameters:
            momentum: [0.011, 0.0, 0.0]
            momentum_box: [-100.0, -100.0, -100.0, 100.0, 100.0, 100.0]

        - dynamic_body_force_box_parameters:
            forcing_direction: 0
            velocity_reference: 21.0
            density_reference: 1.0
            velocity_target_name: inlet
            drag_target_name: [top, bottom]
            output_file_name: forcing.dat
```

The user basically specifies the target bulk velocity and reference density. These will be the targeted momentum values at the `velocity_target_name` sideset. The forcing is computed as 

```
f = drag / volume - (mdot - densityRef * uRef * mdotArea) / (mdotArea * dt)
```

This feature was used in the DES and IDDES simulations of the periodic hill and was instrumental in achieving the correct flow.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [X] Linux
    - [ ] MacOS
  - Compilers 
    - [X] GCC
    - [ ] LLVM/Clang
    - [X] Intel compilers
    - [ ] NVIDIA CUDA
- [X] Passes all unit tests

*For new code updates*
- [X] Documentation updates - additions to user, theory, and verification manuals
- [X] New regression tests exercising the new code

### DIFF
--- a/src/node_kernels/MomentumBodyForceBoxNodeKernel.C
+++ b/src/node_kernels/MomentumBodyForceBoxNodeKernel.C
@@ -10,30 +10,102 @@
 #include "node_kernels/MomentumBodyForceBoxNodeKernel.h"
 #include "Realm.h"
 #include "SolutionOptions.h"
+#include "LowMachEquationSystem.h"
+#include "NaluEnv.h"
+#include "master_element/MasterElementFactory.h"
+#include "TimeIntegrator.h"
 
 #include "stk_mesh/base/MetaData.hpp"
 #include "stk_mesh/base/Types.hpp"
+#include "stk_mesh/base/Selector.hpp"
 #include "utils/StkHelpers.h"
+#include <stk_util/parallel/ParallelReduce.hpp>
+
+#include "ngp_utils/NgpLoopUtils.h"
+
+#include <fstream>
+#include <iomanip>
 
 namespace sierra {
 namespace nalu {
 
 MomentumBodyForceBoxNodeKernel::MomentumBodyForceBoxNodeKernel(
-  const stk::mesh::BulkData& bulk,
-  const std::string coordName,
+  Realm& realm,
   const std::vector<double>& forces,
   const std::vector<double>& box)
   : NGPNodeKernel<MomentumBodyForceBoxNodeKernel>(),
-    coordinatesID_(get_field_ordinal(bulk.mesh_meta_data(), coordName)),
+    nDim_(realm.meta_data().spatial_dimension()),
+    coordinatesID_(get_field_ordinal(
+      realm.meta_data(), realm.solutionOptions_->get_coordinates_name())),
     dualNodalVolumeID_(
-      get_field_ordinal(bulk.mesh_meta_data(), "dual_nodal_volume")),
-    nDim_(bulk.mesh_meta_data().spatial_dimension())
+      get_field_ordinal(realm.meta_data(), "dual_nodal_volume")),
+    outputFileName_(realm.solutionOptions_->dynamicBodyForceOutFile_),
+    dynamic_(realm.solutionOptions_->dynamicBodyForceBox_),
+    w_(16)
 {
   for (int i = 0; i < nDim_; ++i)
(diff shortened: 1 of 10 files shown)


src/node_kernels/MomentumBodyForceBoxNodeKernel.C
```diff
@@ -10,31 +10,103 @@
...
+#include <fstream>
+#include <iomanip>
 
 namespace sierra {
 namespace nalu {
 
 MomentumBodyForceBoxNodeKernel::MomentumBodyForceBoxNodeKernel(
-  const stk::mesh::BulkData& bulk,
-  const std::string coordName,
+  Realm& realm,
```
psakievich, line 33: I would prefer if we didn't pass `Realm` into this object, but rather added a flags or a struct. In general I think having Realm be a god object that contains all the options is one of the things that makes nalu so confusing to work with, hard to read and very difficult to test. Any new code we add should steer away from that practice IMO.

psakievich (approved):
Looks good to me. Just the realm passing thing I commented on, but that is more of a preference.  I hope to someday refactor out the dependency on Realm from the algs and kernels.

src/node_kernels/MomentumBodyForceBoxNodeKernel.C
marchdf: I also agreed with your sentiment, Phil. Realm does really complicate things... However it does look like fixing this would be a larger scope of work that this PR...

sayerhs: I think there is general agreement that Realm and SolutionOptions are not great. You just have to look at the amount of scaffolding we need to create a unit test. However I am not convinced that having skinny structs is any better. You are just moving the problem from the constructor to another place where the struct gets populated. You just need to look at Nalu Parsing to see why that idea can quickly get out of hand. You could just pass the appropriate YAML node as an argument and let the class manage the data and associated checks it needs to do. Right now you've to change things at multiple places to add a parameter to a kernel. Structs don't eliminate the problem and at least I can't see the need for another object when the class can handle this responsibility. 

Of course the argument for not passing YAML nodes but use a struct is: what if we ever wanted to support some other format or use another library.

sayerhs: In fact, passing the appropriate YAML node can be done quite easily with the current codebase quite easily. But is it ok to expose implementation details in Momentum EQS

psakievich: I was actually thinking about starting a github discussion about this very thing.  I agree that just passing yaml would be the best way to do things. Use Realm to store the algs and use factories to create them that take in yaml nodes.  I would like to detangle some of this incrementally throughout the year. I think the advantage to skinny structs is it is an incremental way of moving in that direction.  It's definitely not the solution, but I see it is a start toward encapsulation. If things don't depend on Realm but rather structs means that Realm can be changed easier without cascading consequences because the items the dependencies go through a filtering state as they leave Realm by constructing these skinny structs.  I'm sure there are several ways to go about it, but in general I feel that perpetuating the use of Realm as a mega container will only make the problem harder to solve in the long run.  I would love to see solution options completely removed. One can dream... 

@marchdf sorry to hijack your PR.  feel free to merge whenever you'd like.

sayerhs: @psakievich Please let me know when you move this discussion to a separate Issue. Sorry from my end too @marchdf

PR #904, opened 2021-11-15
Fix stk modify-flags errors.

alanw0:
It appears as though there were some missing sync_to_device calls,
which were previously ignored due to stk's less-than-rigorous
checking, which has now been strengthened somewhat.

Also added an error-print in MeshVelocityAlg because some template
instantiations were using Hex27 elements with an algorithm that is
set up to only support 12 integration points. Hex27 has 216. This is
apparently not being hit in current testing, but some builds were
noticing the problem since AlgTraits::numScsIp_ is a compile-time
number.

**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/gcl/MeshVelocityAlg.C
+++ b/src/gcl/MeshVelocityAlg.C
@@ -135,7 +135,12 @@ MeshVelocityAlg<AlgTraits>::execute()
         }
       }
 
-      for (int ip = 0; ip < AlgTraits::numScsIp_; ++ip) {
+      int nip = std::min(12,AlgTraits::numScsIp_);
+      if (AlgTraits::numScsIp_ > nip) {
+        std::cerr << "Error: MeshVelocityAlg can't support more than 12 integration points." << std::endl;
+      }
+
+      for (int ip = 0; ip < nip; ++ip) {
 
         DoubleType scs_vol_coords[8][3];
 
--- a/include/ngp_utils/NgpFieldBLAS.h
+++ b/include/ngp_utils/NgpFieldBLAS.h
@@ -46,6 +46,8 @@ inline void field_axpby(
   using Traits = NGPMeshTraits<Mesh>;
   using MeshIndex = typename Traits::MeshIndex;
 
+  yField.sync_to_device();
+
   nalu_ngp::run_entity_algorithm(
     "ngp_field_axpby",
     ngpMesh, rank, sel,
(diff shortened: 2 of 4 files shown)


psakievich:
@alanw0 thanks. Looks like some of these missed sync calls came from #880.

alanw0:
Note that this commit should fix the cpu tests that were broken by the stk modify-flags errors. However, there are still some GPU crashes that we're debugging. One is in actuator-search, I'll put up a pull-request for that shortly. Another seems related to master-elements, we're still diagnosing that one.

psakievich:
@alanw0 that is interesting that actuator-search is causing a GPU crash. That should all execute on host.  These could be related/ the same error though since the actuator search will do an interpolation and has calls to master elements.

jrood-nrel (approved):
Excellent, thanks for this @alanw0

alanw0:
> @alanw0 that is interesting that actuator-search is causing a GPU crash. That should all execute on host. These could be related/ the same error though since the actuator search will do an interpolation and has calls to master elements.

Phil it seems to be just a simple issue with passing kokkos-dual-views into a search, where a call to 'modify_host' was tripping a kokkos error-check that's only active in debug builds. I'll put up the pull-request here in a minute, it's a very simple fix.

psakievich (approved):

src/gcl/MeshVelocityAlg.C
```diff
@@ -135,7 +135,12 @@ MeshVelocityAlg<AlgTraits>::execute()
         }
       }
 
-      for (int ip = 0; ip < AlgTraits::numScsIp_; ++ip) {
+      int nip = std::min(12,AlgTraits::numScsIp_);
+      if (AlgTraits::numScsIp_ > nip) {
+        std::cerr << "Error: MeshVelocityAlg can't support more than 12 integration points." << std::endl;
+      }
+
+      for (int ip = 0; ip < nip; ++ip) {
```
sayerhs, line 143: @alanw0 This algorithm is only available for Hex8, Line 57 has a throw to disallow any other element types. The `std::cerr` message is not necessary.

overfelt (approved):

src/gcl/MeshVelocityAlg.C
alanw0: Hey Shreyas!
Ah, ok, I hadn't noticed line 57. I can remove the cerr. Of course it won't get hit anyway, since line 57 will stop things before we get that far. But adding the std::min line was necessary to get the compiler to stop noticing that the Hex27 template instantiation could execute the following for-loop more than 12 times which would over-run array bounds.

alanw0: @sayerhs Yes I'll do that. And also use AlgTraitsHex8::numScsIp_ in the declaration of the array at line 69 of MeshVelocityAlg.h.

PR #918, opened 2021-11-23
IDDES updates:

psakievich:
There appears to be a bug in the Strelets upwinding.  I believe it has to do with some terms not being computed on the first timestep. Possibly due to the order in which the algorithms are called.

However, since we haven't seen compelling evidence that we need the Strelets upwinding for our cases of interest, this PR
makes it so you have to explicitly turn it on when running IDDES with `strelets_upwinding: yes` in `solution_options`.
Other wise we use the default upwinding/Peclet blending methods in the code.
Users can only turn on Strelets upwinding with IDDES otherwise the parameter is ignored.

My advice for now: don't use Strelets upwinding.  I will try to debug it more in the coming weeks but this should enable IDDES for the blade resolved simulations.  @gantech now the linear solvers converge in 8 iterations when running the ellipse toy problem instead of maxing out at 200 iterations.  The scaled non-linear residuals go up a little past 1.0 in the second Picard iteration for the first 2-3 timesteps but then settle down to below 1 going forward. So far it seems this is just getting through the initial condition. The key differentiator is the linear solvers are doing just fine.

Additional fixes were required to get the sync/modified calls correct for IDDES specific fields similar to the work @alanw0 has been doing lately.

**Pull-request type:**
- [x] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [ ] Linux
    - [x] MacOS
  - Compilers 
    - [ ] GCC
    - [x] LLVM/Clang
    - [ ] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*

### DIFF
--- a/src/ShearStressTransportEquationSystem.C
+++ b/src/ShearStressTransportEquationSystem.C
@@ -514,6 +514,21 @@ ShearStressTransportEquationSystem::compute_f_one_blending()
 }
 
 void
+ShearStressTransportEquationSystem::pre_iter_work()
+{
+  const auto turbModel = realm_.solutionOptions_->turbulenceModel_;
+  if (turbModel == SST_IDDES) {
+    const auto& fieldMgr = realm_.ngp_field_manager();
+    const auto& meta = realm_.meta_data();
+    auto& bulk = realm_.bulk_data();
+
+    auto ngpIddesRans = fieldMgr.get_field<double>(
+      get_field_ordinal(meta, "iddes_rans_indicator"));
+    ngpIddesRans.sync_to_device();
+  }
+}
+
+void
 ShearStressTransportEquationSystem::post_iter_work()
 {
   const auto turbModel = realm_.solutionOptions_->turbulenceModel_;
@@ -524,6 +539,7 @@ ShearStressTransportEquationSystem::post_iter_work()
 
     auto ngpIddesRans = fieldMgr.get_field<double>(
       get_field_ordinal(meta, "iddes_rans_indicator"));
+    ngpIddesRans.modify_on_device();
     ngpIddesRans.sync_to_host();
 
     ScalarFieldType* iddesRansInd = meta.get_field<ScalarFieldType>(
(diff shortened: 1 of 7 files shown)


src/node_kernels/TKESSTIDDESNodeKernel.C
```diff
@@ -49,7 +49,7 @@ TKESSTIDDESNodeKernel::setup(Realm& realm)
   fOneBlend_ = fieldMgr.get_field<double>(fOneBlendID_);
   ransIndicator_ = fieldMgr.get_field<double>(ransIndicatorID_);
   // call modify before this field gets modified in kernel execute phase
-  ransIndicator_.modify_on_device();
+  ransIndicator_.sync_to_device();
```
alanw0, line 52: is there a modify_on_device for this field somewhere else?

alanw0 (approved):
Looks good to me in general, just the one question about switching the modify_on_device to sync_to_device.

marchdf (approved):
Thank you for doing this!

src/ShearStressTransportEquationSystem.C
```diff
@@ -524,6 +539,7 @@ ShearStressTransportEquationSystem::post_iter_work()
 
     auto ngpIddesRans = fieldMgr.get_field<double>(
       get_field_ordinal(meta, "iddes_rans_indicator"));
+    ngpIddesRans.modify_on_device();
```
psakievich, line 542: @alanw0  yes it is here in the `post_iter_work` function This is a conditional evaluation that only occurs when the IDDES model is being used.

src/node_kernels/TKESSTIDDESNodeKernel.C
psakievich: Actually looking at this again. this is probably an unnecessary call here

psakievich: Deleted it

PR #919, opened 2021-11-23
Modification to make FLLC compatible with ActDisk

psakievich:
Modify code to be make FLLC compatible with the ActuatorDisks. Testing and validation have been conducted by Myra Blaylock and Alan Hsieh.


tonyinme (approved):
This looks good. If I remember correctly, this change ensures that the FLLC is computed before spreading the forces. If not, the forces are too weak and the correction does not work properly, right?

psakievich:
@tonyinme yes that is correct.  The old call location was after we diluted the forces, so we just moved it to a point where they have yet to be diluted by the spreading operation.

PR #920, opened 2021-12-10
Reorganize test list and add capability for saving gold files

jrood-nrel:


psakievich (commented):
WIth this logic, how much effort would it be to remove the auto tolerance being strictly for gcc 8.4.0 and rather setting an auto tight tolerance when the compiler you're using matches the gold directory?  Or is it safe so that if you set the gold directory the absolute tolerance automatically gets set?  

I just want to avoid the case where I'm comparing against my own auto generated norms, but I'm using a tolerance of `0.05`.

reg_tests/CMakeLists.txt
```diff
@@ -12,27 +12,41 @@ if(RES_LEN EQUAL 0)
...
 endif()
 
-set(NALU_GOLD_NORMS_DIR "${CMAKE_CURRENT_SOURCE_DIR}" CACHE PATH
+set(NALU_WIND_REFERENCE_GOLDS_DIR "${CMAKE_CURRENT_SOURCE_DIR}/test_files" CACHE PATH
   "Custom directory for checking gold norms")
 
-if (NOT ${NALU_GOLD_NORMS_DIR} STREQUAL ${CMAKE_CURRENT_SOURCE_DIR})
-  message(STATUS "Setting custom gold norms path: ${NALU_GOLD_NORMS_DIR}")
-  set(TEST_TOLERANCE "1.0e-15")
+if (NOT ${NALU_WIND_REFERENCE_GOLDS_DIR} STREQUAL ${CMAKE_CURRENT_SOURCE_DIR})
```
sayerhs, line 18: This would make sure that if the user specifies a gold norms for, then test tolerances are strict

reg_tests/check_norms.py
```diff
@@ -54,6 +58,10 @@ def generate_test_norms(testname):
     cmdline = """awk '/Mean System Norm:/ { print $4, $5, $6; }' %s > %s """%(
         logname, norm_name)
     os.system(cmdline)
+    args = parse_arguments()
+    if (args.save_norm_file != None):
```
sayerhs, line 62: This looks like it will copy the norms before it checks that the test actually passed. Is that what was intended? Seems to me that the desired behavior would be to skip touching gold norms on each ctest run, at least making sure that it is not updated for failing tests.

jrood-nrel:
> WIth this logic, how much effort would it be to remove the auto tolerance being strictly for gcc 8.4.0 and rather setting an auto tight tolerance when the compiler you're using matches the gold directory? Or is it safe so that if you set the gold directory the absolute tolerance automatically gets set?
> 
> I just want to avoid the case where I'm comparing against my own auto generated norms, but I'm using a tolerance of `0.05`.

I'm thinking of removing the auto tolerance stuff to make the logic simpler. I'm starting to think I would keep gold files for each compiler like I do in AMR-Wind. Also with this it would hopefully be easier for a user to save and use their own gold files. We could still store the GCC norms here, but I would maybe remove tolerance checks for the default case.

reg_tests/check_norms.py
sayerhs: OK. So do I understand correctly that this breaks the current `NALU_GOLD_NORMS_DIR` developer workflow? The checks will always compare against the reference golds?

reg_tests/CTestList.cmake
```diff
@@ -2,79 +2,111 @@
...
+  set(OUTPUT_FILE_R0 "${testname}_R0.log")
+  set(OUTPUT_FILE_R1 "${testname}_R0.log")
+  set(OUTPUT_FILE_NP "${testname}Np${np}.log")
+  set(RUN_COMMAND "${MPI_COMMAND} -i ${INPUT_FILE} -o ${OUTPUT_FILE}")
+  set(RUN_COMMAND_RST "${MPI_COMMAND} -i ${INPUT_FILE_RST} -o ${OUTPUT_FILE_RST}")
+  set(RUN_COMMAND_R0 "${MPI_COMMAND} -i ${INPUT_FILE_R0} -o ${OUTPUT_FILE_R0}")
+  set(RUN_COMMAND_R1 " && ${MPI_COMMAND} -i ${INPUT_FILE_R1} -o ${OUTPUT_FILE_R1}")
+  set(RUN_COMMAND_NP "${MPI_COMMAND} -i ${INPUT_FILE} -o ${OUTPUT_FILE_NP}")
+  set(COMPARE_GOLDS_COMMAND_BASE " && ${CMAKE_CURRENT_SOURCE_DIR}/check_norms.py --abs-tol ${ABS_TOLERANCE} --rel-tol ${REL_TOLERANCE}")
+  set(COMPARE_GOLDS_COMMAND "${COMPARE_GOLDS_COMMAND_BASE} ${testname} ${NALU_WIND_REFERENCE_GOLDS_DIR}/${testname}/${testname}.norm.gold ${SAVE_GOLDS_COMMAND}")
```
sayerhs, line 31: How does one indicate to CTest that they want to compare against golds saved in a custom gold norms dir?

psakievich (approved):

PR #921, opened 2021-12-16
Added initial framework for solution of the One-Eqn Correlation Based…

wilsonrvornlgov:
… Transition Model of Menter 2015

**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [x] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*


wilsonrvornlgov:
This pull request is for the initial framework for the Gamma correlation-based transition model of Menter (2015) into the Nalu-Wind flow solver.  The commits in this pull request add model constants and enumerations for the Gamma transport equation. 

Future incremental pull requests will add solution of the Gamma transport equation and its modification of the k-omega source terms to compute transition. Additional test cases will also be added.

In addition, files have been added to reg_tests directory for the ERCOFTAC T3A flat plate test case.

include/NaluParsedTypes.h
```diff
@@ -54,6 +54,20 @@ struct SpecDissRate {
   {}
 };
 
+struct GammaInf {
+  double gamma_;
+  GammaInf()
+    : gamma_(1.0)
```
marchdf, line 60: Most of these default to `0.0`. Are you sure you want both of these Gamma ones to default to 1.0?

reg_tests/test_files/BLTFlatPlateT3A/BLTFlatPlateT3A.norm.gold
```diff
@@ -0,0 +1,5 @@
+137372539.0984875 1 0.01
```
marchdf, line 1: I think we typically let Jon generate the golds for the machines the tests are run on. Not super important.

jamelvin (approved):
Looks fine to me.  It seems like the decision was for you to add this incrementally?  This pull request adds some pieces of infrastructure that I don't think are used in this merge (i.e. a new equation system value and some model constants), unless I'm misunderstanding?

This may get a little confusing later as that other infrastructure is added with relevant info in these prior pull requests.  I don't know if there is any other way to break things up, but I'll defer to everyone else on what makes the most sense.

include/NaluParsedTypes.h
wilsonrvornlgov: Thanks for your comment. I was using GammaInf to set the value of gamma to unity at the inflow (see GammaEquationSystem.C). I believe that Shreyas suggested that we set the inflow BC using a 'GammaInf struct'.

wilsonrvornlgov:
> Looks fine to me. It seems like the decision was for you to add this incrementally? This pull request adds some pieces of infrastructure that I don't think are used in this merge (i.e. a new equation system value and some model constants), unless I'm misunderstanding?
> 
> This may get a little confusing later as that other infrastructure is added with relevant info in these prior pull requests. I don't know if there is any other way to break things up, but I'll defer to everyone else on what makes the most sense.

Thanks for your comment. Yes, the pull request adds initial infrastructure for adding the Gamma transport equation in the next MR. I could have probably combined this MR and the next into one MR, but I was advised to keep MR as small as possible. It should not break any of the existing test cases. I will let the reviewers advice me on this issue.

reg_tests/test_files/BLTFlatPlateT3A/BLTFlatPlateT3A.norm.gold
psakievich: Yeah this is fine. @jrood-nrel will just rebless it on the test machine.

psakievich (approved):
Sorry for the late review.  @wilsonrvornlgov thanks for breaking this up.  So if I understand this correctly you've added a flat plate regression test that doesn't use any new features yet, but includes the gamma transition flag (which is unused) and some model constants?  I'm fine merging this if you'd like.  If you wouldn't mind adding a few sentences into the PR description describing the extent of the PR and the intention of follow on PR's that would be very helpful.  I usually don't add regression tests until the new features are being exercised, but I understand you're breaking up a very large PR into smaller ones which is a big task.

I will be more responsive to the follow on PR's.  Feel free to merge this.  Our code culture isn't one that you need to have every requested reviewer approve, but if you'd like specific people to weigh in before merging feel free to ping them directly to indicate this.

wilsonrvornlgov:
> Sorry for the late review. @wilsonrvornlgov thanks for breaking this up. So if I understand this correctly you've added a flat plate regression test that doesn't use any new features yet, but includes the gamma transition flag (which is unused) and some model constants? I'm fine merging this if you'd like. If you wouldn't mind adding a few sentences into the PR description describing the extent of the PR and the intention of follow on PR's that would be very helpful. I usually don't add regression tests until the new features are being exercised, but I understand you're breaking up a very large PR into smaller ones which is a big task.
> 
> I will be more responsive to the follow on PR's. Feel free to merge this. Our code culture isn't one that you need to have every requested reviewer approve, but if you'd like specific people to weigh in before merging feel free to ping them directly to indicate this.

Thanks for the review and comments. Yes, your understanding is correct. It probably makes more sense to add the transition test case after the initial solution of the Gamma transition equation is completed in the next PR.  I have had some feedback to add a basic test case as soon as possible, thus the inclusion in this PR. We can wait to add the test case until the next PR if you prefer.

I added additional comments to the PR description to clarify these points. If you approve, can you merge this PR in the master branch? I don't have permissions to do so.

PR #924, opened 2022-01-07
Restarts with mesh motion/displacements

psakievich:
There are a couple of issues with the way the code currently creates and populates the mesh motion algorithms related to restarts.

First all the initialization occurs in `Realm::initialize_prolog` however, we don't detect that we are doing a restart until `TimeIntegrator::prepare_for_time_integration`. 

My initial idea is to reset the coordinates and displacements in the TimeIntegrator function call, and re-initialization of the meshMotionAlg.  I've started outlining this in the first commit of this PR.  I could still be missing something here though.  

I am going to do some more exploring.  Ideally I'd like to refactor all of this so the time integrator gets created and set before the realms, so we can just know if we are restarting when we construct the realms.  That may be more work than the current timeline will allot for, but it is something I think we should do in the nearish future anyways.


psakievich:
@gantech @ashesh2512 I've tested this on the moving ellipse problem and it works there.  The ellipse is in the right location upon restart. I'm running it out a little further. I think for a regression test we should add a 2d version of that problem to the repo.

psakievich:
@ashesh2512 so this latest commit is failing on the second time step of my regression test (first time step of the restart solves fine).  Looks like the map for the tpetra linear system is getting messed up. 

```
Tpetra::Map constructor (noncontiguous): Minimum global ID = -1152921504606846976 over all process(es) is less than the given indexBase = 1.
libc++abi.dylib: terminating with uncaught exception of type std::invalid_argument: /Users/psakiev/soft/spack-manager/stage/spack-stage-trilinos-develop-helbhlwhzlurushrpij7umwosdkj5etl/spack-src/packages/tpetra/core/src/Tpetra_Map_def.hpp:806:

Throw number = 1

Throw test that evaluated to true: minAllGID_ < indexBase_

Tpetra::Map constructor (noncontiguous): Minimum global ID = -1152921504606846976 over all process(es) is less than the given indexBase = 1.
```
and the logfile looks like this:
```
 Realm shall provide restart files at: currentTime/timeStepCount: 0.45/15 (realm_1)
 *******************************************************
 Time Step Count: 16 Current Time: 0.48
  dtN: 0.03 dtNm1: 0.03 gammas: 1.5 -2 0.5 
  DualNodalVolume min: 2.22982e-06 max: 0.0005 total: 61.6835
 TIOGA: Overset algorithm will ghost 586 elements
 TIOGA: Num. receptor nodes = 3154
  Max Courant: 11.9177 Max Reynolds: 189.2 (realm_1)
    Realm Nonlinear Iteration: 1/1 
 
 realm_1::advance_time_step() 
 NLI    Name           Linear Iter      Linear Res     NLinear Res    Scaled NLR 
 ---    ----           -----------      ----------     -----------    ----------
 1/3    Equation System Iteration
  1/1      myLowMach
         MomentumEQS           11       1.98061e-05      0.0075717             1
         ContinuityEQS          7       1.15806e-06     0.00107155             1
 2/3    Equation System Iteration
  1/1      myLowMach
         MomentumEQS           17       2.39927e-07    8.31548e-05     0.0109823
         ContinuityEQS          7       8.92273e-07    0.000536554      0.500728
 3/3    Equation System Iteration
  1/1      myLowMach
         MomentumEQS           16        1.4806e-07    3.96961e-05    0.00524269
         ContinuityEQS          6       7.58397e-07    0.000265342      0.247625
 Mass Balance Review:
 Density accumulation: 0
 Integrated inflow:    0   
 Integrated open:      0.002453923769963097
 Total mass closure:   0.002453923769963097
 Filter Size 0.5400000000000003
 Realm shall provide output files at : currentTime/timeStepCount: 0.4800000000000002/16 (realm_1)
 Mean System Norm: 0.0001525191878451368 16 0.48
 WallClockTime: 16 Pre: 0.478253 NLI: 3.59799 Post: 0.0218339 Total: 4.09807
 *******************************************************
 Time Step Count: 17 Current Time: 0.51
  dtN: 0.03 dtNm1: 0.03 gammas: 1.5 -2 0.5 
  DualNodalVolume min: 2.22982e-06 max: 0.0005 total: 61.6835
 TIOGA: Overset algorithm will ghost 597 elements
 TIOGA: Num. receptor nodes = 3154
 ```
 Documenting here, and we can circle back on this tomorrow

**Comment by psakievich** [comment]
Here's a picture of the new test in action. Mesh is in the right place on the restart though so we are close. 
image

**Comment by ashesh2512** [comment]
@psakievich 

Looks like you are using coupled overset. Could you please turn on decoupled overset - https://github.com/Exawind/nalu-wind/blob/master/reg_tests/test_files/oversetRotCylNGPTrilinos/oversetRotCylNGPTrilinos.yaml#L41

Additionally, can you please make sure that for nalu-only runs tioga is being compiled with -DTIOGA_HAS_NODEGID=ON

**Comment by psakievich** [comment]
Sure I'll try that today

File: reg_tests/test_files/oversetMovingCylinder/oversetMovingCylinder.yaml [hunk 1]
```diff
@@ -0,0 +1,232 @@
...
+  - name: solve_scalar
+    type: tpetra
+    method: gmres
+    preconditioner: sgs
+    tolerance: 1e-5
+    max_iterations: 50
+    kspace: 50
+    output_level: 0
+
+  # TODO -TPETRA Solver is failing with this test
```
psakievich, line 16: FYI @jhux2. This may be something simple but I haven't had a chance to debug it.

reg_tests/test_files/oversetMovingCylinder/oversetMovingCylinder_rst.yaml
```diff
@@ -0,0 +1,230 @@
...
+
+  - name: realm_1
+    mesh: rst/oversetMovingCylinder.rst
+    use_edges: yes
+    activate_aura: true
+
+    equation_systems:
+      name: theEqSys
+      max_iterations: 3
+      # TODO: restart fails for coupled overset on start of second timestep
```
psakievich, line 45: FYI @alanw0. This appears to be a break that went unnoticed.  @ashesh2512 said it was working it the not too distant past.  We're not sure if this was a nalu-wind change, a change to ghosting in stk or something else. It would be really helpful if the stk team could look into this.

alanw0: I don't see this test on the dashboard. Is it a nightly test? Can you produce a set of inputs that currently fail and that worked at some previous point in time? Then we could help figure out what happened.

psakievich: @ashesh2512 do you have a timeline of when you last ran something with coupled overset and it worked?

ashesh2512: @psakievich That would around mid 2020. We had a milestone that focused on comparison between coupled and decoupled overset. I also want to point out, that the coupled overset appears to do fine for stationary meshes. Moving meshes with significant change in ghosting is what appears to breaks the code.

src/TimeIntegrator.C
```diff
@@ -178,7 +178,12 @@ void TimeIntegrator::prepare_for_time_integration()
 
   // possible restart; need to extract current time (max wins)
   for ( ii = realmVec_.begin(); ii!=realmVec_.end(); ++ii) {
+    const double time_coming_in = currentTime_;
     currentTime_ = std::max(currentTime_, (*ii)->populate_restart(timeStepNm1_, timeStepCount_));
+    // we can do exact comparison because the numbers will be bit-wise exact the
+    // same unless the value was changed by the restart
+    if (time_coming_in != currentTime_)
+      overset_->update_connectivity();
```
ashesh2512, line 186: @psakievich Why is this needed following - https://github.com/Exawind/nalu-wind/pull/924/files#diff-fecc3b6f108bff965d05cbaa485de2d60d273e8f71457705898ac4dd7909c36dR3216 ? What would be the use case here ?

psakievich: @ashesh2512 `prepare_for_time_integration` is called every timestep but we only need to do this extra update if there was a restart. https://github.com/Exawind/nalu-wind/blob/ad5ffdf22df0f5ef174383759359978d52de61f3/src/TimeIntegrator.C#L307-L312

reg_tests/test_files/oversetMovingCylinder/oversetMovingCylinder.yaml
psakievich: Looks like all the overset cases using Muelu for continuity are failing for me on my mac.  Not sure what's going on with that. @jrood-nrel can you check on you mac and see if you get the same thing

src/TimeIntegrator.C
psakievich: Moving this call outside the realm vector loop and removing the call inside realm.

ashesh2512 (approved):

PR #934, opened 2022-02-03
Omit GCL computation from non-hex blocks

psakievich:
Omit registering GCL mesh velocity algorithms on any non-hex parts.  This will allow us to mesh the hub with non hex elements since there is no deformation.  

Adding as a draft since I haven't tested it yet and I think we should add some checks/flags for the input deck to specifically enable this capability

### DIFF
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -1001,13 +1001,13 @@ Realm::setup_interior_algorithms()
   if (has_mesh_deformation()) {
     const AlgorithmType algType = INTERIOR;
     stk::mesh::PartVector mmPartVec = meshMotionAlg_->get_partvec();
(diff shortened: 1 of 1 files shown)


ashesh2512 (approved):

rcknaus (approved):
Looks good.  Probably should emit a statement to the log file saying we skipped block x because y

gantech (approved):

psakievich:
@rcknaus I'm looking at this again trying to clean things up.  Do you have any suggestions on how to pull the part name off the part vector?

rcknaus:
Should just be able to do `part->name()` for each part in the vector that's skipped and create a message from that.

PR #935, opened 2022-02-04
Couple baseline baseline SST k-omega with Gamma transport equation for transition modeling

wilsonrvornlgov:
This pull request is for the incremental implementation of the Gamma correlation-based transition model of Menter (2015) into the Nalu-Wind flow solver. The commits in this pull request couple the solution of the Gamma transport equation with the baseline k-omega turbulence model allowing capability for turbulence transition predictions. The coupling and transition capability is achieved by modifying the source terms of the k and omega transport equations. Transition modeling is triggered by adding 'transition_model: true' as a sub option to 'solution_options:' in the input file.

Transition predictions require k and omega to be set at the inflow consistent with freestream conditions of the experimental data to be compared to (see Menter, 2015).

Test cases for transition prediction have been added to the reg_tests directory for the ERCOFTAC T3A flat plate and the Eppler airfoil geometries.

**Pull-request type:**
- [ ] Bug fix 
- [ ] Documentation update
- [x] Feature enhancement
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [x] GCC
    - [ ] LLVM/Clang
    - [x] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*


src/SpecificDissipationRateEquationSystem.C
```diff
@@ -323,6 +328,13 @@ SpecificDissipationRateEquationSystem::register_inflow_bc(
   std::vector<double> userSpec(1);
   userSpec[0] = sdr.specDissRate_;
 
+  std::printf("Open file for writing sdrFS\n");
```
jamelvin, line 331: Would this be better if it's written to the Nalu-Wind output file?

`NaluEnv::self().naluOutputP0() << "SDR EqnSys Inlet omega value = " <<  userSpec[0] << std::endl;`

src/TurbKineticEnergyEquationSystem.C
```diff
@@ -382,6 +390,13 @@ TurbKineticEnergyEquationSystem::register_inflow_bc(
   std::vector<double> userSpec(1);
   userSpec[0] = tke.turbKinEnergy_;
 
+  std::printf("Open file for writing tkeFS\n");
```
jamelvin, line 393: Same with this one

src/node_kernels/BLTGammaM2015NodeKernel.C
```diff
@@ -157,8 +179,8 @@ BLTGammaM2015NodeKernel::execute(
   fonset1 = Rev / 2.2 / Re0c;
   fonset2 = stk::math::min(fonset1, 2.0);
   rt = density*tke/sdr/visc;
-  fonset3 = stk::math::max(1.0 - 0.0233236151603499 * rt * rt * rt, 0.0);
-  fonset = stk::math::max(fonset2 - fonset3, 0.0);
+  fonset3 = BLTmax(0.0, 1.0 - 0.0233236151603499 * rt * rt * rt);
```
jamelvin, line 182: There are a lot of constants in here, do they need to be abstracted to SolutionOptions?

src/TurbKineticEnergyEquationSystem.C
wilsonrvornlgov: Those 2 write statements are a bit of a hack to get the inflow values for k & omega down to the source terms for k & omega. So they are written to a file when the inflow_bc is registered (TurbKineticEnergyEquationSystem.C) and then the file is opened up and read in the node kernel (node_kernels/TKESSTBLTM2015NodeKernel.C). There is probably a more direct way to accomplish. I am open to suggestions.

src/node_kernels/BLTGammaM2015NodeKernel.C
wilsonrvornlgov: I assume you are referring to the use of constants like 2.2, 2.0, 0.0233236151603499, etc.? If so, is there an advantage to abstracting them in SolutionOptions? Those values aren't turbulence modeling constants in the sense of say sigma_k in the k-omega equations. They are presented as ordinary floating numbers in the Menter (2015) publication, as opposed to them presented using symbols. I am open to your suggestions.

jamelvin: Yeah, they probably don't belong in `SolutionOptions` then.  You could move them to local variables and put them in the constructor/setup or in the header file, but this is also probably fine.

Let's see if anyone else wants to weigh in on this.

src/TurbKineticEnergyEquationSystem.C
jamelvin: I don't see any easy way to access the inflow boundary options in the nodeKernel, but perhaps this is better suited as a `turbulence_model_constant`?  I would think you could set the value when you register the inflow bdry so as not to require the user to enter these quantities in two places.  I think that is probably safer than trying to write and then open/read a file in the working directory.  @gantech @psakievich @marchdf let us know if you have any comments on if this is necessary/reasonable approach?

wilsonrvornlgov:
Screen Shot 2022-02-09 at 9 37 15 AM
Screen Shot 2022-02-09 at 9 36 59 AM

@jamelvin @psakievich The floating point numbers are from algebraic operations presented in Menter (2015). For example, the 0.0233236151603499 comes from Eq (4) for F_onset3 (see attached fig) where (1/3.5)^3 = 0.0233236151603499. In Eq (7) for TuL, 100*sqrt(2/3) = 81.6496580927726. If will define those constants with **constexpr double** and put them in the setup.

src/node_kernels/BLTGammaM2015NodeKernel.C
wilsonrvornlgov: @jamelvin @psakievich Please see my reply below. I will define this numbers using **constexpr double** in the setup.

psakievich:
@wilsonrvornlgov thanks for being flexible.  It may seem a little pedantic but it will help us remember what is going on with this model in the future to see the variable names instead of the raw doubles.

wilsonrvornlgov:
> @wilsonrvornlgov thanks for being flexible. It may seem a little pedantic but it will help us remember what is going on with this model in the future to see the variable names instead of the raw doubles.

@psakievich I added the following to my header file private: section (BLTGammaM2015NodeKernel.h) to define those two constants:

  
constexpr double fonset3_const_;
constexpr double TuL_const_;

I got a compile error: 
================
In file included from /home/wilsonrv/exawind/source/nalu-wind/src/node_kernels/BLTGammaM2015NodeKernel.C(8):
/home/wilsonrv/exawind/source/nalu-wind/include/node_kernels/BLTGammaM2015NodeKernel.h(84): error: "constexpr" is not valid here

Should I have declared them differently?

jamelvin:
> > @wilsonrvornlgov thanks for being flexible. It may seem a little pedantic but it will help us remember what is going on with this model in the future to see the variable names instead of the raw doubles.
> 
> @psakievich I added the following to my header file private: section (BLTGammaM2015NodeKernel.h) to define those two constants:
> 
> constexpr double fonset3_const_; constexpr double TuL_const_;
> 
> # I got a compile error:
> In file included from /home/wilsonrv/exawind/source/nalu-wind/src/node_kernels/BLTGammaM2015NodeKernel.C(8): /home/wilsonrv/exawind/source/nalu-wind/include/node_kernels/BLTGammaM2015NodeKernel.h(84): error: "constexpr" is not valid here
> 
> Should I have declared them differently?

@wilsonrvornlgov I think you need a `static` in front, see: https://github.com/Exawind/nalu-wind/blob/e9142052b09d6a6ddecbe9b83dedd1f7b4588fac/include/ngp_algorithms/SSTAMSAveragesAlg.h#L66

psakievich:
@wilsonrvornlgov @jamelvin is right you want the `static` keyword since it is in the header.  This keeps it from getting redefined whenever the the header is included in another file.  You will also need to supply the value at the declaration like the example @jamelvin posted since we want it to be available at compile time.

wilsonrvornlgov:
> 

@psakievich I am assuming that I can't use **static constexpr double TuL_const_ = stk::math::pow(0.3333333333333333, 3.0)**; since it has a call to 'pow' and is not available at compile time?

Perhaps I can add a comment before the declaration in the header file?:

  //fonset3_const_ = stk::math::pow(1.0/3.5, 3.0) as defined in Menter (2015)
  static constexpr double fonset3_const_ = 0.0233236151603499;
  //TuL_const_ = 100.0 * stk::math::sqrt(2.0/3.0) as defined in Menter (2015)
  static constexpr double TuL_const_ = 81.6496580927726;

aprokop:
> @psakievich I am assuming that I can't use **static constexpr double TuL_const_ = 100.0 * stk::math::sqrt(2.0/3.0);** since it has a call to 'sqrt' and is not available at compile time?

Yes, `sqrt` is not constexpr in any of the C++ standards (https://en.cppreference.com/w/cpp/numeric/math/sqrt).

psakievich:
> //fonset3_const_ = stk::math::pow(1.0/3.5, 3.0) as defined in Menter (2015) static constexpr double fonset3_const_ = 0.0233236151603499; //TuL_const_ = 100.0 * stk::math::sqrt(2.0/3.0) as defined in Menter (2015) static constexpr double TuL_const_ = 81.6496580927726;
This is perfect.  It is fine to have the double precision number if other developers can tell where it came from at the decleration.

src/TurbKineticEnergyEquationSystem.C
wilsonrvornlgov: > >One other thought on this, it is feasible (probably not advisable) for a user to have two different inflows with two different tkeFS and sdrFS values. This looks like it would take last read as the value that gets passed everywhere. Is that the intent? Perhaps these free stream values shouldn't be on the bc data if there is only intended to be 1 true value for the entire domain.>>

@psakievich I was only thinking of the case where you have one principle free stream. Yes, you could have a situation where you have, for example, one free stream from the west and one from the south. In that case, you would need bookkeeping for two values of tkeFS and sdrFS. I will attempt to save off the free stream values and add them to the Realm object when the inflow BCs are initialized.

src/SpecificDissipationRateEquationSystem.C
wilsonrvornlgov: Transfer of freestream inflow values to  node_kernels via file write/read has been replaced by adding them to Realm class.

src/TurbKineticEnergyEquationSystem.C
wilsonrvornlgov: Transfer of freestream inflow values to  node_kernels via file write/read has been replaced by adding them to Realm class.

Transition model floats are now defined using constexpr.

wilsonrvornlgov:
@psakievich @jamelvin I believe I resolved these issues. Please let me know if I missed any of your questions.

src/SpecificDissipationRateEquationSystem.C
```diff
@@ -322,6 +327,7 @@ SpecificDissipationRateEquationSystem::register_inflow_bc(
   SpecDissRate sdr = userData.sdr_;
   std::vector<double> userSpec(1);
   userSpec[0] = sdr.specDissRate_;
+  realm_.sdrFS = userSpec[0];
```
psakievich, line 330: I'm still not sure we want this to live on the boundary condition.  Looking at the model it seems like this is a global parameter for the model. Pulling it off the bc user data could be problematic if a user supplies multiple inflow bc's and gives different values.  This will just overwrite the value with the last bc registered.  Since this is really a model constant for the transition model I think we should put it in `SolutionOptions::turbModelConstantMap_`.  I think there should be one entry point for the model 

If the model requires the inflow bc's to all match then we can do a check here to make sure it is within an epsilon of the specified turbulence model constant if the users specifiy it, and if they don't then we can just take the value specified in the `turbModelConstantMap_`.

psakievich (commented):
@wilsonrvornlgov let me know your thoughts on my suggestion about the sdr term. I can help you figure out the mechanics of requiring a parameter in the turbModelConstants if you'd like.

src/SpecificDissipationRateEquationSystem.C
wilsonrvornlgov: Sorry for the delay in responding to your comment. I had to put this aside due to another project deadline. 

The issue of turbulence inflow conditions is not that straightforward and we may need to have a brief Teams call to fully discuss. However, I will attempt to describe the issue below.

=====================

It is not clear that the present approach of fixing inflow values for k and omega over the region from inlet boundary to the leading edge (LE), can be easily generalized for multiple inflows. Recall, that the objective is to have freestream turbulence quantities of the calculation match those from the experiments at the leading edge of the body. This is difficult to accomplish in the calculation since the freestream turbulence will experience large decay from the inflow to the LE. The approach taken by others and myself here, is to fix the freestream values in a region from inflow to LE.

To explain the problem with multiple inflows, I attached a sketch of an airfoil calculation with two inflows (A & B), which have different inlet values for k & omega. Let's assume that we fixed the bookkeeping problem so that we can input and save conditions for inflows A & B separately. We need turbulence values set at Inflow A to convect to the no slip surface (Region A in the sketch) without being dissipated. Similarly, we need those from Inflow B to do likewise in Region B. The problem is that Regions A & B overlap, so it is not as simple as adding up source terms in Regions A and B.  The turbulence values are not additive in the overlap region.

IMG_0615
.

A more general approach to fix the longstanding problem of decay of the freestream turbulence for transition has recently been proposed in the last several months (Lopes, 2021, attached below). With this approach, the constants of the k-omega dissipation terms (beta and beta_star) are modified outside the boundary layer such that the freestream turbulence doesn't experience decay and target turbulence values at the LE can be achieved more naturally. Inside the boundary layer, those coefficients are not modified so that the original K-omega SST model remains intact.

[Lopes_Decay.pdf](https://github.com/Exawind/nalu-wind/files/8149087/Lopes_Decay.pdf)

I would recommend that the issue of simulations with multiple inflow boundaries be postponed until the approach of Lopes, 2021 or similar can be implemented. I wouldn't be able implement this approach now with my present 4 day a week schedule and other project demands.

Let me know if you have questions or feedback.

wilsonrvornlgov:
@psakievich @michaelasprague I have the final PR for transition modeling ready for submission, once we can close this PR and merge into the master.

src/SpecificDissipationRateEquationSystem.C
wilsonrvornlgov: Here is a link to the full PDF version of the Lopes PhD thesis:

https://www.researchgate.net/publication/353340282_Simulation_of_Transition_from_Laminar_to_Turbulent_Regime_in_Practical_Applications_of_Incompressible_Flow

Reference:
R. Lopes, Simulation of Transition from Laminar to Turbulent Regime in Practical Ap- plications of Incompressible Flow. PhD thesis, Universidade de Lisboa-Instituto Superior T ́ecnico, 2021.

wilsonrvornlgov: @psakievich I have added a check for multiple inflow BCs and pushed my changes. If you approve, could you merge and close this PR?

wilsonrvornlgov:
@psakievich Any chance we could get this closed soon?

psakievich (approved):

PR #937, opened 2022-02-23
UVM dependency removal

ldh4:
[WIP] Modifications to make Nalu-wind buildable without Cuda UVM

Pre-requisite work for converting Nalu-wind to be compatible with HIP.
This is to make Nalu wind to build without Cuda UVM

@tasmith4 @alanw0 @jhux2 @ddement

### DIFF
--- a/include/LinearSystem.h
+++ b/include/LinearSystem.h
@@ -70,6 +70,15 @@ public:
   
 };
 
+struct CoeffApplierDestructor
+{
+  using ptr_t = CoeffApplier;
+
+  void free(ptr_t* coeffApplier) {
+    kokkos_free_on_device(coeffApplier);
+  }
+};
+
 class LinearSystem
 {
 public:
@@ -114,6 +123,8 @@ public:
    */
   virtual void buildDirichletNodeGraph(const std::vector<stk::mesh::Entity>&) {}
   virtual void buildDirichletNodeGraph(const stk::mesh::NgpMesh::ConnectedNodes) {}
+
+  virtual void free_coeff_applier(CoeffApplier*) {}
 
   // Matrix Assembly
   virtual void zeroSystem()=0;
--- a/include/LinearSystem.h
+++ b/include/LinearSystem.h
@@ -75,7 +75,9 @@ struct CoeffApplierDestructor
   using ptr_t = CoeffApplier;
 
   void free(ptr_t* coeffApplier) {
-    kokkos_free_on_device(coeffApplier);
+    if(coeffApplier != nullptr) {
+      kokkos_free_on_device(coeffApplier);
+    }
   }
 };
 
--- a/include/SolverAlgorithm.h
+++ b/include/SolverAlgorithm.h
@@ -66,8 +66,6 @@ struct NGPApplyCoeff
   mutable stk::mesh::NgpField<double> diagField_;
   stk::mesh::NgpField<int> iblankField_;
 
-  // Teuchos::RCP<CoeffApplier> deviceSumInto_;
-  // CoeffApplier* deviceSumIntoPtr_;
   CoeffApplier* deviceSumInto_;
 
   const unsigned nDim_{3};
--- a/include/SolverAlgorithm.h
+++ b/include/SolverAlgorithm.h
@@ -14,6 +14,7 @@
 
 #include <Algorithm.h>
 #include <KokkosInterface.h>
+#include <LinearSystem.h>
 
 #include <stk_mesh/base/Entity.hpp>
 #include <stk_mesh/base/Ngp.hpp>
@@ -65,7 +66,8 @@ struct NGPApplyCoeff
   mutable stk::mesh::NgpField<double> diagField_;
   stk::mesh::NgpField<int> iblankField_;
 
-  CoeffApplier* deviceSumInto_;
+  Teuchos::RCP<CoeffApplier> deviceSumInto_;
+  CoeffApplier* deviceSumIntoPtr_;
 
   const unsigned nDim_{3};
   const bool hasOverset_{false};
--- a/include/SolverAlgorithm.h
+++ b/include/SolverAlgorithm.h
@@ -66,8 +66,9 @@ struct NGPApplyCoeff
   mutable stk::mesh::NgpField<double> diagField_;
   stk::mesh::NgpField<int> iblankField_;
 
-  Teuchos::RCP<CoeffApplier> deviceSumInto_;
-  CoeffApplier* deviceSumIntoPtr_;
+  // Teuchos::RCP<CoeffApplier> deviceSumInto_;
+  // CoeffApplier* deviceSumIntoPtr_;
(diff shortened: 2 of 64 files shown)


tasmith4:
After further investigation, @ldh4 and I learned from @alanw0 that tests without `NGP` in the name are not expected to pass in any build.  After running the unit tests with that filter, they all pass on ascicgpu both with and without UVM.

alanw0:
> After further investigation, @ldh4 and I learned from @alanw0 that tests without `NGP` in the name are not expected to pass in any build. After running the unit tests with that filter, they all pass on ascicgpu both with and without UVM.

@tasmith4 @ldh4 More precisely, tests without `NGP` in the name are not expected to pass in a Cuda build. All unit tests are expected to pass in a CPU build. However, I am surprised to hear that all of the NGP unit-tests pass on Cuda without UVM. For example, the `*.NGP*` filter would include at least 3 unit tests that have tpetra usage (MixtureFractionKernelHex8Mesh.NGP_adv_diff_edge_tpetra*). I guess it's possible those tests don't use the graph assembly code, but I would be surprised. (And the one thing I think we're pretty certain about is that the graph assembly code is not using the correct types for at least some of its Views.)

In any case, I'll work with you and Dong Hun tomorrow to help establish where we're at with all this.

tasmith4:
You're right @alanw0, I should have been more precise, thanks for clarifying.  I was also very surprised all the NGP tests pass without UVM (and am willing to be skeptical of the result, although I double-checked my build), we'll have to discuss this more today as you say.

alanw0:
@tasmith4 In particular we're passing those views (that this commit changes to host views) into the setAllIndices method on the tpetra crsgraph. I guess the explanation could be that we're getting those host views from a dual view, and inside setAllIndices tpetra is doing the sync for us and then using the device views. If that's what is happening then great!

tasmith4:
I don't think that's what's happening, so I'm now even more confused by why these tests pass.

alanw0:
I guess I didn't look closely... Now that I look at this commit again, it *doesn't* change the types of the views we're passing into setAllIndices. It is changing the types of other stuff. So we'll just have to work through this today and see what is going on. We'll probably want to put in a (temporary) check in the code, in TpetraLinearSystem.C, for whether uvm is actually on or not.

include/SolverAlgorithm.h
```diff
@@ -65,7 +66,8 @@ struct NGPApplyCoeff
   mutable stk::mesh::NgpField<double> diagField_;
   stk::mesh::NgpField<int> iblankField_;
 
-  CoeffApplier* deviceSumInto_;
+  Teuchos::RCP<CoeffApplier> deviceSumInto_;
```
PaulMullowney, line 69: This concerns me. I think we need a generic CoeffApplier definition.

src/FixPressureAtNodeAlgorithm.C
```diff
@@ -113,6 +113,8 @@ FixPressureAtNodeAlgorithm::execute()
       }
     });
   });
+
+  eqSystem_->linsys_->free_coeff_applier(deviceCoeffApplier);
```
PaulMullowney, line 117: Why is this necessary? Were we leaking memory?

src/SolverAlgorithm.C
```diff
@@ -73,6 +73,9 @@ NGPApplyCoeff::NGPApplyCoeff(EquationSystem* eqSystem)
   if (hasOverset_) {
     iblankField_ = nalu_ngp::get_ngp_field<int>(eqSystem->realm_.mesh_info(), "iblank");
   }
+
+  deviceSumInto_ = Teuchos::RCP<CoeffApplier>(eqSystem->linsys_->get_coeff_applier(), CoeffApplierDestructor(), true);
```
PaulMullowney, line 77: See comment above on generic CoeffApplier.

include/LinearSystem.h
```diff
@@ -70,6 +70,15 @@ class CoeffApplier
   
 };
 
+struct CoeffApplierDestructor
+{
+  using ptr_t = CoeffApplier;
+
+  void free(ptr_t* coeffApplier) {
+    kokkos_free_on_device(coeffApplier);
```
PaulMullowney, line 78: A safety check here might be helpful.

src/overset/AssembleOversetDecoupledAlgorithm.C
```diff
@@ -40,6 +40,8 @@ void AssembleOversetDecoupledAlgorithm::execute()
     fringeNodes.size(), KOKKOS_LAMBDA(const size_t& i) {
       coeffApplier->resetRows(1, &fringeNodes(i), 0, numDof, 1.0, 0.0);
     });
+
+  eqSystem_->linsys_->free_coeff_applier(coeffApplier);
```
PaulMullowney, line 44: We're doing this again. We must have been leaking memory?

include/SolverAlgorithm.h
ldh4: This has been reverted. It does look like this was the main culprit for the double-free issue we were encountering.

src/FixPressureAtNodeAlgorithm.C
ldh4: On the TpetraLinearSystem class, we had to change it so that it is no longer holding onto the host/device coeff appliers as its member variables. So, a new coeff applier is constructed on each request. Because of this, we had to introduce a way to properly deallocate coeff appliers when no longer needed. This is one of the changes that we will likely to revisit to find a better way to handle this.
Just as a note, `free_coeff_applier` is an empty function in the `LinearSystem.h`, so this line will not affect `HypreLinearSystem`.

src/SolverAlgorithm.C
ldh4: This line was also reverted.

src/overset/AssembleOversetDecoupledAlgorithm.C
ldh4: Same reason as the above.

include/LinearSystem.h
ldh4: Added.

ldh4:
A couple notes on this commit:
- This builds and runs with and without unified memory support (tested only with `cuda` backend for now).
- Building without unified memory currently shows deficiencies in both memory usage and runtime performance. We are still investigating the source of this and will have follow-on tasks to resolve this.
- Building with unified memory does not have this problem and maintains the pervious performance. When tested with `nrel5mw_refined_rcb` problem (~700M nodes) on 90 nodes of `summit`, the build with this patch ran to its completion where as the build from `Nalu-wind` `master` timed out.

tasmith4 (approved):

ldh4:
Checking numbers from `nrel5mw_refined_rcb` runs again with its termination count reduced by half showed that there isn't any significant difference in runtime and memory usage between this patched version (uvm turned on) and the master, as expected. Runtime had ~5% speedup.

PR #938, opened 2022-02-26
SST and AMS with SST for ABL with Coriolis

janiadcock:
**Pull-request type:**
- [ ] Bug fix 
- [x] Documentation update
- [x] Feature enhancement

Add ability to use SST to simulate the neutral ABL with the Coriolis effect. Previously SST was only accurate for the neutral ABL without the Coriolis effect. The key change was adding in a mixing length limiter to SST, which can be turned on/off via the input file.

Add ability to use AMS to simulate the neutral ABL with the Coriolis effect. The key changes were: (1) setting the RANS component to SST with a mixing length limiter, (2) using the mixing length limiter when calculating the RANS time scale used in AMS, (3) adding the input option to select strong symmetry boundary conditions through the abltop boundary condition (previously this was always set to weak symmetry), and (4) adding the input option to set AMS to RANS mode below the sand grain roughness height used in the SST for ABL bottom boundary condition. 

## Checklist

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [x] Intel compilers
    - [ ] NVIDIA CUDA
- [ ] Passes all unit tests

Haven't run on unit or regression tests.

*For new code updates*
- [x] Documentation updates - additions to user, theory, and verification manuals


docs/source/theory/boundaryConditions.rst
```diff
@@ -966,6 +966,16 @@ which can be written in general component form as,
 
 .. math:: F^n_i = F_j n_j n_i.
 
+Strong versus Weak Symmetry
+~~~~~~~~~~~~~~~
+
+There are two implementations of the symmetry boundary condition: strong and weak. In the strong symmetry implementation, the normal velocity is set to zero at the boundary. Strong symmetry has only been implemented for a cartesian mesh, meaning it can be used for flat but not curved surfaces. In the weak symmetry implementation, normal velocity can be nonzero. The Poisson solve tries to enforce mass conservation. Weak symmetry has not been implemented for Active Model Splits (AMS.)
```
marchdf, line 972:
```suggestion
There are two implementations of the symmetry boundary condition: strong and weak. In the strong symmetry implementation, the normal velocity is set to zero at the boundary. Strong symmetry has only been implemented for a cartesian mesh, meaning it can be used for flat but not curved surfaces. In the weak symmetry implementation, normal velocity can be nonzero. The Poisson solve tries to enforce mass conservation. Weak symmetry has not been implemented for Active Model Splits (AMS).
```

src/ShearStressTransportEquationSystem.C
```diff
@@ -607,6 +622,119 @@ ShearStressTransportEquationSystem::compute_f_one_blending()
   fOneBlend.modify_on_device();
 }
 
+/** Compute mixing length for 2003 SST implementation
+ */
+void
+ShearStressTransportEquationSystem::compute_lt()
```
marchdf, line 628: One thing I am worried about... Do you think we can make `lt` optional? I would assume that for a lot of SST runs we don't need to extra things. What do you think? I guess we've replaced `gamma` which used to be an in place calculation with a field.

src/ShearStressTransportEquationSystem.C
```diff
@@ -607,6 +622,119 @@ ShearStressTransportEquationSystem::compute_f_one_blending()
...
+
+  const stk::mesh::Selector sel =
+    (meta.locally_owned_part() | meta.globally_shared_part()) &
+    (stk::mesh::selectField(*gamma_)); 
+
+  nalu_ngp::run_entity_algorithm(
+    "SST::compute_gamma", ngpMesh, stk::topology::NODE_RANK, sel,
+    KOKKOS_LAMBDA(const MeshIndex& mi) {
+      const double f1 = fOneBlend.get(mi, 0); 
+      const double l_t = lt.get(mi, 0);
```
marchdf, line 708: I think you can move `l_t` into the `if (lengthScaleLimiter)` block

janiadcock: Good catch; changed this.

include/Enums.h
```diff
@@ -298,6 +299,7 @@ static const std::string TurbulenceModelConstantNames[] = {
   "ams_peclet_offset",
   "ams_peclet_slope",
   "ams_peclet_scale",
+  "avgTimeScaleCoeff",
```
jamelvin, line 302: Let's use `avgTimeCoeff` instead to be consistent with my other implementation

reg_tests/test_files/ablNeutralEdgeAMS/ablNeutralEdgeAMS.yaml
```diff
@@ -0,0 +1,332 @@
...
+       - avg_res_adequacy_parameter
+       - resolution_adequacy_parameter
+       - metric_tensor
+       - element_courant
+       - average_production
+       - average_dudx
+       - forcing_components
+       - sst_f_one_blending
+       - l_t
+       - gamma
```
jamelvin, line 288: let's remove `l_t` and `gamma`

marchdf (approved):

src/ngp_algorithms/SSTAMSAveragesAlg.C
```diff
@@ -103,23 +109,34 @@ SSTAMSAveragesAlg::execute()
...
       if (tke.get(mi, 0) == 0.0)
         beta.get(mi, 0) = 1.0;
+      else if ((RANSBelowKs) && (coords.get(mi,2) <= k_s)) {
+        beta.get(mi, 0) = 1.0;
+      }
       else {
         beta.get(mi, 0) =
           (tke.get(mi, 0) - avgTkeRes.get(mi, 0)) / tke.get(mi, 0);
-
+
```
marchdf, line 139: remove spurious tab

docs/source/theory/boundaryConditions.rst
```diff
@@ -966,6 +966,16 @@ which can be written in general component form as,
 
 .. math:: F^n_i = F_j n_j n_i.
 
+Strong versus Weak Symmetry
+~~~~~~~~~~~~~~~
+
+There are two implementations of the symmetry boundary condition: strong and weak. In the strong symmetry implementation, the normal velocity is set to zero at the boundary. Strong symmetry has only been implemented for a cartesian mesh, meaning it can be used for flat but not curved surfaces. In the weak symmetry implementation, normal velocity can be nonzero. The Poisson solve tries to enforce mass conservation. Weak symmetry has not been implemented for Active Model Splits (AMS).
```
psakievich, line 972:
```suggestion
There are two implementations of the symmetry boundary condition: strong and weak. In the strong symmetry implementation, the normal velocity is set to zero at the boundary. Strong symmetry has only been implemented for a cartesian mesh, meaning it can be used for flat but not curved surfaces that are aligned with the principle Cartesian directions. In the weak symmetry implementation, normal velocity can be non-zero since the cell averaged quantities in Nalu-Wind's discretization are stored at the nodes and therefore reside on the boundaries. The Poisson solve enforces mass conservation globally to the order of linear solver convergence. The weak symmetry boundary condition will converge with with a first order rate under additional refinement.  One can enforce a stronger representation of the boundary by amplifying the penalty factor used in the weak boundary condition: `symmetry_bc_penalty_factor`. Its default value is 2.0 which is the minimum required for stability.  Amplifying the penalty factor will enforce the boundary in a stronger sense, but can also lead to a more difficult matrix solve as with any large penalty term.

It should be noted that in either boundary condition representation (strong or weak) there is an associated error from the boundary condition.  In the weak form the error manifests itself in non-zero local velocities at the boundary, and in the strong form the associated error is moved off the boundary and into the domain. Weak symmetry has not been implemented for Active Model Splits (AMS).
```
The description I've added above is a more precise description of the symmetry bc options in nalu-wind.  There are a few things going on with the bc in terms of the control volume, the Poisson solve's solver tolerances and the order of accuracy of the BC.
Also, what do you mean by the following?
> Weak symmetry has not been implemented for Active Model Splits (AMS)

Is there an equation that needs an additional symmetry implementation, or is it that it hasn't been tested? Theoretically either boundary condition should work correct?

src/NaluParsing.C
```diff
@@ -229,6 +229,30 @@ namespace sierra
       abltopBC.theBcType_ = ABLTOP_BC;
       const YAML::Node& abltopUserData = node["abltop_user_data"];
       abltopBC.userData_ = abltopUserData.as<ABLTopUserData>();
+      if (node["symmetry_user_data"]) {
```
psakievich, line 232: This looks like it was copied from the symmetry bc. Consider extracting this as a function so we don't have to maintain the same syntax in two places?

src/ngp_algorithms/SSTAMSAveragesAlg.C
```diff
@@ -103,23 +109,34 @@ SSTAMSAveragesAlg::execute()
...
+    k_s = 30.*z0_;
+  }
 
   nalu_ngp::run_entity_algorithm(
     "SSTAMSAveragesAlg_computeAverages", ngpMesh, stk::topology::NODE_RANK,
     sel, KOKKOS_LAMBDA(const Traits::MeshIndex& mi) {
       // Calculate alpha
       if (tke.get(mi, 0) == 0.0)
         beta.get(mi, 0) = 1.0;
+      else if ((RANSBelowKs) && (coords.get(mi,2) <= k_s)) {
```
psakievich, line 133: Another hard coded z.

src/node_kernels/SDRSSTAMSNodeKernel.C
```diff
@@ -59,9 +59,18 @@ SDRSSTAMSNodeKernel::setup(Realm& realm)
...
   betaOne_ = realm.get_turb_model_constant(TM_betaOne);
   betaTwo_ = realm.get_turb_model_constant(TM_betaTwo);
-  gammaOne_ = realm.get_turb_model_constant(TM_gammaOne);
-  gammaTwo_ = realm.get_turb_model_constant(TM_gammaTwo);
   tkeProdLimitRatio_ = realm.get_turb_model_constant(TM_tkeProdLimitRatio);
+
+  lengthScaleLimiter_ = realm.solutionOptions_->lengthScaleLimiter_;
+  if (lengthScaleLimiter_) {
+    const NodeKernelTraits::DblType earthAngularVelocity = realm.solutionOptions_->earthAngularVelocity_;
+    const NodeKernelTraits::DblType pi = std::acos(-1.0);
```
psakievich, line 67: you can also use `M_PI` which is hardcoded in `math.h` to replace creating Pi variables.  Not requesting a change. Merely an FYI.

src/node_kernels/SDRSSTAMSNodeKernel.C
```diff
@@ -90,10 +99,36 @@ SDRSSTAMSNodeKernel::execute(
...
 
+  NodeKernelTraits::DblType gammaOne_apply;
+  NodeKernelTraits::DblType gammaTwo_apply;
+  // apply limiter to gamma
+  if (lengthScaleLimiter_) {
+    // calculate mixing length
+    const NodeKernelTraits::DblType l_t = stk::math::sqrt(tke)/(stk::math::pow(betaStar_, .25)*sdr);
+ 
+    // calculate maximum mixing length
+    const NodeKernelTraits::DblType l_e = .00027*referenceVelocity_/corfac_;
```
psakievich, line 112: Could you add a comment on where `0.00027` comes from?

psakievich (changes requested):
I would like changes to the documentation regarding the symmetry bc.  I've added the text there.  Also I would like to consider @jamelvin 's point about hard coding z for ABL's. If you don't want to make a change in this PR then we should at least have a check to throw if someone tries to run in another configuration.  Sorry it took so long for me to review.

src/node_kernels/MomentumSSTAMSForcingNodeKernel.C
```diff
@@ -35,7 +35,8 @@ MomentumSSTAMSForcingNodeKernel::MomentumSSTAMSForcingNodeKernel(
       solnOpts.get_turb_model_constant(TM_periodicForcingLengthY)),
     periodicForcingLengthZ_(
       solnOpts.get_turb_model_constant(TM_periodicForcingLengthZ)),
-    nDim_(bulk.mesh_meta_data().spatial_dimension())
+    nDim_(bulk.mesh_meta_data().spatial_dimension()),
+    kappa_(solnOpts.get_turb_model_constant(TM_kappa))
```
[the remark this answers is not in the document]
psakievich: I didn't see it either.

src/ShearStressTransportEquationSystem.C
janiadcock: Sounds good. I switched `gamma` and `lt` back to in place.

reg_tests/test_files/ablNeutralEdgeSST/ablNeutralEdgeSST.yaml
```diff
@@ -0,0 +1,304 @@
...
+            aOne: 0.31
+            betaOne: 0.0236
+            betaTwo: 0.0276
+            gammaOne: 0.3255
+            gammaTwo: 0.3011
+            sigmaKOne: 0.85
+            sigmaKTwo: 1.0
+            sigmaWOne: 0.5
+            sigmaWTwo: 0.67
+            SDRWallFactor: 10.0 # for timestepping not turbulence model
```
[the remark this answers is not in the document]
janiadcock: Thanks for catching that. Also fixed it in aslNeutralEdgeSST/aslNeutralEdgeSST.

src/node_kernels/MomentumSSTAMSForcingNodeKernel.C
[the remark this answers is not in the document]
janiadcock: Removed it here and in the include file.

include/Enums.h
janiadcock: Done.

reg_tests/test_files/ablNeutralEdgeAMS/ablNeutralEdgeAMS.yaml
janiadcock: Removed them here and in `ablNeutralEdgeSST`.

docs/source/theory/boundaryConditions.rst
janiadcock: Thanks for adding this detail. When I was looking at using weak and strong symmetry myself I'd actually been wishing I could find this kind of information. :) 

Some clarifying questions:

When you write "In the weak symmetry implementation, normal velocity can be non-zero since the cell averaged quantities in Nalu-Wind's discretization are stored at the nodes and therefore reside on the boundaries." Is this implying that in strong symmetry cell average quantities are not stored at the nodes? Or is saying that it is possible to have a non-zero normal velocity in general, because the cell average quantities are not stored at the nodes?

What does "The weak symmetry boundary condition will converge with a first order rate under additional refinement" mean? What is converging and what is being refined?

What is the goal of the penalty factor? Aka what does it penalize in the weak boundary condition? 

Thanks!

src/NaluParsing.C
janiadcock: Do you have a suggestion for how to do this? I tried to avoid this by using e.g. `get_if_present_no_default` or `get_if_present` but kept running into errors, which Jeremy, Marc, and I couldn't get around.

docs/source/theory/boundaryConditions.rst
psakievich: @janiadcock I'm glad you are adding this. We didn't do a good job of documenting these changes. Partially because the results were never a slam dunk success as you may have noticed :) 

> When you write "In the weak symmetry implementation, normal velocity can be non-zero since the cell averaged quantities in Nalu-Wind's discretization are stored at the nodes and therefore reside on the boundaries." Is this implying that in strong symmetry cell average quantities are not stored at the nodes? Or is saying that it is possible to have a non-zero normal velocity in general, because the cell average quantities are not stored at the nodes?

This is due to the discretization itself. It is inconsistent at the boundary.  The nodes are always storing the control volume average quantities (finite volume formulation) and the standard control volume is cut in half at the boundaries ([picture](https://nalu-wind.readthedocs.io/en/latest/source/theory/discretizationApproach.html#id7)).  For the strong symmetry we are just setting the volume averaged quantity to zero through a Dirichlet bc.  This is also related to your second question.

> What does "The weak symmetry boundary condition will converge with a first order rate under additional refinement" mean? What is converging and what is being refined?

Under grid refinement the weak implementation will become the strong implementation, i.e. zero normal velocity every where.  This is due to control volume shrinking which does a couple of things: 1) it reduces the gradient being captured by the cell (standard grid refinement convergence effects) 2) it reduces the error from having the control volume averaged quantity not reside at the centroid of the control volume. The second item is a first order error.

> What is the goal of the penalty factor? Aka what does it penalize in the weak boundary condition?

So the penalty is on the normal velocity and is scaled by a viscous term.  The weak form of the boundary condition is formulated as a removal of tangential stress ([documentation](https://nalu-wind.readthedocs.io/en/latest/source/theory/boundaryConditions.html#id13)).  
The penalty implementation can be found [here](https://github.com/Exawind/nalu-wind/blob/9e9da505939eda42f4f2de153a61cd6bc2456cca/src/edge_kernels/MomentumSymmetryEdgeKernel.C#L148-L162).
and goes to zero when the normal velocity is zero.  There are some nuances to this such as: what is the normal at a node on a curved surface? There is also effects from the pressure stabilization that are not accounted for in the penalty term's derivation.   @rcknaus derived it so he would be a good authority for any deeper questions. Over all we found that the penalty can help in some situations, but it's not a silver bullet by any means.

src/NaluParsing.C
psakievich: Something like this?

```c++
// function somewhere else:
SymmetryUserData get_symmetry_data(const YAML::Node& node)
{
        const YAML::Node& symmetryUserData = node["symmetry_user_data"];
        auto data = symmetryUserData.as<SymmetryUserData>();
        if (symmetryUserData["symmetry_type"]) {
          const std::string symmType = symmetryUserData["symmetry_type"].as<std::string>();
          if(symmType == "generalized_weak" || symmType == "default"){
            // do nothing already set, but keep for parse checking
          }
          else if(symmType == "x_direction_strong"){
            data_.symmType_ = SymmetryUserData::SymmetryTypes::X_DIR_STRONG;
          }
          else if(symmType == "y_direction_strong"){
            data_.symmType_ = SymmetryUserData::SymmetryTypes::Y_DIR_STRONG;
          }
          else if(symmType == "z_direction_strong"){
            data_.symmType_ = SymmetryUserData::SymmetryTypes::Z_DIR_STRONG;
          }
          else{
            throw std::runtime_error(
              "Unrecognized value for symmetry_type: " + symmType
              );
          }
        } 
  return data;
} 

// your code now
      if (node["symmetry_user_data"]) {
         abltopBC.symmetryUserData_ = get_symmetry_data(node);
      }      
```

src/node_kernels/SDRSSTAMSNodeKernel.C
janiadcock: Thanks.

src/node_kernels/SDRSSTAMSNodeKernel.C
janiadcock: Yes.

src/NaluParsing.C
janiadcock: Ah, okay. I misunderstood. I thought you were objecting to me copying `node["symmetry_user_data"]` instead of using a function like `get_if_present_no_default`.

I made the suggested change.

docs/source/theory/boundaryConditions.rst
```diff
@@ -943,38 +943,44 @@ field values are used for property evaluations. When flow is leaving the
...
 advection operator.
 
-.. _theory_symmetry_bc:
+.. _theory_strong_symmetry_bc:
 
-Symmetry Boundary Condition
+Strong Symmetry Boundary Condition
+++++++++++++++++++++++++++++++++++
+
+There are two implementations of the symmetry boundary condition: strong and weak. In the strong symmetry implementation, the normal velocity is set to zero at the boundary. Strong symmetry has only been implemented for a cartesian mesh, meaning it can be used for flat surfaces that are aligned with the principle Cartesian directions. It cannot be used for curved surfaces or flat surfaces that are not aligned with the principle Cartesian directions. Both the strong and weak symmetry boundary conditions have an associated error. In the strong form the associated error lies not on the boundary but in the domain.
```
psakievich, line 951:
```suggestion
There are two implementations of the symmetry boundary condition: strong and weak. In the strong symmetry implementation, the normal velocity is set to zero at the boundary. Strong symmetry has only been implemented for a Cartesian mesh, meaning it can be used for flat surfaces that are aligned with the principle Cartesian directions. It cannot be used for curved surfaces or flat surfaces that are not aligned with the principle Cartesian directions. Both the strong and weak symmetry boundary conditions have an associated error. In the strong form the associated error lies not on the boundary but in the domain.
```

docs/source/theory/boundaryConditions.rst
```diff
@@ -943,38 +943,44 @@ field values are used for property evaluations. When flow is leaving the
...
 
 which can be written in general component form as,
 
 .. math:: F^n_i = F_j n_j n_i.
 
-Strong versus Weak Symmetry
-~~~~~~~~~~~~~~~
+The momentum equation also penalizes non-zero normal velocity. The strength of this penality depends on the penalty factor. One can enforce a stronger representation of the boundary by amplifying the penalty factor through the `symmetry_bc_penalty_factor` variable in `solution_options`. Its default value is 2.0 which is the minimum required for stability. Amplifying the penalty factor will enforce the boundary in a stronger sense, but can also lead to a more difficult matrix solve as with any large penalty term.
+
+Both strong and weak symmetry boundary conditions have associated error. In the weak form the error manifests as non-zero local velocities at the boundary. As the mesh is refined, the weak symmetry boundary condition converges to the strong symmetry boundary condition (zero normal velocity) with a first order rate. Note that this doesn't mean the error goes away; the error just moves off the boundary and into the domain.
```
psakievich, line 976:
```suggestion
Both strong and weak symmetry boundary conditions have associated error. In the weak form the error manifests as non-zero local velocities at the boundary. As the mesh is refined, the weak symmetry boundary condition converges to the strong symmetry boundary condition (zero normal velocity) with a first order rate. 
```
The error _does_ go away under mesh refinement for both the strong and weak formulations of the boundary condition, but it only decreases with first order rate.  You can eventually get to DNS.  The main point in commenting on error is that switching from weak to strong does not make the boundary representation's error disappear.

psakievich (commented):
Looking good. Couple more nit-picks on the symmetry description, but I also want to give an overall thank you to you for making these much needed improvements to the documentation. 👏

docs/source/theory/boundaryConditions.rst
janiadcock: Made this change here and also for next "Cartesian".

docs/source/theory/boundaryConditions.rst
janiadcock: Changed the last line to include the additional info. from your description here (rather than removing the last line.)

marchdf:
Hi all, just checking in to see if anything else is missing or if we are ready to merge?

psakievich (approved):

janiadcock:
@marchdf I've implemented all the changes but need to process one set of runs to make sure the changes for not hardcoding the z-direction worked right. I also would like to get the reg tests to run. Once Eagle is back online I'll do those two things and then I think we'll be ready to merge.

src/ngp_algorithms/SSTAMSAveragesAlg.C
```diff
@@ -349,6 +372,9 @@ SSTAMSAveragesAlg::execute()
       // Handle case where tke = 0, should only occur at a wall boundary
       if (tke.get(mi, 0) == 0.0)
         resAdeq.get(mi, 0) = 1.0;
+      else if ((RANSBelowKs) && (coords.get(mi,2) <= k_s)) {
```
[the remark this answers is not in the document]
janiadcock: Changed it to no longer hard code.

janiadcock:
I've made all the requested changes. I haven't been able to get reg tests to run though--trying to figure out why.

marchdf (approved):

Issue #929, opened 2022-01-20
Nalu-Wind seg fault on Summit regression tests


PaulMullowney:
Nalu-Wind reg tests are segfaulting on Summit. The output below is for a RELEASE build. RELWITHDEBINFO builds show other issues. I think we should solve this one first as it happens during mesh initialization.

I'm using gcc 9.3.0 and cuda 11.0.3
   Nalu-Wind SHA: 3343b72905
   Trilinos SHA :      dd53a1e0fbe

In the build/reg_tests/airfoilRANSEdgeNGPTrilinos.rst directory, I generate a valgrind callstack via:

export CUDA_LAUNCH_BLOCKING=1 && jsrun --smpiargs="-gpu" -n 1 -a 1 -c 1 -g 1 valgrind --tool=memcheck --leak-check=full --log-file=trilinos_vg%p ../../../naluX  -I ../../../../reg_tests/test_files/airfoilRANSEdgeNGPTrilinos.rst/airfoilRANSEdgeNGPTrilinos.rst.yaml -o airfoilRANSEdgeNGPTrilinos.rst.log

I've attached the log file and vg output. Here's the relevant call stack (without line numbers).

==3068855== Jump to the invalid address stated on the next line
==3068855==    at 0x0: ???
==3068855==    by 0x1110C093: int sierra::nalu::nodes_per_entity<sierra::nalu::ElemDataRequestsGPU>(sierra::nalu::ElemDataRequestsGPU const&, sierra::nalu::METype) [clone .constprop.0] (in /gpfs/alpine/cfd116/scratch/mullowne/nalu-wind/build_gpu_tr_develop_release_hypre_master/naluX)
==3068855==    by 0x1112BD87: void sierra::nalu::nalu_ngp::run_face_elem_algorithm<stk::mesh::DeviceMesh, sierra::nalu::nalu_ngp::FieldManager, sierra::nalu::ElemDataRequests, __nv_hdl_wrapper_t<false, false, __nv_dl_tag<void (sierra::nalu::WallFuncGeometryAlg<sierra::nalu::AlgTraitsFaceElem<sierra::nalu::AlgTraitsEdge_2D, sierra::nalu::AlgTraitsQuad4_2D> >::*)(), &sierra::nalu::WallFuncGeometryAlg<sierra::nalu::AlgTraitsFaceElem<sierra::nalu::AlgTraitsEdge_2D, sierra::nalu::AlgTr\
[...]

PR #939, opened 2022-03-01
MasterElement nodes_per_entity Workaround

PaulMullowney:
ElemDataRequests is used to construct/initialize ElemDataRequestsGPU. A handoff of MasterElement pointers occurs during this construction. We used MasterElement pointers to the get nodes_per_entity value, depending on the MasterElement pointer type, through a Kokkos parallel_reduce. This call attempts to deference a null pointer when doing Release build types.

I find it a bit strange that we have this complex machinery for pulling a single value of GPU-class data back to the host.
The workaround is to just add data members and interfaces to ElemDataRequests that adds these data members, as host data, and initializes properly from AlgTraits, which stores the data for each element type.

**Pull-request type:**
- [x] Bug fix 
- [ ] Documentation update
- [ ] Feature enhancement

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [ ] Intel compilers
    - [x] NVIDIA CUDA
- [ ] Passes all unit tests

*For new code updates*

### DIFF
--- a/include/AssembleElemSolverAlgorithm.h
+++ b/include/AssembleElemSolverAlgorithm.h
@@ -67,7 +67,7 @@ public:
     const int bytes_per_team = 0;
     const int bytes_per_thread = calculate_shared_mem_bytes_per_thread(
       lhsSize, rhsSize_, scratchIdsSize, meta_data.spatial_dimension(),
-      dataNeededNGP, reqType);
+      nodesPerEntity_, dataNeededNGP, reqType);
 
     stk::mesh::Selector elemSelector = meta_data.locally_owned_part() &
                                        stk::mesh::selectUnion(partVec_) &
--- a/include/AssembleFaceElemSolverAlgorithm.h
+++ b/include/AssembleFaceElemSolverAlgorithm.h
@@ -65,7 +65,7 @@ public:
 
       const int bytes_per_team = 0;
       const int bytes_per_thread = calculate_shared_mem_bytes_per_thread(
-        lhsSize, rhsSize, scratchIdsSize, nDim, faceDataNGP, elemDataNGP);
+          lhsSize, rhsSize, scratchIdsSize, nDim, faceDataNeeded_.nodesPerElement_, elemDataNeeded_.nodesPerElement_, faceDataNGP, elemDataNGP);
 
       const auto nodesPerFace = nodesPerFace_;
       const auto nodesPerElem = nodesPerElem_;
--- a/include/ElemDataRequests.h
+++ b/include/ElemDataRequests.h
@@ -216,6 +216,11 @@ public:
   MasterElement *get_cvfem_surface_me() const {return meSCS_;}
   MasterElement *get_fem_volume_me() const {return meFEM_;}
 
+  void addNodesPerElement(const int nodesPerElement) {
+      nodesPerElement_=nodesPerElement;
+  }
+  int nodesPerElement_ = -1;
+
 private:
   const stk::mesh::MetaData& meta_;
   std::array<std::set<ELEM_DATA_NEEDED>, MAX_COORDS_TYPES> dataEnums;
--- a/include/ElemDataRequestsGPU.h
+++ b/include/ElemDataRequestsGPU.h
@@ -191,10 +191,10 @@ private:
   FieldInfoView fields;
   FieldInfoView::HostMirror hostFields;
 
-  MasterElement *meFC_;
-  MasterElement *meSCS_;
-  MasterElement *meSCV_;
-  MasterElement *meFEM_;
+  MasterElement *meFC_=nullptr;
+  MasterElement *meSCS_=nullptr;
+  MasterElement *meSCV_=nullptr;
+  MasterElement *meFEM_=nullptr;
 };
 
 } // namespace nalu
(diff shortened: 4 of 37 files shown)


PaulMullowney:
Ugh. I need to clean up the code in include/ngp_utils/NgpMEUtils.h

alanw0 (commented):
I don't know the master-element stuff as well as James @overfelt does.
This seems like a pretty extensive work-around. It wasn't possible to find out where/when the me pointer was null?
Does this work-around eliminate the need to have master-element on device? Or are we still using the device master-element for other things? (In which case we would still need to make sure it's not null.)

alanw0:
If you can tell James (or did you already?) where you saw the me pointer being null, he can no doubt figure out why and fix it. Did you catch it somewhere with a throw or a print or something? When we were trying to debug the nodes_per_entity thing, the me pointer was not null in there. That was the weird stuff where the seg-fault seemed to be happening for other unknown reasons.

PaulMullowney:
> If you can tell James (or did you already?) where you saw the me pointer being null, he can no doubt figure out why and fix it. Did you catch it somewhere with a throw or a print or something? When we were trying to debug the nodes_per_entity thing, the me pointer was not null in there. That was the weird stuff where the seg-fault seemed to be happening for other unknown reasons.

This is issue #929

PaulMullowney:
> > If you can tell James (or did you already?) where you saw the me pointer being null, he can no doubt figure out why and fix it. Did you catch it somewhere with a throw or a print or something? When we were trying to debug the nodes_per_entity thing, the me pointer was not null in there. That was the weird stuff where the seg-fault seemed to be happening for other unknown reasons.
> 
> This is issue #929

I think the me pointer seems ok. However it doesn't seem to like me->nodesPerElement_

sayerhs:
@alanw0 we need master element instance on the device to compute things like `SCS_AREAV` etc.

alanw0:
> I think the me pointer seems ok. However it doesn't see to like me->nodesPerElement_

I got confused by the statement in your commit message that there was an attempt to dereference a null pointer. If that were the case, we could debug that.
If James says this commit is ok, it's ok with me.

PaulMullowney:
> @alanw0 we need master element instance on the device to compute things like `SCS_AREAV` etc.

I'll stick by my original comment. Using a parallel_reduce to copy a class data member from device to host is not a good design choice, imo. The data should reside on both the host and the device since it's known at compile time through the AlgTraits.h file.

There is a problem somewhere in the MasterElement code or it's derived classes. I can't find it. In the meantime tests are crashing nightly so we don't know what other problems might be cropping in. This is a workaround that gets us running some tests.

alanw0:
@PaulMullowney I hear your frustration. I agree with the comment about the parallel_reduce. That info is indeed known on host at compile time. Of course nodes-per-element can vary across a hybrid mesh (mixed topology mesh). As long as nodes-per-element isn't being stored on a class that gets used across multiple topologies, this should be fine.
I agree it's very important to get nightly test coverage back up.

overfelt:
I think this is a work around that somehow avoids the problem just like the code has been changed in the past and the error just went away.   But I thought we had agreed that kicking the can down the road was the best way forward for now.   Unless someone has a better idea?  I don't.  In my testing all the pointers were valid, not a null pointer in sight.  Besides, there is a check for null pointer in the code.

jrood-nrel (approved):

overfelt (approved):

PaulMullowney:
> Before merging, there may be a simpler fix. The code is struggling with me->nodesPerElement_; in the parallel_reduce.
> 
> The base class master element has nodesPerElement_ as a public data member. https://github.com/Exawind/nalu-wind/blob/master/include/master_element/MasterElement.h#L304
> 
> However, a derived class has a shadowing variable of the same name that is private and is initialized at compile time: https://github.com/Exawind/nalu-wind/blob/master/include/master_element/Quad42DCVFEM.h#L291
> 
> Why not just initialize the base class variable at run time? There are a number of other variables that have the same pattern.

Actually, it looks like MasterElement class members are explicitly initialized in the constructors via MasterElement::nodesPerElement_ = nodesPerElement_. 
It still begs the question, which nodesPerElement_ is used in calls to me->nodesPerElement_? The base class or the derived.

alanw0:
That does sound like a potentially promising discovery. I would suggest that the derived-class variables should be deleted.

overfelt:
How about we merge this workaround because compile time dimensions are better than runtime.  Then add a nodes_per_entity() and num_integration_points() to ElemDataRequests and ElemDataRequestsGPU and get rid of NgpMEUtils.  It seems that ElemDataRequests has host side pointers to element classes only and ElemDataRequestsGPU has device side only.  So ElemDataRequests functions are called on host only and ElemDataRequestsGPU functions on GPU only?

PaulMullowney:
I tried removing all the instances of nodesPerElement_ (and some similar data) in the MasterElement derived classes. The code is back to seg faulting in nodes_per_entity().

alanw0:
> I tried removing all the shadow instances of nodesPerElement_ (and some similar data) in the MasterElement derived classes. The code is back to seg faulting in nodes_per_entity(). I think we should merge my workaround and see where this gets us.

Ugh. I guess go ahead and merge your work-around discussed above?

jrood-nrel:
I want to see if this will fix the tests tonight.

PR #958, opened 2022-03-21
Fix a wall BC field type in the SST EqSys

tasmith4:
Turns out the `wall_normal_distance_bip` field was declared in `LowMachEquationSystem` with type `GenericFieldType` and in `ShearStressTransportEquationSystem` with `ScalarFieldType`.  These two types are treated largely the same in practice by STK Mesh, but the differing template parameters cause issues for STK Mesh.  In simulations such as the one @janiadcock was running where both equation systems are used, the specified field types need to match in both `declare_field` calls inside `register_wall_bc`.

This PR assumes the `GenericFieldType` is more descriptive, since the field is clearly not scalar, but is being declared with length `numScsBip`.

FYI @janiadcock @psakievich @jrood-nrel @alanw0


psakievich (approved):

tasmith4:
@jrood-nrel I think this is ready to merge but don't have approval to merge apparently, can you take care of this?

PR #963, opened 2022-04-11
Adding SUST source terms to SST, default to off

marchdf:
**Pull-request type:**
- [ ] Bug fix 
- [x] Documentation update
- [x] Feature enhancement

This adds the capability to add the SUST source terms to SST (described at [the NASA LARC website](https://turbmodels.larc.nasa.gov/sst.html)). The user gets to pick two more turbulence parameters, tke_amb and sdr_amb, that set the freestream values. An appropriate source term is then added to the SST equations. The constants default to 0.0 to preserve the current behavior.

This capability is necessary for external aerodynamic flows to avoid the drastic decay of SDR and TKE from the inflow BC to the airfoil.

This was previously done in #944 but had to be reverted.

## Checklist

of reviewers and developers.

*All pull requests*
  - Operating systems 
    - [x] Linux
    - [ ] MacOS
  - Compilers 
    - [ ] GCC
    - [ ] LLVM/Clang
    - [x] Intel compilers
    - [ ] NVIDIA CUDA
- [x] Passes all unit tests

*For new code updates*
- [x] Documentation updates - additions to user, theory, and verification manuals


marchdf:
Hey @PaulMullowney, could I get your help in running the tests that you were running that were showing issues? I am still very confused because all this should be doing by default is doing `+0` on the RHS.

psakievich (approved):
I don't see any issues here. As soon as @PaulMullowney can confirm his case then I think we are good to merge. If there is an issue from this PR it is likely a larger bug in Nalu-Wind.

PR #972, opened 2022-05-24
stk::mesh::MeshBuilder Conversion

ddement:
stk::mesh::BulkData's constructor will be deprecated in favor of the stk::mesh::MeshBuilder factory object in the near future.  This change updates Nalu-Wind to use this new factory for all BulkData construction.

### DIFF
--- a/include/Realm.h
+++ b/include/Realm.h
@@ -414,8 +414,7 @@ class Realm {
   double l2Scaling_;
 
   // ioBroker, meta and bulk data
-  stk::mesh::MetaData *metaData_;
-  stk::mesh::BulkData *bulkData_;
+  std::shared_ptr<stk::mesh::BulkData> bulkData_;
   stk::io::StkMeshIoBroker *ioBroker_;
   std::unique_ptr<SideWriterContainer> sideWriters_;
 
--- a/src/InputOutputRealm.C
+++ b/src/InputOutputRealm.C
@@ -94,18 +94,18 @@ InputOutputRealm::register_io_fields() {
     // loop over target parts and declare/put the field
     for ( size_t j = 0; j < targetNames.size(); ++j ) {
       const std::string targetName = targetNames[j];
-      stk::mesh::Part *targetPart = metaData_->get_part(targetName);
+      stk::mesh::Part *targetPart = meta_data().get_part(targetName);
       if ( NULL == targetPart ) {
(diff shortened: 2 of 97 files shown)


jrood-nrel (approved):

psakievich (changes requested):
As a soft request I'd like to see moving from a raw pointer to a smart pointer in realm.  If that blows up into something big then we can back off it from this PR, but it looks pretty straightforward from what I see here.

As more of a question, @ddement is the long term goal for `stk::mesh::MeshBuilder` to unify `MetaData` and `BulkData`?  If so is there more we can do to accept that API in our infrastructure? I see hints of it throughout the unit test code.

alanw0:
@psakievich I don't think we can completely unify BulkData and MetaData. In Sierra we have a couple of scenarios where a MetaData persists and the associated BulkData gets destroyed and a new one gets constructed and attached to the persisting MetaData.
However, it is not necessary to separately hold a BulkData and a MetaData. If you have a BulkData, you can always get a MetaData from it.

psakievich:
@alanw0 it seems like it would make sense to stop carrying around meta and bulk data at the realm level in nalu-wind then, and just opt for a single `stk::mesh::MexhBuilder` or `BulkData` instead

ddement:
@psakievich I just pushed up a version that stores BulkData as a shared_ptr.  Because of the way it's set in the unit test HelperObjects, using a unique_ptr would be a bit more work.  We could do it using move semantics from where the BulkDatas are created in the unit tests, but it might require rethinking the logic flow for some of them.  shared_ptr should work for now.  

Also, I removed Realm's separate storage of a MetaData pointer.  That pointer is only ever accessed/set inside Realm itself or inside the HelperObjects, and it is only ever set right next to where BulkData is being set.  BulkData now manages the lifetime of MetaData.  Instead, the meta_data() method on Realm just calls through to BulkData::mesh_meta_data().

psakievich (approved):
@ddement this looks great. Thanks for making that change.

psakievich:
@tasmith4 let us know if the stk team doesn't have merge capabilities and who should have them from the stk team.

alanw0:
@psakievich @jrood-nrel I think it would be good if all of the stk folks are added, keeping in mind that we will still adhere to practices like waiting for approvals, and passing CI if it's working, etc.

psakievich:
@alanw0 not sure if you noticed on this PR but we have a bare bones CI running again thanks to the docker images from @eugeneswalker :)

Issue #973, opened 2022-05-27
Runtime error on Summit: cudaStreamSynchronize(stream) error( cudaErrorIllegalAddress): an illegal memory access was encountered


jhux2:
Using VOTD Trilinos and Nalu-Wind, the smaller NREL5MW simulation on Summit fails with this error:

```
terminatemSynchronize(stream) error( cudaErrorIllegalAddress): an illegal memory access was encountered called after throwing an instance of 'std::runtime_error'
  what():  cudaStreamSynchronize(stream) error( cudaErrorIllegalAddress): an illegal memory access was encountered /ccs/home/jjhu/exawind/sources/Trilinos-dev/packages/kokkos/core/src/Cuda/Kokkos_Cuda_Instance.cpp:162
```

jhux2:
@PaulMullowney

jhux2:
I was able to get a stacktrace:

<details>

```
#17 Kokkos::parallel_reduce<Kokkos::TeamPolicy<Kokkos::Cuda, Kokkos::Schedule<Kokkos::Dynamic> >, [snip]
    at /gpfs/alpine/scratch/jjhu/cfd116/exawind/install/trilinos-gcc-8.3.1-cuda-11.0.3-spectrum-mpi-10.4.0.3-relwithdebinfo-014b69b3/include/Kokkos_Parallel_Reduce.hpp:1848
[...]

PR #975, opened 2022-05-31
A small unit test that fails in Kokkos::parallel_reduce.

overfelt:
I think this test should not throw an exception but
reducing over a zero length loop leads to a memory
access error.

It is not clear if this is a nalu, stk or kokkos problem.

### DIFF
--- a/unit_tests/ngp_kernels/UnitTestNgpLoops.C
+++ b/unit_tests/ngp_kernels/UnitTestNgpLoops.C
@@ -944,3 +944,51 @@ TEST_F(NgpLoopTest, NGP_basic_face_elem_reduce)
 
   basic_face_elem_reduce(*bulk, *coordField, exposedAreaVec);
 }
+
+void basic_reduce_with_teams(const int num_mesh_buckets)
+{
+  using Mesh           = stk::mesh::NgpMesh;
+  using Traits         = sierra::nalu::nalu_ngp::NGPMeshTraits<Mesh>;
+  using TeamPolicy     = typename Traits::TeamPolicy;
+  using TeamHandleType = typename Traits::TeamHandleType;
+  using ReducerValueType = int;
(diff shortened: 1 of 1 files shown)


alanw0 (commented):
This looks like a good test. But I'm not sure if I'm correctly interpreting how you say it fails. Can it "expect" wrong behavior to be a passing test until it is fixed? I would hesitate to push in a crashing test...

overfelt:
@alanw0   No, this should not be merged, it is for example purposes only for discussion.  There is an expect-throw construct in gtest but then fixing the problem would cause the test to fail which is no good.  I managed to reduce the example to Kokkos only code and submitted a ticket to the Kokkos repository so it seems stk is off the hook.

alanw0:
@overfelt sounds good, thanks!

jhux2:
@overfelt Is this part of a larger Kokkos refactor?  The file `packages/kokkos/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp` doesn't seem to exist in Trilinos develop.  It would be nice to be able to do a "hot-fix" without having to wait for another Kokkos promotion.

overfelt:
The Kokkos pull request is to the develop branch of Kokkos which has a Kokkos_Cuda_Parallel_Team.hpp file for which the fix will be applied.  The Kokkos master branch does not have this file as yet.   But Kokkos_Cuda_Parallel_Team.hpp  was created from Kokkos_Cuda_Parallel.hpp and if you hunt through this file you can find the same lines of code as were patched in the pull request.  Not an ideal solution but I don't have a better suggestion.

jhux2:
@overfelt Thank you for tracking this down and submitting a reproducer to Kokkos!  The Kokkosk patch should help clean up the Nalu-Wind dashboard and address #973.

PR #987, opened 2022-06-28
Fix memory leak

ldh4:
Fixed a memory leak from NGPApplyCoeff that was introduced in #937.


tasmith4 (approved):

jrood-nrel (approved):

jrood-nrel:
Hmm, this looked like a fairly straightforward fix, but it's failing the unit tests.

ldh4:
@jrood-nrel Yeah, we were able to run the regression test that we were using to identify the memory leaks, but we did also just find out that this causes some failures in other tests. Looking into it right now.

jrood-nrel:
It looks like this has caused some new segfaults.

PR #990, opened 2022-07-05
Fixes to seg fault in coeff applier destructor

ldh4:
Fix to segfault caused by #987 memory leak fix patch

### DIFF
--- a/include/LinearSystem.h
+++ b/include/LinearSystem.h
@@ -137,6 +137,7 @@ public:
     return nullptr;
   }
 
+  virtual bool owns_coeff_applier() { return true; }
 
   virtual void sumInto(
     unsigned numEntities,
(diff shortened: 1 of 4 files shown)


PaulMullowney (commented):
Do the Hypre tests run again with this fix?

ldh4:
@PaulMullowney We ran regression tests that were populated using `nalu-wind+tioga+hypre+fftw` spec and saw that previously segfaulting tests introduced by #987, which included Hypre tests, are no longer segfaulting with this fix.

Currently I am checking to see if this change will reintroduce the memory leak that #987 fixed.

tasmith4 (approved):
@ldh4 you should finish your testing, but from a code perspective this looks good to me.

psakievich (approved):

PR #999, opened 2022-07-13
Implementing SUST for DES and IDDES

marchdf:
Co-authored-by: Jeremy Melvin <jamelvin@gmail.com>


alanw0:
If you add a test, either unit-test or regression-test, then it will help two things:
1. it will add confidence that your new/changed code is correct.
2. it will prevent other developers from accidentally breaking your code when they change other code, or when they change your code to perform routine maintenance, fix compiler warnings, whatever.

psakievich (approved):
Looks good. Thanks for adding the unit tests!

PR #1002, opened 2022-07-14
Refactor boundary conditions (again)

psakievich:
Fix remaining issues with #998

### DIFF
--- a/src/BoundaryConditions.C
+++ b/src/BoundaryConditions.C
@@ -7,9 +7,6 @@
 // for more details.
 //
 
-
-
-#include <Realm.h>
 #include <BoundaryConditions.h>
 #include <NaluEnv.h>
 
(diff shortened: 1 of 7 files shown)


rcknaus (approved):
Nice!

alanw0 (approved):
Looks like a nice cleanup.

src/BoundaryConditions.C
```diff
@@ -7,133 +7,117 @@
...
-//--------------------------------------------------------------------------
-//-------- load -----------------------------------------------
-//--------------------------------------------------------------------------
+// factory method to create any supported bc
+std::unique_ptr<BoundaryCondition>
+BoundaryConditionCreator::load_single_bc_node(const YAML::Node& node)
+{
+  std::unique_ptr<BoundaryCondition> this_bc;
+  if (node["wall_boundary_condition"]) {
+    this_bc = std::move(register_bc<WallBoundaryConditionData>(node));
```
rcknaus, line 37: Do you need the `move` here? register_bc returns an r-value, right?

jrood-nrel (approved):

src/BoundaryConditions.C
psakievich: No, right now it returns a `std::unique_ptr` but maybe it shouldn't

psakievich: Relooking at this now and this is what I intended.  The `unique_ptr` gets passed through a couple of function calls, but ends up in the storage vector at the end.  A little messy still I guess, but functional.

psakievich: Oh wait nvm. I think you are right. I'll update it

PR #1010, opened 2022-08-05
Fix fallthrough for sst-iddes source kernel / compiler warnings

rcknaus:
The SST-IDDES node kernel has an implicit fallthrough on the switch statement, activating both the sources (the KE source doesn't apply though and errors out).

Going through the compiler warnings, [`omegaHat` being unused](https://github.com/Exawind/nalu-wind/blob/master/src/ngp_algorithms/TurbViscKOAlg.C#L87) is probably unintentional and means the computed stress limiter isn't being used.  [`Dk` and the computation of `betaStarLowRe`](https://github.com/Exawind/nalu-wind/blob/master/src/node_kernels/SDRKONodeKernel.C#L116) is unused, probably also unintentionally although I know this stuff is new and being actively worked on.

### DIFF
--- a/src/HypreLinearSystem.C
+++ b/src/HypreLinearSystem.C
@@ -2400,7 +2400,6 @@ double
 HypreLinearSystem::copy_hypre_to_stk(stk::mesh::FieldBase* stkField)
 {
   auto& meta = realm_.meta_data();
-  auto& bulk = realm_.bulk_data();
   const auto selector =
     stk::mesh::selectField(*stkField) & meta.locally_owned_part() &
     !(stk::mesh::selectUnion(realm_.get_slave_part_vector())) &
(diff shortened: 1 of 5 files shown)


jamelvin:
Thanks for catching these @rcknaus!

@marchdf For `Dk`, https://github.com/Exawind/nalu-wind/blob/master/src/node_kernels/SDRKONodeKernel.C#L116, we're missing the limiter on `Pk` in this file

```  
// Clip production term and clip negative productions
  Pk = stk::math::min(tkeProdLimitRatio_ * Dk, stk::math::max(Pk, 0.0));
```
  
  `omegaHat` I think should be dropped for `KO`

jamelvin:
@rcknaus I added those changes for the compiler warnings (Pk and omegahat) as a PR on your branch

psakievich (approved):

marchdf (approved):

psakievich (approved):

PR #1018, opened 2022-08-13
FSI: Add interface class for turbines and/or actuators

psakievich:
General refactoring to allow for an abstraction for FSI+actuators that will interface with Realm.  Most of the changes were just moving files from the `[src/include]/actuators` folder to the `[src/include]/aero/actuators` folder.  

A new class is adde `AeroContainer` that will become the high level interface for interacting with FSI or actuator methods.


neilmatula (commented):
Looks good to me.

neilmatula (approved):

PR #1023, opened 2022-08-26
Suppress unit test calling openfast

psakievich:


tasmith4 (approved):
@psakievich I'm fine with this, but can you create a companion issue so we don't lose track that we need to circle back to this?

jrood-nrel:
Is it because we randomize the test order?

psakievich:
@jrood-nrel I don't think that is the issue.  I'm not sure yet. But I'm also tired of trying to make it work.  The test is not adding value equal to what it is costing us. Unfortunately whenever openfast gets run it creates a whole bunch of files.  I think there is a conflict in the files/and/or some read operation. It could also be an issue with the controller library dynamically linking with some overlap.

PR #1029, opened 2022-09-01
Fix Mem leak for ADM with FLLC

psakievich:
FLLC arrays weren't getting resized for the ADM
For a single turbine the offsets are still correct so thats why those
cases worked.

This is a quick and robust fix for the infrastructures, but it is quite
wasteful with memory.  Especially if there are a large number of ADM
points.

In general I think we are over due for a refactor of this feature to
drive memory consumption down by moving toward the implementation in
AMR-Wind.


lawrenceccheung (approved):
Thanks Phil, this looks good.  It works for the test case, and I'm testing it out on the production runs now.

Lawrence

Issue #1028, opened 2022-08-31
Clang warnings about overridden functions


tasmith4:
The nightly dashboards report several clang warnings about overridden functions in the master element code.  @overfelt these warnings began to appear after #1022 merged, although I'm a little confused about why that PR caused those warnings.  Can you take a look?

tasmith4:
https://sierra-cdash.sandia.gov/index.php?project=Nalu-Wind

https://my.cdash.org/index.php?project=Nalu-Wind

sayerhs:
@tasmith4 clang will complain if you add `override` to certain `virtual` functions that are overridden but not others. In this case, I suspect it is unhappy about `shape_fcn` and others that are overridden but don't have override directives. To remove warnings, you will either have to mark all overridden functions with `override` or remove `override` from the lines added in #1022

tasmith4:
Fixed by #1030.  Thanks @overfelt!

PR #1030, opened 2022-09-01
Removal of Fortran. Issue902

overfelt:
It is getting more difficult to work around calling the Fortran routines from the MasterElement class.  There are virtual functions on the MasterElement base class that take Kokkos views.  Since the functions are virtual, there are no template parameters for the memory space of the view.  For GPUs the views are on device memory.  But this means not calling these functions from host.  So, when needed on host a kernel is launched to fill the views on device and deep_copy back to host.  Another option would be to create a second virtual function for host memory views.  But a better fix would be to port all CPU only kernels to GPU enabled like all the others.  This is not needed for efficiency, but just to make the code consistent.


tasmith4:
@overfelt does this resolve #1028?

overfelt:
@tasmith4 Yes, this merge request should fix the warnings mentioned in #1028 .  I need to clean up and squash all the commits and get Alan to review before merging to master.

alanw0 (approved):
I'm not sure if I understand everything here, with some functions seemingly switching to templated, but if it builds and passes tests, it looks fine to me.

overfelt:
@alanw0 The tests do pass, but that was also true when I found errors in what I added.  I wonder about our test coverage.  

The templated functions are a thing I added to make the interface cleaner.  It would be really nice to have templated virtual functions but C++ does not support that.  So I added a non-virtual templated function in the base class, then template specializations for the cases I care about and support, calls from these specializations to non-templated virtual functions in protected space so they are hidden from the user interface, then the virtual overrides in the derived class call a templated function in the derived class public space.  From a calling program perspective it makes it appear as virtual templated functions.  The other option is to expose the protected virtual functions used and put them in the public space and call them directly. All of this will have to change if we decided to get rid of RDC.

PR #1031, opened 2022-09-03
Remove some now unused Fortran files.

overfelt:
There is one left, MasterElementWork.F, which will just have
to be rewritten in C++ but it is a lot shorter than it used
to be.


jrood-nrel:
Loving this.

Issue #902, opened 2021-11-09
Refactoring Discussion


psakievich:
It is a general consensus that the current design of `Realm` and `SolutionOptions` objects make this code more difficult to test and develop in.  This is because they are essentially [God-Objects](https://en.wikipedia.org/wiki/God_object).
This is only getting worse as we perpetuate the use of this style of global parameter passing on the Realm interface due to how difficult it will be to detangle the current mess.

I'd like to propose that we work to tackle some of these issues at a low but consistent effort over the next FY while we still have a good chunk of ECP funding.  I believe some concentrated efforts would really help make the code less intimidating, more sustainable, and easier to program in.

At the highest level I think we need to move toward a design where `Realm` is simply a container where the `EquationSystems` reside and not a place where parameters reside.  All the parameters and settings should be pushed closer to the actual code that is affected by parameters such as, do we have overset turned on. This will create encapsulation and allow us to create, and test new features quicker and with fewer bugs.

Some of the top priorities IMO should be:

- [ ] Remove all unmanaged object creation (new/delete pairs vs `std::unique_ptr` and `std::shared_ptr`)
- [ ] Remove all fortran code ( see #899 )
- [ ] Simplify field registration ( See #901 )
- [ ] Move toward factory creation of kernels and algorithms in Realm and the equation systems using `YAML::Nodes` as the factory inputs rather than creating SolutionOptions and dragging it everywhere

I think the first three items on this list will naturally create a path for the last one, but I also think they will take a lot of work. I'm not sure we will be able to get all this done, but I think we can move in that direction and improve things while we do.  I'm starting this issue to facilitate a discussion since we don't have github discussions enabled for this repo. 

tagging: @rcknaus @marchdf @alanw0 @jrood-nrel @overfelt @neilmatula @gantech @sayerhs @tasmith4 @PaulMullowney

PR #1032, opened 2022-09-06
Removal of fortran.  Issue #902

overfelt:
This removes the last of the Fortran source code from nalu-wind.


psakievich (approved):
Woohoo! Thanks for doing this @overfelt!

PR #1036, opened 2022-09-09
Move to cxx17

psakievich:
Posting as a draft PR for now.  Trying to get the dashboard a little greener

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@ option(ENABLE_OPENMP "Enable OpenMP flags" OFF)
 option(ENABLE_BOOST  "Enable Boost libraries" OFF)
 option(NALU_WIND_SAVE_GOLDS  "Save gold files to directory when running tests" OFF)
 
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_CXX_STANDARD_REQUIRED)
 


tasmith4:
thanks @psakievich.  yeah it might be overly idealistic, but I do think we're sort of close with the dashboard, and would be great to have that green before we flip the C++17 switch.

PaulMullowney:
I also might wait on getting the crusher compile to work. I don't want add in more chances for the hip compilers to struggle.

psakievich:
@PaulMullowney that is a good point. Perhaps it would be good to make this configurable if we run into issues there.

tasmith4 (approved):
I'm still hesitant, because I would feel better about this if the dashboard was absolutely green, and if vanilla spack-manager could build the latest nalu-wind@master directly on crusher.  On the other hand we're going to have to resolve these issues anyway soon enough.

psakievich:
@tasmith4 I think it is going to be a bit before we can have vanilla spack-manager working on crusher based on today's NGP meeting.  I don't want to wait that long. I'm fine waiting until next week for the dashboard to settle down, but I don't think this is going to impact our builds very much with the possible exception of ROCM/HIP. Nalu-Wind and Trilinos are both being pinned and periodically updated for that effort so this isn't really putting it at risk.

In terms of the dashboard being completely green, there are a couple of memory leaks that are going to take some effort, but we also know exactly what needs to be done.  It seems highly unlikely that c++17 is going to impact those issues at all, and they aren't super high priorities right now either.  Other than that we are green assuming `trilinos@develop` issues are resolved with today's rebless.  I'd rather not hold up moving the standard because of this small backlog of known, but time consuming issues. Also if there are any issues we want to give time for identification and feedback to trilinos/kokkos before kokkos 4.0 gets merged into trilinos.

tasmith4:
good points, yeah probably better just to go ahead

jrood-nrel (approved):

PR #1043, opened 2022-09-19
Bug fix Surface-Moment output

psakievich:
Setting an output frequency of 0 will lead to a floating point exception Adding a safe way to return if no output is desired.

### DIFF
--- a/src/SurfaceForceAndMomentAlgorithm.C
+++ b/src/SurfaceForceAndMomentAlgorithm.C
@@ -142,7 +142,7 @@ SurfaceForceAndMomentAlgorithm::execute()
     return;
   }
 
-  const bool processMe = (timeStepCount % frequency_) == 0 ? true : false;
+  const bool processMe = (timeStepCount % frequency_) == 0;
 
   // do not waste time here
   if (!processMe)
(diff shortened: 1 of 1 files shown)


src/SurfaceForceAndMomentAlgorithm.C
```diff
@@ -137,6 +137,11 @@ SurfaceForceAndMomentAlgorithm::execute()
 {
   // check to see if this is a valid step to process output file
   const int timeStepCount = realm_.get_time_step_count();
+
+  if (frequency_ < 1) {
+    return;
+  }
+
   const bool processMe = (timeStepCount % frequency_) == 0 ? true : false;
```
marchdf, line 145: can we fix the silly ternary expression while we are at it?

jrood-nrel: Agreed

psakievich: 😂 I was in such a rush I didn't even notice that.

psakievich, line 145:
```suggestion
  const bool processMe = (timeStepCount % frequency_) == 0;
```

jrood-nrel (approved):

PR #1045, opened 2022-09-28
FSI: Add low level utility functions

psakievich:
This PR will add low level functions for FSI that can live as utilities and be executed inline on device.

### DIFF
--- a/include/aero/AeroContainer.h
+++ b/include/aero/AeroContainer.h
@@ -17,7 +17,7 @@
 #include "aero/actuator/ActuatorModel.h"
 
 #ifdef NALU_USES_OPENFAST
-#include "OpenFAST.h"
+#include "OpenFAST.H"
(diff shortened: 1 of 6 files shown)


psakievich:
Wiener-Milenkovic stuff is all tested on device and seems to be working well.  The displacement functions I'm not sure of how to test yet.  I will need to generate some basic parameters of the blade. However, I think that is all stuff that can be tested in a follow on PR that is more focused on blade mesh operations.  At this point I think we have enough to start making full kernels that can operate on meshes and still be tested on device. I imagine we will refactor the functions in the @gantech's branches as needed during this process so I don't want to get ahead of myself by implementing them all at this point.

neilmatula (commented):
Looks good to me, but it would be best to get a second reviewer.

ashesh2512 (approved):

PR #1049, opened 2022-10-05
Pin pressure for convTaylorVortex

psakievich:
The convective Taylor vortex regression test has not been converging in the linear solve of the pressure system for quite some time.  Pinning pressure restores convergence.

@jrood-nrel @tasmith4 this will likely require a re-bless.


PaulMullowney (approved):
Good find!

ashesh2512:
That's interesting that it had been doing okay for so long. I recall having to pin it in the past.

ashesh2512 (approved):

jrood-nrel (approved):

PR #1054, opened 2022-10-20
Fix more warnings

psakievich:
Fix remaining GPU warnings.  Working on the segfault introduced by #1052

### DIFF
--- a/src/node_kernels/MomentumSSTAMSForcingNodeKernel.C
+++ b/src/node_kernels/MomentumSSTAMSForcingNodeKernel.C
@@ -64,21 +64,22 @@ MomentumSSTAMSForcingNodeKernel::MomentumSSTAMSForcingNodeKernel(
   forcingCompID_ = get_field_ordinal(meta, "forcing_components");
 
   // setup vectors
-  if (!solnOpts.eastVector_.empty() && !solnOpts.northVector_.empty()) {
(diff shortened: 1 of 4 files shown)


src/node_kernels/MomentumSSTAMSForcingNodeKernel.C
```diff
@@ -64,14 +64,21 @@ MomentumSSTAMSForcingNodeKernel::MomentumSSTAMSForcingNodeKernel(
   forcingCompID_ = get_field_ordinal(meta, "forcing_components");
 
   // setup vectors
-  DoubleView::HostMirror eastHost("eastHost", nDim_);
-  DoubleView::HostMirror northHost("northHost", nDim_);
-  for (int i = 0; i < nDim_; i++) {
-    eastHost(i) = solnOpts.eastVector_[i];
-    northHost(i) = solnOpts.northVector_[i];
+  if (!solnOpts.eastVector_.empty() && !solnOpts.northVector_.empty()) {
```
tasmith4, line 67: I feel like this would read more cleanly if it was all enclosed by `if (solnOpts.RANSBelowKs_)` (i.e. I think it would make it way more clear that all this stuff is only required in that case).

tasmith4 (approved):

PaulMullowney (approved):

PR #1058, opened 2022-10-25
FSI: Add funcs for driving FSI through AeroModel

psakievich:
In this PR we add the high level functions that will need to be public and external facing from the AeroModel to add FSI into the non-linear iteration process.

### DIFF
--- a/include/aero/solids/SolidBody.h
+++ b/include/aero/solids/SolidBody.h
@@ -1,0 +1,28 @@
+// Copyright 2017 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS), National Renewable Energy Laboratory, University of Texas Austin,
+// Northwest Research Associates. Under the terms of Contract DE-NA0003525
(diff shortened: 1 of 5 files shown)


include/aero/solids/SolidBody.h
```diff
@@ -0,0 +1,28 @@
...
+/** A parent class for implementing a solid body for aero dynamics applications
+ * This class should become the owner of surface post processing algorithms
+ * and instances will live inside the AeroContainer class
+ *
+ */
+class SolidBody
+{
+  // virtual class or template? do we think we'd want this to live on device
+  virtual void register_nodal_fields() = 0;
+  // std::unique_ptr<SurfaceFMPostProcessing> surfacePostProcessor;
```
psakievich, line 23: @ashesh2512 I think this is how we will want to implement you SMD idea.  We will tie in the mesh motion and surface processing algs to these objects.

PR #1061, opened 2022-10-30
radar: optionally don't save search data

rcknaus:
Resets the cached data of the search that's saved per line.  It can add up to a lot of memory when there's hundreds of search lines. Reusing the appropriate data between lines when possible is a better solution here; in lieu of that, this allows a user to just not save the data.


psakievich (approved):
I'll take a look at the CI and see what is going on today

PR #1062, opened 2022-10-31
WIP: AMD Support 

PaulMullowney:
With these mods, Nalu source files compiles but fails the link stage on crusher. I've built the GPU version on Summit. Unit tests, regression tests, and nrel5mw run successfully.

@overfelt I would like your input on my changes to master element and related code. There are warnings during device compilation that I would like to resolve.

@rcknaus I disabled the Lidar filesystem code when compiling for HIP. This is c++ 17 thing which doesn't compile on crusher at the moment.


jrood-nrel (approved):

include/AssembleElemSolverAlgorithm.h
```diff
@@ -87,7 +87,7 @@ class AssembleElemSolverAlgorithm : public SolverAlgorithm
         auto bktId = elem_buckets.device_get(team.league_rank());
         auto& b = ngpMesh.get_bucket(entityRank, bktId);
 
-#ifndef KOKKOS_ENABLE_CUDA
+#if !defined(KOKKOS_ENABLE_CUDA) && !defined(KOKKOS_ENABLE_HIP)
```
psakievich, line 90: I think we need to wrap this into a single macro and have something like `NALU_USING_GPU` or `NALU_USING_NGP` defined in a top level header like `NgpTypes.h` or something like that.

src/Simulation.C
```diff
@@ -51,6 +51,11 @@ Simulation::Simulation(const YAML::Node& root_node)
   cudaDeviceGetLimit(&default_stack_size, cudaLimitStackSize);
   cudaDeviceSetLimit(cudaLimitStackSize, nalu_stack_size);
 #endif
+
+#ifdef KOKKOS_ENABLE_HIP
```
psakievich, line 55: Are these stale or place holders for the future?

overfelt:
It seems most of the MasterElement changes were just decorating the implementation functions with KOKKOS_FUNCTION as they already were specified in the header declaration. We did not do that in the past but if it is needed and does not break the Cuda build, seems harmless.

The change to the MasterElement base class also looks harmless.  One of the functions that is now marked as KOKKOS_FUNCTION should not be called on device as it is host only but since that is only apparent at runtime, and if marking it avoids a compile error, it seems harmless.

rcknaus (approved):
Looks good for the std::filesystem stuff. I probably should switch back to `Ioss::FileInfo`.

psakievich (approved):

src/Simulation.C
PaulMullowney: yes

PR #1076, opened 2022-11-10
fix memory error in promotion unit test

rcknaus:
Should fix the address sanitizer issue in the unit tests. The code has been like this for a long time, so I'm surprised it's just been flagged recently.


psakievich (approved):

psakievich:
Could be due to the fact other things were failing first. This is the cleanest the dashboard has been in a while.

PR #1083, opened 2022-11-22
Use Exawind CDash

jrood-nrel:
I think we might want to start by prefixing the build name as well to be safe.


psakievich (approved):

psakievich:
Closes #1081

PR #1085, opened 2022-11-28
TKE Add missing sync call

psakievich:


alanw0 (approved):
This looks correct, and is fine to push.
But... what about the idea of having nalu_ngp::field_copy do the sync and modify internally? Are there scenarios where we wouldn't want it to do that?

rcknaus:
Since we're overwriting all of tkenp1 here, you don't need to `sync_to_device`.  Probably means we're doing something unnecessary in modifying tkenp1 before this.

 One question for @alanw0: how does the sync modification work with field rotation.  for instance, if we `modify_on_host` the np1 state, the rotate fields (np1->n->nm1->np1), does the modify tag go with it to the n state or stay on the np1?

alanw0:
> Since we're overwriting all of tkenp1 here, you don't need to `sync_to_device`. Probably means we're doing something unnecessary in modifying tkenp1 before this.
> 
> One question for @alanw0: how does the sync modification work with field rotation. for instance, if we `modify_on_host` the np1 state, then rotate fields (np1->n->nm1->np1), does the modify tag go with it to the n state or stay on the np1?

Ah, good point, we don't need to sync tkeNp1. But instead we probably need to sync tkeN right?

Good question on the rotation. I would have to dig through the code to remind myself to be certain. But for some reason I think the rotation syncs all the states back to host, then rotates, then marks all modified-on-host. So you don't want to do that more often than necessary.

psakievich:
> This looks correct, and is fine to push.
> But... what about the idea of having nalu_ngp::field_copy do the sync and modify internally? Are there scenarios where we wouldn't want it to do that?

That is what I thought, but the code is throwing when @lawrenceccheung is trying to use it for an ABL case with restarts. Let's see if this fixes their issue.  I think we need to revisit this though.

PR #1086, opened 2022-11-29
AMD MemSpace Fix

PaulMullowney:
Fix so that the memory space is defined properly. This fixes many problems.

Unit tests run now. Many of them pass but not all.

### DIFF
--- a/include/KokkosInterface.h
+++ b/include/KokkosInterface.h
@@ -45,7 +45,7 @@ namespace nalu {
 
 #if defined(KOKKOS_ENABLE_CUDA)
 typedef Kokkos::CudaSpace MemSpace;
-#elif defined(KOKKOS_HAVE_HIP)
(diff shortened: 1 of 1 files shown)


psakievich (approved):

alanw0 (commented):
Good catch.
I'm going to guess that KOKKOS_HAVE_OPENMP isn't a thing either. Probably doesn't matter though, since we don't care about openmp.

alanw0:
@ddement @djglaze @ldh4 did you guys see this?

PaulMullowney:
If you grep for KOKKOS_HAVE_OPENMP in kokkos you will find it in a few places. I think this is legacy that never got tested/built. In my previous PR, I think that I copied that line and replaced OPENMP with HIP. In any event, this fixes a few things.

PR #1091, opened 2022-12-13
Changes required for HIP porting

djglaze:
This commit contains the remainder of the ddement/hip_porting branch that has not already been merged into master by other developers.


PaulMullowney (approved):
Looks good. Does including these changes fix any tests?

djglaze:
> Looks good. Does including these changes fix any tests?

It's complicated.  It doesn't fix any of the current 99/148 failures under a ROCm-5.2 build, but it did fix many tests on the original branch under ROCm-5.1.  Most of those changes have already been merged into master, so this commit likely doesn't change much.  The GCL changes appear to change the failure mode for those tests from a hang to an immediate abort, which may be an improvement.

PaulMullowney:
> > Looks good. Does including these changes fix any tests?
> 
> It's complicated. It doesn't fix any of the current 99/148 failures under a ROCm-5.2 build, but it did fix many tests on the original branch under ROCm-5.1. Most of those changes have already been merged into master, so this commit likely doesn't change much. The GCL changes appear to change the failure mode for those tests from a hang to an immediate abort, which may be an improvement.

PrgEnv-amd/8.3.3 currently loads rocm 5.1. We also load 5.2 as part of the compilation. I wonder if we should downgrade crusher mods to 5.1.0?

djglaze:
> PrgEnv-amd/8.3.3 currently loads rocm 5.1. We also load 5.2 as part of the compilation. I wonder if we should downgrade crusher mods to 5.1.0?
>
> I'm also trying to get OLCF to upgrade PrgEnv-amd to more recent rocm releases.

We spent several hours yesterday with Phil Sakievich, trying to hot-wire a build to use ROCm-5.3 and failed spectacularly.  Our current suspicion is that the nonsense run-time errors we're getting now are due to something odd with our ROCm-5.2 environment.  Maybe it would make sense to temporarily go back to 5.1 until we can figure out what is going on, and possibly get the OLCF folks to give us an environment where we can build with 5.3.

psakievich (approved):

PR #1094, opened 2022-12-16
HIP Fix Number of threads in TeamPolicy

PaulMullowney:
Set number of threads to 128 rather than use Kokkos::AUTO

### DIFF
--- a/include/ngp_utils/NgpLoopUtils.h
+++ b/include/ngp_utils/NgpLoopUtils.h
@@ -42,7 +42,7 @@ inline TeamPolicy
 ngp_mesh_team_policy(
   const size_t sz, const size_t bytes_per_team, const size_t bytes_per_thread)
 {
-  TeamPolicy policy(sz, Kokkos::AUTO);
+  TeamPolicy policy(sz, NTHREADS_PER_DEVICE_TEAM);
   return policy.set_scratch_size(
     1, Kokkos::PerTeam(bytes_per_team), Kokkos::PerThread(bytes_per_thread));
 }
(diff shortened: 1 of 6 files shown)


jrood-nrel (approved):

PaulMullowney:
Actually. Don't merge this yet. I found another example that I want to fix.

src/Simulation.C
```diff
@@ -56,8 +56,8 @@ Simulation::Simulation(const YAML::Node& root_node)
 #endif
 
 #ifdef KOKKOS_ENABLE_HIP
-  // hipDeviceGetLimit(&default_stack_size, cudaLimitStackSize);
-  // hipDeviceSetLimit(cudaLimitStackSize, nalu_stack_size);
+  hipDeviceGetLimit(&default_stack_size, hipLimitMallocHeapSize);
+  hipDeviceSetLimit(hipLimitMallocHeapSize, nalu_stack_size);
```
alanw0, line 60: Excellent, I had thought that we believed these functions were not yet available. Are they newly available in 5.2 or later?

alanw0 (approved):

src/Simulation.C
PaulMullowney: We need 5.3 or later for these

PaulMullowney: Given that 5.4 seems to be the default in spack-manager, I think we're ok

psakievich: We can put a hard constraint on nalu-wind in spack that makes it so it won't build with anything less than 5.3 when needed.

include/ngp_utils/NgpLoopUtils.h
```diff
@@ -42,7 +42,11 @@ inline TeamPolicy
 ngp_mesh_team_policy(
   const size_t sz, const size_t bytes_per_team, const size_t bytes_per_thread)
 {
+#if defined(KOKKOS_ENABLE_HIP)
+  TeamPolicy policy(sz, 128);
```
psakievich, line 46: Can we make this number a macro so it only needs to be changed in one place going forward?

psakievich (changes requested):

include/ngp_utils/NgpLoopUtils.h
PaulMullowney: > Can we make this number a macro so it only needs to be changed in one place going forward?

Good idea

PaulMullowney:
Ready to merge. We're at 80% of the unit tests passing which is pretty good.

psakievich (approved):

psakievich:
@ddement let's try to avoid merging PR's that are failing the style check.

PR #1096, opened 2022-12-22
Nalu-Wind: Add framework for preset solvers.

neilmatula:


alanw0 (commented):
I like the idea of having solver presets and configurations, but I don't like putting these new classes and the yaml include into the existing LinearSolvers.h header. Can't they go in a new header (or multiple new headers)?
Generally we've regretted jamming more stuff into existing headers. We generally benefit from keeping things as separated/modularized as feasible.

neilmatula:
Yep, we can totally put this in a new header file.  I was considering that as a followup commit anyway.

include/LinearSolvers.h
```diff
@@ -87,7 +88,66 @@ class LinearSolvers
   //! Reference to the sierra::nalu::Simulation instance
   Simulation& sim_;
 
+  void register_presets();
+
+private:
+};
+
+class PresetSolverBase
```
psakievich, line 96: Overall this is fine.  It is a little more involved than I was imagining. I think it is kind of hard to read the solver settings this way.

I was thinking we could just embed the yaml in a header using [string literals](https://en.cppreference.com/w/cpp/language/string_literal), and embed the xmls in another header using string literals.  For an example in nalu-wind check here: https://github.com/Exawind/nalu-wind/blob/00452bfc20666bc92f5833f2dc66f9fb1f72a5b8/unit_tests/actuator/UnitTestActuatorFLLC.C#L22-L39

Then we can access the presets from a `std::map` rather than needing to create objects for each set of solver settings. something like
```c++
// in someheader.h
const char* scalar_tpetra_yaml = R"eq(  - name: solve_scalar
     type: tpetra
     method: gmres
     preconditioner: sgs
     tolerance: 1e-6
     max_iterations: 75
     kspace: 75
     output_level: 0)eq";
// new files
#include<someheader.h>
std::map<std::string, const char*> presets = {"scalar_tpetra", scalar_tpetra_yaml};
```
Overall I think this would be less infrastructure. This is my opinion though. I'm curious what the other reviewers think.

Also there might be better formats for saving the solver parameters than xml for `hypre` and `tpetra`.  Like a `teuchos::parameterlist`. Not sure what the equivalent is for `hypre`

psakievich (changes requested):
These format issues were because a PR was pushed without passing the format CI.  Will you please revert the format changes and then rebase?  That will remove the superfluous changes due to clang-format that aren't part of this PR.

marchdf:
@neilmatula this feels close to being able to merge? Or maybe no longer necessary?

PR #1102, opened 2023-01-11
Use Kokkos::LaunchBounds for HIP

alanw0:
We don't use LaunchBounds with Cuda. So ifdefs only add LaunchBounds to the Range/TeamPolicy objects if HIP is enabled.

nalu-wind is now able to build on crusher without the gpu-max-threads-per-block flag, and all NGP unit tests pass.

### DIFF
--- a/unit_tests/UnitTestNgpMesh1.C
+++ b/unit_tests/UnitTestNgpMesh1.C
@@ -10,6 +10,7 @@
 #include "UnitTestUtils.h"
 #include "UnitTestRealm.h"
 
+#include "KokkosInterface.h"
 #include "SimdInterface.h"
 #include "ElemDataRequests.h"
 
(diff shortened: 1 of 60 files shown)


alanw0:
Maybe somebody can show me how to run the clang-formatter on this code...

psakievich:
@alanw0 here is a script with how I do it
https://github.com/sandialabs/spack-manager/blob/f268a382142a201075f9f964017e56ec91382a80/configs/cee/env.sh#L6

unit_tests/UnitTestNgpMesh1.C
```diff
@@ -181,3 +182,177 @@ TEST_F(Hex8MeshWithNSOFields, NGPMeshField)
 
   test_ngp_mesh_field_values(*bulk, velocity, massFlowRate);
 }
+
+//struct MyBaseClass
```
psakievich, line 186: I think it would be good to replace these stale comments with a brief comment about the nuances of why these tests were added.  They are all fresh in our minds now, but 2-3 years from now it might be hard to remember why we wanted/needed  to test a fake field generation for nalu-wind.

psakievich (changes requested):
I really like the `DeviceRangePolicy` and `HostRangePolicy` additions. It makes the loops very intentional and easier to read/understand.  A couple of small clean ups noted with comments.  For `clang-format` you can run `clang-format -i` on the files.  I added a link to a script which shows how to apply it to the whole code base.  

Also, not sure what text editor you're using, but many of them have hooks to auto apply clang-format when you make changes.  VSCode has this.

unit_tests/UnitTestNgpMesh1.C
alanw0: Thanks @psakievich , I added a commit in which I ran clang-format and addressed your comments cleaning up the new unit-test.

alanw0:
Hmm, looks like the clang-format I ran on crusher is significantly different than the one github is running here... I'll figure this out in a little while.

jrood-nrel (approved):

alanw0:
Gents, I ran clang-format again, from another machine, using a llvm-13 version that Phil had. This seems to make the format check happy. It's annoying that the one on crusher wasn't doing the same thing... The one on crusher seems to be version 15.
But it still seems as though the difference is that the one on crusher wasn't going into the 'if KOKKOS_ENABLE_HIP' branches...

PaulMullowney:
Ok. I'm building now.

psakievich (approved):

alanw0:
> Ok. I'm building now.

ok I'll wait to merge it after you check some things. If this is neutral or an improvement, I'll go ahead and merge but if you find that this damaged anything I'll fix it up further before merging.

PaulMullowney:
I can repro that all unit tests pass. I've been able to get ablNeutralNGPHypreSegregated to run on 4 GPUs successfully. I think this is a big improvement..

alanw0:
> I can repro that all unit tests pass. I've been able to get ablNeutralNGPHypreSegregated to run on 4 GPUs successfully. I think this is a big improvement. That being said, the sims are taking forever to cleanup and exit. Maybe this is a file system issue?

Hmm I don't know why cleanup and exit is slow. I was only running unit-tests, only 1 mpi rank...

PaulMullowney:
> > I can repro that all unit tests pass. I've been able to get ablNeutralNGPHypreSegregated to run on 4 GPUs successfully. I think this is a big improvement. That being said, the sims are taking forever to cleanup and exit. Maybe this is a file system issue?
> 
> Hmm I don't know why cleanup and exit is slow. I was only running unit-tests, only 1 mpi rank...

This only seems to happen with 4 or more GPUs. Ugh. I would say merge this though we have more problems to debug :(

PR #1103, opened 2023-01-12
Fix some RangePolicies

psakievich:
#1102 got a few RangePolicies wrong in the OpenFAST executors.  All actuator code currently executes on host.


alanw0 (approved):
Glad you caught these!

PaulMullowney (approved):

Issue #1088, opened 2022-12-01
Improve installation documentation regarding the existence of a docker container


neumantm:
I recently wanted to check out nalu-wind, but struggled with the installation.
Later I learned (by reading issues), that there is a docker container which I could have used.
Knowing this, would probably have helped me getting started with nalu-wind.

neumantm:
If this is desired and there a no reasons against it, I could also create a PR.

psakievich:
@neumantm that would be very helpful and appreciated!

PR #1104, opened 2023-01-12
Add small doc section about building in docker container

neumantm:
Resolves: #1088

### DIFF
--- a/docs/source/user/build_spack.rst
+++ b/docs/source/user/build_spack.rst
@@ -211,3 +211,21 @@ the dependencies by using ``spack location -i <package>``. For example in the
    make
 
 There are also ``do-config`` scripts available for this according to machine under the configs directory `here <https://github.com/exawind/build-test>`__. These scripts may also provide the capability to access and use pre-built dependencies from a set of modules if they are available on the machine. This should allow you to have a build of Nalu-Wind in which you are able to continuosly modify the source code and rebuild.
+
+Building inside Docker Container
+--------------------------------
+
+It is also possible to build (and run) Nalu-Wind inside a docker container with a prepared environment.
+This has the benefit of requiring less setup and usually being faster and can be useful in some situations (e.g. quickly testing something).
+
+For this, use the container `ecpe4s/exawind-snapshot <https://hub.docker.com/r/ecpe4s/exawind-snapshot>`_:
+
+::
+
+   docker run -it ecpe4s/exawind-snapshot bash
+
+Inside the container you can directly install Nalu-Wind with spack:
+
+::
+
+   spack install nalu-wind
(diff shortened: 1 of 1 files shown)


docs/source/user/build_spack.rst
```diff
@@ -211,3 +211,21 @@ the dependencies by using ``spack location -i <package>``. For example in the
...
+
+::
+
+   docker run -it ecpe4s/exawind-snapshot bash
+
+Inside the container you can directly install Nalu-Wind with spack:
+
+::
+
+   spack install nalu-wind
```
psakievich, line 231:
```suggestion
   spack install nalu-wind
   
 `Spack-Manager <https://github.com/sandialabs/spack-manager>`_ is used generate the container and is pre-installed in the container.  Additional development of the code can be done through the Spack-Manager's developer tools.  A quick-start guide for these tools is available `here <https://sandialabs.github.io/spack-manager/user_profiles/developers/developer_workflow.html>`_.  This workflow is how the CI builds are generated in GitHub actions.
 
If you are uncomfortable using spack for development then you can also set up your own build workflow in more of a "roll-your-own" development environment using CMake and just utilize the pre-installed TPL's that are available in the container through spack.  This really comes down to your development preferences and your comfort level with CMake.
```

Technically `nalu-wind` is already installed in the container so doing `spack install nalu-wind` should just re-use that install.  I've added some documentation on how you can use `spack-manager` to do some extra development.

psakievich (changes requested):
Thanks for getting this started.  It's super helpful when people add documentation right after they figure things out.  I've added some suggested text. We can iterate on them further as needed.

docs/source/user/build_spack.rst
neumantm: Ah right, didn't notice that. Then maybe we should write `spack load nalu-wind` (indicating it is already installed) instead?

psakievich (approved):
LGTM. `spack load nalu-wind` seems like a nice addition that would be more semantically correct

docs/source/user/build_spack.rst
```diff
@@ -211,3 +211,25 @@ the dependencies by using ``spack location -i <package>``. For example in the
...
+It is also possible to build (and run) Nalu-Wind inside a docker container with a prepared environment.
+This has the benefit of requiring less setup and usually being faster and can be useful in some situations (e.g. quickly testing something).
+
+For this, use the container `ecpe4s/exawind-snapshot <https://hub.docker.com/r/ecpe4s/exawind-snapshot>`_:
+
+::
+
+   docker run -it ecpe4s/exawind-snapshot bash
+
+Inside the container you can directly install Nalu-Wind with spack:
```
psakievich, line 227:
```suggestion
Inside the container you can directly load the pre-installed version of Nalu-Wind with spack:
```

PR #1117, opened 2023-01-25
Fixed regression test selection on crusher

djglaze:
When running regression tests on crusher, the full set was being selected.  Only a subset are currently expected to run, presumably matching the subset that runs on CUDA platforms.  The test selection logic was enhanced to guard against both ENABLE_CUDA and ENABLE_ROCM.

### DIFF
--- a/reg_tests/CTestList.cmake
+++ b/reg_tests/CTestList.cmake
@@ -158,7 +158,7 @@ function(add_test_r_cat testname np ncat)
     endif()
 endfunction(add_test_r_cat)
 
-if(NOT ENABLE_CUDA)
+if(NOT ENABLE_CUDA AND NOT ENABLE_ROCM)
 
   #=============================================================================
   # Regression tests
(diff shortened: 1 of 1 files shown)


PaulMullowney:
Do all the new tests pass with this change?

jrood-nrel (approved):

psakievich (approved):

PaulMullowney (approved):

djglaze:
> Do all the new tests pass with this change?

Nope.  3 of the 16 tests fail.  My next task is to patch up the code so that they run correctly.

PaulMullowney:
> > Do all the new tests pass with this change?
> 
> Nope. 3 of the 16 tests fail. My next task is to patch up the code so that they run correctly.

which tests?

djglaze:
> > > Do all the new tests pass with this change?
> > 
> > 
> > Nope. 3 of the 16 tests fail. My next task is to patch up the code so that they run correctly.
> 
> which tests?

- drivenCavity_p1
- conduction_p4
- taylorGreenVortex_p3

PaulMullowney:
> > > > Do all the new tests pass with this change?
> > > 
> > > 
> > > Nope. 3 of the 16 tests fail. My next task is to patch up the code so that they run correctly.
> > 
> > 
> > which tests?
> 
> * drivenCavity_p1
> * conduction_p4
> * taylorGreenVortex_p3

taylorGreen has been a challenge on GPUs (CUDA) for a while. Not sure about the other 2.

PR #1118, opened 2023-01-27
make matrix free optional

rcknaus:
* brought this up at the exawind meeting.  Defaulted to "ON" currently, since that's the previous behavior. If we make tpetra optional, switch the flag to "OFF".
* Be less brave about the null space for pressure. This should make it less likely to have issues on different platforms at some small performance penality.
* Removed some soon-to-be-private kokkos header includes


alanw0 (approved):
I think this looks good.
I already have changes in progress to make trilinos solvers optional. As part of those I was turning off matrix-free at the same time, but it's better to give matrix-free a separate option.
It turns out that we won't be able to turn off tpetra for the trilinos build, because it's needed by zoltan2 which we also need. But we will be able to turn off muelu, belos, etc. I think.

PaulMullowney (approved):

PaulMullowney:
Thanks for doing this so quickly! Have you tested builds with/without this option?

psakievich (approved):

rcknaus:
> Thanks for doing this so quickly! Have you tested builds with/without this option?

Yup. Works both ways.

PR #1121, opened 2023-01-29
FSI: Deflection ramping strategy based on angle about hub and radial location

psakievich:
This PR implements a ramping strategy based on the angle between the blade root, hub and mesh point to deal with the interface between blocks. It uses a radial strategy to ramp along the span.  The product of these two factors becomes the total displacement ramp along the blades.

The inputs for the user will be a distance along the span to ramp from zero to full displacements i.e. 3m, and in theta they will give an angle where blade deflections will be zero ( a little before 60 degrees for 3 blades makes sense), and a ramp distance in radians to go from a factor of 1.0 to 0.0.

The mechanics are demonstrated in the unit-tests.  Working on testing on the coarse mesh.

Updated a temporal ramping that will allow blades to only go rigid body rotations up to a `startRamp` time and then full blade deflections will be gradually ramped in until the `endRamp` time is reached.  I'm somewhat hopeful this will allow us to drop the requirement of running and blending with AeroDyn 🤞 

New inputs:
```yaml
deflection_ramping:
  temporal_ramp_start: [time when we want to start blending in full blade deflections]
  temporal_ramp_end: [time when full blade deflections will be active]
  span_ramp_distance: [distance from the blade root to ramp blade deflections for pitch]
  zero_theta_ramp_angle: [azimuthal location from the blade core where deflections should also be zero (degrees)]
  theta_ramp_span: [angle to perform the azimuthal ramping (degrees)]
```

Not sure if we need the theta ramp any more though. It is optional now.


psakievich:
Ramping looks good, but there is another bug. 
image

psakievich:
Pretty sure I know what's wrong. This mesh has cells in the blade block below the blade root. That is where everything screwy is happening.  We need to either split the blocks so that is hub data or add a computation that just treats those with the hub displacements. @gantech

psakievich:
image
Latest changes.

marchdf (approved):

psakievich:
Okay I think this is ready to merge.  @gantech I looked making the span ramping non-dimensional, but I think it requires some mapping from the blades to get the last point on each blade (if they had different number of points and were different sizes).  We could constrain that and take the option away, implement a mapping (different PR please), or just leave it as is. I'm partial to option 3 for now...

src/mesh_motion/CMakeLists.txt
```diff
@@ -10,9 +10,4 @@ target_sources(nalu PRIVATE
    ${CMAKE_CURRENT_SOURCE_DIR}/MotionTranslationKernel.C
    ${CMAKE_CURRENT_SOURCE_DIR}/MotionWavesKernel.C
 )
-
-if(ENABLE_OPENFAST_FSI)
-   target_sources(nalu PRIVATE
-     ${CMAKE_CURRENT_SOURCE_DIR}/FrameOpenFAST.C
-   )
 endif()
```
psakievich, line 13:
```suggestion
```

gantech (approved):

PR #1123, opened 2023-02-01
Make trilinos solver support optional

alanw0:
Uses cmake variable ENABLE_TRILINOS_SOLVERS
and nalu-wind macro NALU_USES_TRILINOS_SOLVERS

These both mirror corresponding names that are used to enable/disable HYPRE.

There is an interaction with ENABLE_MATRIXFREE:
Matrix-free can only be enabled if trilinos-solvers are enabled.


CMakeLists.txt
```diff
@@ -25,14 +27,22 @@ option(ENABLE_PARAVIEW_CATALYST
 option(ENABLE_TIOGA "Use TIOGA TPL to perform overset connectivity" OFF)
 option(ENABLE_OPENMP "Enable OpenMP flags" OFF)
 option(ENABLE_BOOST  "Enable Boost libraries" OFF)
-option(ENABLE_MATRIXFREE "Enable high-order matrix-free computation" ON)
-option(NALU_WIND_SAVE_GOLDS  "Save gold files to directory when running tests" OFF)
+option(ENABLE_MATRIXFREE "Enable high-order matrix-free computation" OFF)
```
psakievich, line 30: I think we should leave this defaulted to on or else it might go stale

psakievich (changes requested):

PaulMullowney (approved):
Thanks Alan!
I assume you've test with/without.?
Have you seen any improvement in link time on crusher with this?
Is there a coming spack-manager PR?

CMakeLists.txt
rcknaus: Yeah, I envisioned it more as a [dependent option](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html) on by default with tpetra on and always off if tpetra is off.

alanw0: Yes, I can tweak that... the complication is, it can't be on if trilinos-solvers is off, and there's no spack-spec mechanism for flipping it...

alanw0:
> 

Yes it builds with/without.
Haven't tested it on crusher yet, but I will do that next.
I had thought there was also a spack-manager change, but I must have been confused looking at the spack change that we put in yesterday. So it looks like it's just the spack change and this nalu-wind change.

CMakeLists.txt
alanw0: @rcknaus I can adjust the CMake file to leave matrix-free on by default, and turn it off if trilinos-solvers are off. That's easier than adding the spack mechanism.

rcknaus: @alanw0 that works

alanw0:
Turns out we do have a spack-manager change... I just put the PR up for that.

CMakeLists.txt
```diff
@@ -14,7 +14,9 @@ option(ENABLE_DOCUMENTATION "Build documentation." OFF)
 option(ENABLE_SPHINX_API_DOCS "Link Doxygen API docs to Sphinx" OFF)
 option(ENABLE_WIND_UTILS "Build wind utils along with Nalu-Wind" OFF)
 option(ENABLE_FFTW "Use the FFTW library to support ABLTopBC" OFF)
-option(ENABLE_HYPRE "Use HYPRE Solver library" OFF)
+option(ENABLE_HYPRE "Use HYPRE Solver library" ON)
```
alanw0, line 17: I'm not sure I meant to push this change that turns hypre on by default. Should I turn this back off?

psakievich: No we want it on by default now

PaulMullowney: I would make it on by default. The trilinos solvers just aren't getting the same level of attention.

psakievich (approved):

PR #1126, opened 2023-02-06
extend allowable line length

tasmith4:


jrood-nrel:
I have no preference on what this value is and I don't want to be in a debate about it, so I will remove myself as a reviewer.

psakievich (approved):

PR #1127, opened 2023-02-06
First phase: ifdef out unit-tests that fail on gpu platform.

alanw0:
This allows us to run (and pass) 434 unit-tests, compared to the 150 that run with the '*NGP*' filter.

Next phase will be to evaluate these ifdefd-out unit-tests and see which are easy/feasible to get running on a gpu platform.

 But this now allows us to run the unittestX executable on GPU platforms without specifying the '*NGP*' filter.

### DIFF
--- a/unit_tests/UnitTestKokkosME.C
+++ b/unit_tests/UnitTestKokkosME.C
@@ -90,10 +90,11 @@ copy_DoubleType0_to_double(
   }
 }
 
+template <typename SHMEM>
 void
 compare_old_scv_volume(
-  const sierra::nalu::SharedMemView<DoubleType**>& v_coords,
-  const sierra::nalu::SharedMemView<DoubleType*>& scv_volume,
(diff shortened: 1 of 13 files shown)


jrood-nrel (approved):

psakievich (changes requested):
@alanw0 will you also remove the filter we apply [here](https://github.com/Exawind/nalu-wind/blob/f58c204419131fc6c2c36958d67642ba48550b74/reg_tests/CTestList.cmake#L127-L139) as part of this PR?

PaulMullowney (approved):

alanw0:
> @alanw0 will you also remove the filter we apply [here](https://github.com/Exawind/nalu-wind/blob/f58c204419131fc6c2c36958d67642ba48550b74/reg_tests/CTestList.cmake#L127-L139) as part of this PR?

Yes I can do that.

psakievich:
Thanks!

Issue #1070, opened 2022-11-04
exceptions when running examples


jthies:
Hi,
we're new to nalu-wind and trying to run an example or two to see how it works.
There may be obvious usage errors, in which case I apologize for opening an issue.
I followed the user guide [here](https://nalu-wind.readthedocs.io/_/downloads/en/latest/pdf).

I managed to compile everything, but in the simulation step with naluX, I get these exceptions:

## abl_neutral:'''
'''
  what():  TKE active with wall bc, however, no value of tke or wall function specified
'''

## turbine_uniform_inflow

 what():  Symmetry::fatal_error()

In the latter case, there was an outdated entry in a yaml file (.i had to be replaced by .yaml), so I suspect that these examples are not part of continuous integration testing. Any idea what might be going on? I checked out nalu-wind at revision ead59abf5 (master branch),
and openfast at 5f3fb6ef74f4. I had to patch the cmake system here and there but did not change the code or example files.

tasmith4:
@psakievich do you know what's going on here or do you know who might?  It wouldn't surprise me at all if our examples were out-of-date ...

psakievich:
Yes these are all out of date. We have had an issue for sometime to remove them see #836. Since we haven't been keeping them up to date.  That said, it might be worth adding this into our nightly test process to run the examples, but I'm not sure when we could get to it.  I think there is value in the examples cases but it takes consistent effort and reminders to keep them up to date.

I'm also not sure if the wind-utils repo is up-to-date either.

@jthies for CI tested input files please see: https://github.com/Exawind/nalu-wind/tree/master/reg_tests/test_files/ablNeutralEdge

The regression test files are kept up to date and tested nightly.  We apologize for the inconvenience, but you should be able to debug any diffs with this as a reference.

Also for help building you can checkout  the nalu-wind package in https://github.com/spack/spack
We are very heavily dependent on spack these days and it is our preferred method of deployment.

Another option is to pull down one of our docker containers https://hub.docker.com/r/ecpe4s/exawind-snapshot/tags.
A new image is pushed every night to docker hub.

jthies:
Thanks for the suggestions!

- The problem with the second test case (turbine_uniform_inflow) is that mesh parts have different names in the different yaml files. It's running in principle, although I now run into another issue.
- I had a few minor issues with wind-utils and openfast, in particular, in both packages there's a line liike 
'''
link_libraries(yaml-cxx)
'''
missing.
[...]

PR #1130, opened 2023-02-07
Document how to run a simple example using regression test case data

neumantm:
This is useful to quickly check if the installation of nalu-wind works.

This was inspired by [a comment by @psakievich](https://github.com/Exawind/nalu-wind/issues/1070#issuecomment-1304081221)

### DIFF
--- a/docs/source/user/building.rst
+++ b/docs/source/user/building.rst
@@ -7,3 +7,4 @@ Building Nalu-Wind
 
    Building Nalu-Wind with Spack <build_spack.rst>
    Building Nalu-Wind Manually <build_manually.rst>
+   Verifying that the installation of Nalu-Wind works <verify_installation.rst>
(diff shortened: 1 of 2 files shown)


jrood-nrel:
Maybe we don't have enough documentation on it, but using `ctest` is the preferred method to run the regression tests.

neumantm:
Yes I understood this.
The goal of this is to document how to (mis)use a regression test case as a very simple example to see if the installation works.
That is what @psakievich suggested in the comment I linked.
Maybe I should change the title of the PR.

psakievich:
@neumantm would you also add in an example of running the unittests?  I think my preference for people testing the code is first to run the unit tests and then they can manually run the regression tests.  We always build the unit tests right now, and as a plus they don't require the mesh repo which is 600MB (recently down from 1GB thanks to @jrood-nrel :pray:)

neumantm:
Sure. Makes sense. 
They just need to run `unittestX`, am I right?

psakievich (approved):

psakievich:
@neumantm yes that is correct.

marchdf:
@psakievich this also should have been merged?

PR #1142, opened 2023-02-14
Change FSIturbine::projectPt2Line to be device-compatible

alanw0:
Also convert perpProjectDist_Pt2Line.

These methods are converted into free-functions instead of class members, since they have no side-effects and don't reference class-member data. Free-functions are also trivial to unit-test.

### DIFF
--- a/src/aero/fsi/FSIturbine.C
+++ b/src/aero/fsi/FSIturbine.C
@@ -16,6 +16,7 @@
 #include "stk_util/parallel/ParallelReduce.hpp"
 #include "stk_mesh/base/FieldParallel.hpp"
 #include "stk_mesh/base/Field.hpp"
+#include "stk_math/StkMath.hpp"
 #include "master_element/MasterElement.h"
 #include "master_element/MasterElementFactory.h"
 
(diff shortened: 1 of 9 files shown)


src/aero/fsi/FSIturbine.C
```diff
@@ -31,6 +32,68 @@ namespace sierra {
...
+
+  double nonDimCoord = 0.0;
+
+  double num = 0.0;
+  double denom = 0.0;
+
+  for (int i = 0; i < 3; i++) {
+    num += (pt[i] - lStart[i]) * (lEnd[i] - lStart[i]);
+    denom += (lEnd[i] - lStart[i]) * (lEnd[i] - lStart[i]);
+  }
```
psakievich, line 56: If we use `vs::Vectors` this whole loop reduces to:
```suggestion
num = (pt-lStart)&(lEnd-lStart);
dnom = vs::mag_sqr(lEnd-lStart);
```

psakievich (changes requested):
@alanw0 this looks good. I'm wondering if you would mind using the `vs::Vector` datatype instead of `stk::Point`?  That is the datatype I've been using in the rest of the utilities since it makes these sort of vector/tensor operations simple to read. I'd like to keep consistency in this area of the code if we can to minimize excessive type conversions `&` is a dot product and `^` is a cross product.

We should make it an automatic type conversion between `vs::Vectort` and `stk::Point` though.

I also think it might be good to extract this function totally out of the `FSIturbine` and put it with the utilities like here https://github.com/Exawind/nalu-wind/tree/master/include/aero/aero_utils since it is generic code.

src/aero/fsi/FSIturbine.C
alanw0: ok, I'll look into that. I hadn't noticed `vs::Vector`.

psakievich: Yeah it is new. We pulled it from AMR-Wind to help with some of the vector math in the fsi stuff. It is a device compatible data structure.

alanw0:
@psakievich has the version of clang-format changed? I ran the one you pointed me to on our system (under llvm-13) and it's not doing these formatting changes...

psakievich (approved):

psakievich:
@alanw0 yes it should be llvm@13

alanw0:
> @alanw0 yes it should be llvm@13

@psakievich well then that's weird, I don't know of an easy way to make the correct formatting changes.

PR #1143, opened 2023-02-15
More progress towards unit-testing the fsiTurbine object.

alanw0:
Still just doing construction/setup. But the CylinderMesh fixture now creates the fields needed by fsiTurbine.


psakievich:
@alanw0 this is looking good. I want to do a little refactoring of the FSIturbine class. I'm running into some problems with data inconsistencies.  Let's discuss outside this PR when you get a chance?

PR #1145, opened 2023-02-21
Fix Cuda compiler warnings.

overfelt:
By removing the KOKKOS declaration from host only
functions and marking functions inline.


alanw0 (approved):

psakievich (approved):

PaulMullowney:
Yeah. I remember trying to fix those warnings and failing on crusher :(

PR #1148, opened 2023-02-22
Revert "Fix Cuda compiler warnings."

jrood-nrel:
Reverts Exawind/nalu-wind#1145

I think this broke compiling on Crusher, so I'm putting this as a draft in case anyone else can confirm, if we need to revert this.


overfelt:
Bummer!  It broke the Dr. Crusher build?  There is another approach of removing the template and using only function overloading for the two types of arguments we are passing but that is so close to what we do now it might still fail.  The step after that is to rename the one function to two different functions which would be unfortunate to have to do.

psakievich (approved):

PR #1150, opened 2023-02-23
Start using the FieldManager/FieldRegistry

psakievich:
Add a `FieldManager` object to `Realm` and start using it to register fields.  The starting point here is the fields registered in `Realm`. Need to figure out how to update the mesh repo since the git history was purged.

### DIFF
--- a/include/FieldDefinitions.h
+++ b/include/FieldDefinitions.h
@@ -24,9 +24,28 @@ struct FieldDefinition
 
 using FieldDefScalar = FieldDefinition<ScalarFieldType>;
 using FieldDefVector = FieldDefinition<VectorFieldType>;
+using FieldDefGeneric = FieldDefinition<GenericFieldType>;
+using FieldDefGenericInt = FieldDefinition<GenericIntFieldType>;
+using FieldDefTpetraId = FieldDefinition<TpetIDFieldType>;
+using FieldDefLocalId = FieldDefinition<LocalIdFieldType>;
+using FieldDefGlobalId = FieldDefinition<GlobalIdFieldType>;
 
-using FieldDefTypes = std::variant<FieldDefScalar, FieldDefVector>;
-using FieldPointerTypes = std::variant<ScalarFieldType*, VectorFieldType*>;
+using FieldDefTypes = std::variant<
+  FieldDefScalar,
+  FieldDefVector,
+  FieldDefGeneric,
+  FieldDefGenericInt,
+  FieldDefTpetraId,
+  FieldDefLocalId,
+  FieldDefGlobalId>;
+using FieldPointerTypes = std::variant<
+  ScalarFieldType*,
+  VectorFieldType*,
+  GenericFieldType*,
+  GenericIntFieldType*,
+  TpetIDFieldType*,
+  LocalIdFieldType*,
+  GlobalIdFieldType*>;
 
 } // namespace nalu
 } // namespace sierra
--- a/include/FieldDefinitions.h
+++ b/include/FieldDefinitions.h
@@ -29,6 +29,7 @@ using FieldDefGenericInt = FieldDefinition<GenericIntFieldType>;
 using FieldDefTpetraId = FieldDefinition<TpetIDFieldType>;
 using FieldDefLocalId = FieldDefinition<LocalIdFieldType>;
 using FieldDefGlobalId = FieldDefinition<GlobalIdFieldType>;
+using FieldDefHypreId = FieldDefinition<HypreIDFieldType>;
 
 using FieldDefTypes = std::variant<
   FieldDefScalar,
@@ -37,7 +38,9 @@ using FieldDefTypes = std::variant<
   FieldDefGenericInt,
   FieldDefTpetraId,
   FieldDefLocalId,
-  FieldDefGlobalId>;
+  FieldDefGlobalId,
+  FieldDefHypreId>;
+
 using FieldPointerTypes = std::variant<
   ScalarFieldType*,
   VectorFieldType*,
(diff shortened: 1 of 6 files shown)


src/FieldRegistry.C
```diff
@@ -27,10 +27,19 @@ Registry()
...
+  FieldDefTpetraId TpetraId = {stk::topology::NODE_RANK};
+  FieldDefGlobalId GlobalId = {stk::topology::NODE_RANK};
+  FieldDefHypreId HypreId = {stk::topology::NODE_RANK};
+
+  // clang-format off
   static const std::map<std::string, FieldDefTypes> registry = {
     {"velocity", MultiStateNodalVector},
     {"temperature", MultiStateNodalScalar},
+    {"hypre_global_id", HypreId},
+    {"tpet_global_id", TpetraId},
```
alanw0, line 39: It seems like these should be conditional on whether hypre or trilinos-solvers are enabled or not.

include/FieldDefinitions.h
```diff
@@ -29,6 +29,7 @@ using FieldDefGenericInt = FieldDefinition<GenericIntFieldType>;
 using FieldDefTpetraId = FieldDefinition<TpetIDFieldType>;
 using FieldDefLocalId = FieldDefinition<LocalIdFieldType>;
 using FieldDefGlobalId = FieldDefinition<GlobalIdFieldType>;
+using FieldDefHypreId = FieldDefinition<HypreIDFieldType>;
```
alanw0, line 32: Somehow I hadn't noticed FieldDefinition previously. It seems to just add rank and num-states to field, but field already holds those attributes. i.e., each field already knows its rank and num-states.

src/Realm.C
```diff
@@ -921,26 +921,15 @@ void
 Realm::setup_nodal_fields()
 {
 #ifdef NALU_USES_HYPRE
-  hypreGlobalId_ = &(meta_data().declare_field<HypreIDFieldType>(
-    stk::topology::NODE_RANK, "hypre_global_id"));
+  fieldManager_->register_field("hypre_global_id", meta_data().get_parts());
 #endif
-  tpetGlobalId_ = &(meta_data().declare_field<TpetIDFieldType>(
-    stk::topology::NODE_RANK, "tpet_global_id"));
+  fieldManager_->register_field("tpet_global_id", meta_data().get_parts());
```
alanw0, line 926: This is pre-existing before your commit, but tpet_global_id should be conditional on NALU_USES_TRILINOS_SOLVERS

overfelt (approved):

src/Realm.C
```diff
@@ -921,26 +921,15 @@ void
...
-#ifdef NALU_USES_HYPRE
-    stk::mesh::put_field_on_mesh(*hypreGlobalId_, *parts[ipart], nullptr);
-#endif
-    stk::mesh::put_field_on_mesh(*tpetGlobalId_, *parts[ipart], nullptr);
-    stk::mesh::field_fill(
-      std::numeric_limits<LinSys::GlobalOrdinal>::max(), *tpetGlobalId_);
-  }
+  auto globId = fieldManager_->get_field_ptr("nalu_global_id");
+  stk::mesh::field_fill(
+    std::numeric_limits<LinSys::GlobalOrdinal>::max(),
```
alanw0, line 931: Another pre-existing thing... 0 would be a better initial value than max (assuming this initial value should be obviously invalid), because stk-mesh knows that 0 is an invalid ID (and will throw in certain places if it encounters 0) whereas it doesn't know that max is invalid.

alanw0 (approved):
I entered a couple of minor gripes, but I wouldn't necessarily hold this up over them.

include/FieldDefinitions.h
psakievich: Yes, once the field has been created it knows this information. The registry exists to define all the information about the field needed to create it. The end intent is that a field can be registered and populated to parts from a single call that only requires the field's name as a string. That is when the field actually gets registered through the `FieldManager::register_field` call.

src/FieldRegistry.C
psakievich: Since this is the registry I don't think it matters too much. I should test that there aren't any problems though. These are just a list of all the fields that __could__ theoretically get created so as long as the types exist it should work fine.  We could if/def these, but since we do it on the actual registration calls (and we __have__ to do it there) it is not strictly necessary here.  My opinion is it would make this portion of the code messier, so I'd rather leave them out if possible.

alanw0: Yes I agree. I hadn't been thinking of this as a static set of info rather than fields actually being registered.

PaulMullowney (approved):

psakievich:
I'm not a huge fan of how we have to access the field pointers with `std::get`. This is something I'd like to improve in the future so we can get rid of the cached pointers. In the very least I think the cached pointers in realm need to go, but I think we should do that in a follow on PR b/c it touches a fair amount of code.

src/Realm.C
```diff
@@ -912,26 +921,22 @@ void
...
-#ifdef NALU_USES_HYPRE
-    stk::mesh::put_field_on_mesh(*hypreGlobalId_, *parts[ipart], nullptr);
+#ifdef NALU_USES_TRILINOS_SOLVERS
+  fieldManager_->register_field("tpet_global_id", meta_data().get_parts());
+  // TODO work on removing this variable from realm by accessing fields through
+  // the manager instead
+  tpetGlobalId_ =
+    std::get<TpetIDFieldType*>(fieldManager_->get_field_ptr("tpet_global_id"));
+  stk::mesh::field_fill(
+    std::numeric_limits<LinSys::GlobalOrdinal>::min(), *tpetGlobalId_);
```
psakievich, line 935:
```suggestion
    std::numeric_limits<LinSys::GlobalOrdinal>::max(), *tpetGlobalId_);
```

alanw0 (approved):

PR #1151, opened 2023-02-24
FSI: Remove theta ramp

psakievich:
The theta ramping we implemented was ill-posed and would occasionally generate nan's when computing the angle between a point and the hub-to-blade-root vector.  Fortunately, it is not required at all anymore since we are using the hub deflections as the base deflection.

This PR removes the theta ramping all together. This allows my surrogate mesh case to run fine.

### DIFF
--- a/src/aero/fsi/FSIturbine.C
+++ b/src/aero/fsi/FSIturbine.C
@@ -91,18 +91,11 @@ fsiTurbine::fsiTurbine(int iTurb, const YAML::Node& node)
       !defNode,
       "defleciton_ramping inputs are required for FSI Turbines with blades");
     DeflectionRampingParams& defParams = deflectionRampParams_;
-    double* zeroTheta = &defParams.zeroRampLocTheta_;
-    double* thetaRamp = &defParams.thetaRampSpan_;
     // clang-format off
     get_required(defNode, "temporal_ramp_start", defParams.startTimeTemporalRamp_);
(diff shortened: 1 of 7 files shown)


marchdf (approved):

gantech (commented):
This looks good. Did we test this on the sample case? Does the pitch ramping look ok?

src/aero/fsi/FSIturbine.C
```diff
@@ -240,6 +242,12 @@ fsiTurbine::setup(std::shared_ptr<stk::mesh::BulkData> bulk)
     deflectionRamp_ = &(meta.declare_field<ScalarFieldType>(
       stk::topology::NODE_RANK, "deflection_ramp"));
 
+  distanceToRoot_ = meta.get_field<ScalarFieldType>(
+    stk::topology::NODE_RANK, "distance_to_root");
+  if (distanceToRoot_ == NULL)
+    distanceToRoot_ = &(meta.declare_field<ScalarFieldType>(
```
gantech: Is there a way we can add fields that are created only in debug mode?

psakievich:
> This looks good. Did we test this on the sample case? Does the pitch ramping look ok?

Yes the theta ramping didn't really have an affect after we moved to using the hub for the stiff blade deflections.

src/aero/fsi/FSIturbine.C
psakievich: I can make it so we only use this field if it is requested. I thought I deleted this field though.

PR #1155, opened 2023-03-01
Clip min wall dist only if wall part is present

gantech:


src/ShearStressTransportEquationSystem.C
```diff
@@ -530,28 +530,30 @@ ShearStressTransportEquationSystem::clip_min_distance_to_wall()
...
-      const double minD = ndtw.get(mi, 0);
-
-      ndtw.get(mi, 0) = stk::math::max(minD, wallNormDist.get(mi, 0));
-    });
-  ndtw.modify_on_device();
-
-  stk::mesh::parallel_max(realm_.bulk_data(), {minDistanceToWall_});
-  if (realm_.hasPeriodic_) {
-    realm_.periodic_field_max(minDistanceToWall_, 1);
+  if (!wallBcPart_.is_empty()) {
```
marchdf, line 533: Could we just do right as you enter this function:
```
if (!wallBcPart_.is_empty()) {
   return;
 }
```
 and then not have the extra indent stuff? This is just stylistic and maybe people don't like that.

psakievich: Oh people like it 🙂

gantech: Done! @ashesh2512 had the same comment for me in person!

gantech:
I think it's good to go. Someone with the authority... please merge!

psakievich (approved):

PR #1156, opened 2023-03-02
Experiment with pulling fsiTurbine::mapLoads into free-functions

alanw0:
The free-functions have more arguments than I was hoping for, but they should be relatively easily unit-tested, and they don't depend on OpenFast.

### DIFF
--- a/unit_tests/aero/UnitTestFSIturbine.C
+++ b/unit_tests/aero/UnitTestFSIturbine.C
@@ -14,6 +14,7 @@
 
 #include "KokkosInterface.h"
 #include "aero/fsi/FSIturbine.h"
+#include "aero/fsi/OpenfastFSI.h"
 
 namespace {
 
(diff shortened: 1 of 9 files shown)


unit_tests/aero/UnitTestFSIturbine.C
```diff
@@ -57,4 +206,74 @@ TEST_F(CylinderMesh, construct_FSIturbine)
...
+
+  fsiTurb->computeLoadMapping();
+
+  // stk::mesh::field_fill(0.1, *tforceSCS_);
+  set_x_mesh_disp(*bulk, 0.1);
+  set_x_values(fsiTurb->brFSIdata_.twr_def, 0.1);
+
+  fsiTurb->mapLoads();
+
+  constexpr double expectedEndLoad = -120;
```
psakievich, line 262: How is this computed? I must be missing it, but I don't see where the loads are set or computed in the unit test fixture.

alanw0: Yes it took me quite a while to figure out how to make it produce non-zero loads.
fsiTurbine::mapLoads calls `calc_loads_->execute()` which computes the loads. I was able to make it produce nice round numbers like -120, -240, by setting fields like viscosity, density etc to all ones in the unit-test fixture.

unit_tests/aero/UnitTestFSIturbine.C
```diff
@@ -39,15 +40,163 @@ const std::string fsiInputs = "tower_parts: [block_1] \n"
...
+  "sim_start: init\n"
+  "t_start: 0.0\n"
+  "t_max:   13.106159895150718\n"
+  "n_checkpoint: 144\n"
+  "dt_FAST: 0.0011376874908984999\n"
+  "n_substeps: 40\n"
+  "Turbine0:\n"
+  "  turbine_base_pos: [5.0191, 0., -89.56256]\n"
+  "  turbine_hub_pos: [5.0191, 0.0, 0.0 ]\n"
+  "  FAST_input_filename: nrel5mw.fst\n"
```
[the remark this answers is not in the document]
alanw0: I'm not sure about that. This is basically a paste of what you sent me on slack. I fiddled with it enough to get the OpenfastFSI object to construct and setup without throwing. But the actually calculation of the loads (which is what I really wanted to unit-test may not be using these values if I'm not getting all of the parsing hooked up correctly.

unit_tests/aero/UnitTestFSIturbine.C
```diff
@@ -57,4 +206,74 @@ TEST_F(CylinderMesh, construct_FSIturbine)
...
+
+  EXPECT_EQ(0, params.nBRfsiPtsTwr);
+  EXPECT_EQ(0, params.numBlades);
+
+  set_tower_ref_pos(*bulk, *fsiTurb);
+
+  fsiTurb->computeLoadMapping();
+
+  // stk::mesh::field_fill(0.1, *tforceSCS_);
+  set_x_mesh_disp(*bulk, 0.1);
```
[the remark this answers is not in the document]
alanw0: It seemed like the calculation of the loads was only producing zeros if I didn't set any displacements. I can look at that again if you don't think that sounds right.

unit_tests/aero/UnitTestFSIturbine.C
psakievich: ah interesting

psakievich (changes requested):
After discussion I agree with the strategy here. I think we just need to add a few comments to outline the portions of the test that are purely checking computation to monitor for changes, and not necessarily assuring correctness. The same goes for operations that are added like adding displacements to get loads.  If that is still required let's just document that in the code so we know it is fair game to change the behavior in future refactors.

psakievich (approved):

PR #1157, opened 2023-03-02
FieldManager: Improve interface to various field states

psakievich:
This PR allows users to register a field and set the number of states at registration time. This will override what is in the registry for the field description.

An ability to get the specific field state is also added to `FieldManager::get_field_ptr`

### DIFF
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -2716,13 +2716,7 @@ Realm::register_nodal_fields(stk::mesh::Part* part)
   // register high level common fields
   // Declare volume/area_vector fields
   const int numVolStates = does_mesh_move() ? number_of_states() : 1;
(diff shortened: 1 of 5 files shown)


src/Realm.C
```diff
@@ -2716,13 +2716,7 @@ Realm::register_nodal_fields(stk::mesh::Part* part)
   // register high level common fields
   // Declare volume/area_vector fields
   const int numVolStates = does_mesh_move() ? number_of_states() : 1;
```
psakievich, line 2718: @gantech @ashesh2512 is this right? Shouldn't it be `has_mesh_deformation()`?  We only need multiple dnv states for deformation right?

alanw0 (approved):

PR #1159, opened 2023-03-07
New master element factory,

overfelt:


PaulMullowney:
These are significant changes. What prompted this refactoring? I would ask that this branch be tested on crusher and the nrel5mw sim shown to run successfully.

alanw0:
> These are significant changes. What prompted this refactoring? I would ask that this branch be tested on crusher and the nrel5mw sim shown to run successfully.

We discussed this at our meeting last thursday. The idea is to make it easier for classes like ElemDataRequestsGPU to get the needed master-element. Currently we just have a pointer and depend on the user to give a device-compatible version of it to ElemDataRequests, then ElemDataRequestsGPU has to assume it's good and unit-tests are showing that it's not always good. With a new singleton-registry approach, it should be easy to get a master-element, given a topology, for host or device from anywhere. That's the goal.

alanw0 (approved):
James! I think these changes look good, I like how it is now more clear that you're getting a host or device pointer, and the the GPU object can get a device pointer given a host pointer.
However, given how touchy crusher seems to be, I agree with Paul that it would be nice to know for sure that these changes work on crusher. Have you (or can you) check these up on crusher?

jrood-nrel:
I was planning to try this on Crusher with the 16 turbine case.

alanw0:
> I was planning to try this on Crusher with the 16 turbine case.

ok, great.

overfelt:
Is the crusher spack install working?  Last time I tried it seemed like a software update might have broken the install.

jrood-nrel (approved):
This is running the 16 turbine case on Crusher with Nalu-Wind on the GPU.

PaulMullowney (approved):

PR #1079, opened 2022-11-15
Add VolumeOfFluid EqSys

psakievich:
Getting started on multiphase. Pinging @wjhorne @rcknaus

PR #1160, opened 2023-03-08
Move unit test to correct location in Cmake

psakievich:
Fix issue with Crusher builds caused by #1079


PaulMullowney (approved):

psakievich:
Not sure if this will be too costly, but it would be good to test the Hypre only path in CI since we need to track it for the challenge problem.

alanw0 (approved):

jrood-nrel (approved):

PR #1169, opened 2023-03-30
New field register work.

overfelt:
Register the fields in UnitTestAlgorithm.C with the new FieldReister.

Added a Tensor type to complete the Scalar, Vector, Tensor list of data types.  Since it came up twice in this test it seemed like a reasonable thing to have.  Like Vector, Tensor has compile time dimensions based on 2D and 3D fields.


overfelt:
Looked into removing Realm from this unit test.  Since the meta data and bulk data come right off of Realm and are needed to run the test, it seems the realm is also needed.

alanw0 (approved):

psakievich (approved):

PR #1172, opened 2023-04-12
Work to merge NgpFieldManager into new field manager.

overfelt:
So, the idea is to get rid if nalu_ngp::FieldManager and push the logic from that class into nalu::FieldManager. One class that heavily uses nalu_ngp::FieldManager is ElemDataRequestsGPU.  So template ElemDataRequestsGPU on the FieldManager type so either can be used. Then slowly change all of the occurrences of ElemDataRequestsGPU from using nalu_ngp::FieldManager to using nalu::FieldManager.


alanw0 (approved):
James you are slashing and burning!
Looks good to me, assuming the gpu tests are still working.
I need to raise my std-c++ game, I'm not very familiar with new-fangled stuff like std::visit and std::get etc.

psakievich:
@alanw0 you should check it out. We could potentially migrate a lot of our `std::vector` of pointers to `std::variant`. This talk has cool demo of `std::variant` https://www.youtube.com/watch?v=G9MxNwUoSt0

psakievich (approved):
Looks good to me. Thanks @overfelt!

overfelt:
These code changes work on CPU and GPU but the interface needs to be improved.  The ElemDataRequestsGPU class uses just ints for field lengths and dimensions but needs to use the new FieldManager which has templated function calls.  So there is this code:
```
      stk::mesh::NgpField<double>& fld_ptr = dim == 1 ? 
      fieldMgr.get_field_ptr<ScalarFieldType*>(name) :
      fieldMgr.get_field_ptr<VectorFieldType*>(name);
```
So there is still work to do but it does look promising.

alanw0 (approved):

PR #1174, opened 2023-04-17
Begin to sketch out a beam class

alanw0:


include/aero/aero_utils/Beam.h
```diff
@@ -0,0 +1,105 @@
...
+    const size_t vecSize = numPoints * 6;
+    refPos.resize(vecSize);
+    def.resize(vecSize);
+    load.resize(vecSize);
+
+    for (size_t pt = 0; pt < numPoints; ++pt) {
+      for (int d = 0; d < 6; ++d) {
+        refPos[pt * 6 + d] = m_refPos(pt).position_[d];
+        def[pt * 6 + d] = m_def(pt).position_[d];
+        load[pt * 6 + d] = m_load(pt).position_[d];
```
overfelt, line 78: @alanw0 , indexing position_ from 0 to 5 even though it is of length 3 probably works since orientation_ follows position_, but seems misleading.  Maybe m_load(pt)[d] would be more clear?

alanw0: Thanks James, good catch. I meant to do `m_load(pt)[d]`.

psakievich (approved):
This looks like a good start to me.

PR #1178, opened 2023-04-27
Add more use of the new FieldManager in the unit tests.

overfelt:


unit_tests/UnitTestSuppAlgDataSharing.C
```diff
@@ -169,46 +171,45 @@ public:
...
+      "elemTensorField", parts, 0, 4);
 
   fill_mesh("generated:10x10x10");
 
-  TestAlgorithm testAlgorithm(*bulk);
-
   // TestSuppAlg constructor says which data it needs, by inserting
   // things into the 'dataNeededByKernels_' container.
 
   SuppAlg* suppAlg = new TestSuppAlg(
```
psakievich, line 210: This is fine for an intermediate step, but ideally I'm imagining we would just pass a `FieldManager` into `TestAlgorithm`'s ctor and do all the initialization there.

unit_tests/kernels/UnitTestKernelUtils.C
```diff
@@ -956,702 +955,6 @@ calc_exposed_area_vec(
   areaVec.sync_to_host();
 }
 
-#if !defined(KOKKOS_ENABLE_GPU)
```
psakievich, line 959: @overfelt I trust you on this, but I'm also curious as to why are we removing all these tests?

src/FieldRegistry.C
```diff
@@ -54,21 +57,31 @@ Registry()
     {"dwdx",                      SingleStateNodalVector},
     {"edge_area_vector",          SingleStateEdgeVector},
     {"effective_viscosity" ,      SingleStateNodalScalar},
+    {"elemCentroid",              SingleStateElemVector},                          
     {"element_volume",            SingleStateElemScalar},
+    {"elemScalarField",           SingleStateElemScalar},  // Used in testing
```
psakievich, line 62: Not for this PR, but a potential refactor that comes to mind is a separate registry for unit tests.
In the mean time, to make this more organized could we stick these unit-test only fields at the end of the registry and do a block comment to indicated fields below are only for tests?

unit_tests/UnitTestSuppAlgDataSharing.C
overfelt: If TestAlgorithm should be able to work with only a FieldManager and not a whole stk::mesh::BulkData then I'll try to pass just a FieldManager.  The BulkData is used quite a bit and it is not obvious that FieldManager is all that is needed.  Maybe after looking at the functions that use the BulkData, I will see how FieldManager could be used instead.

unit_tests/kernels/UnitTestKernelUtils.C
overfelt: I started to refactor all of these functions when I noticed the `#if 0` disabling a couple of functions.  After deleting all of that code, I saw that what was left was not called from any other function that I could find, so I just deleted that also.  These are not tests, but utility functions and I figured if the unit tests compiled, linked and ran without the these utilities, we could remove it.

unit_tests/UnitTestSuppAlgDataSharing.C
psakievich: I was thinking of just adding an argument to the constructor of the TestAlgorithm so it takes the BulkData and a FieldManager. My main thought is that all the field registration that a given algorithm needs should be encapsulated in the Algorithm itself.

FWIW FieldManager has a [reference](https://github.com/Exawind/nalu-wind/blob/58ca0ce7a3deee179ec25b999ac4b60a2b8a945d/include/FieldManager.h#L44) to stk::mesh::MetaData and BulkData can be populated from there. Right now it is private, but we could expose it if that makes sense. If we find that we are always passing FieldManager and Meta/BulkData together then I would say we should expose it.

src/FieldRegistry.C
overfelt: Yes we could put all of the testing field names at the end of the std::map constructor.  The alphabetical listing is just a convention.  It does seem like we have quite a few field names and there are probably more to  add. I still think the amount of memory used in these maps is trivial and the map search still quite fast at this size.

unit_tests/UnitTestSuppAlgDataSharing.C
overfelt: I changed the little test algorithm, `TestSuppAlg`, to register its own fields on the field manager during construction.  There was a change to the mesh constructor since the mesh constructor finalizes the MetaData to exclude new fields. I just moved the mesh constructor after the algorithm constructor, but not sure how that will change in the general case.

psakievich: Okay that's good to know. Something to be aware of.  I suspect that for the main code it won't really make an impact since we register fields all over the place.  mesh finalize is a specific call in Realm IIRC.

psakievich (approved):

PR #1180, opened 2023-04-30
Initialize current coordinates after mesh transformation

### LINKED ISSUE #1179: Overset issue with mesh motion
There appears to be some strange behavior when multiple mesh blocks are defined, but mesh motion is only provided on some of them.
This was observed for a case with rigid-body rotation prescribed for the rotor, but a stationary tower.
In this case, the overset seems to be calculated wrong for the tower, as in the attached screenshot.  The residuals also blow up, particularly TKE.
However, if zero mesh motion is explicitly provided for the tower in the mesh motion input block, everything is fine.
So it would seem that mesh motion has to be explicitly defined on all blocks, even if they are not intended to move.
Screen Shot 2023-04-27 at 5 56 28 PM

ashesh2512:
Mesh transformation changes the model coordinates of the mesh. Current coordinates, used by overset for hole-cutting, should be initialized after that. Resolves #1179


psakievich (approved):

ashesh2512:
@jrood-nrel The gold norms were generated on Mac. Might need to be updated for the test in this PR.

PR #1181, opened 2023-05-02
Add IDDES unit test

marchdf:
The implementation of the term used to limit production in IDDES and DES was not the one proposed in the DES/IDDES paper by Gritskevich (2011), see equation 3 in the appendix. This changes also enables Nalu-Wind to use the same formulation as amr-wind.


marchdf:
Talked to @sbidadi9 and it seems like Nalu-Wind went with the Menter 2003 paper, not Gritskevich 2011. AMR-Wind did the opposite. I am going to remove the changes to DES here and change things on the amr-wind side instead. This PR will now just be the addition of a DES unit test.

sbidadi9:
Discussed with @marchdf about not changing the production term being limited for IDDES model. The current implementation is based on the paper by Menter (2003):

image

Sharing the link to the paper as well for reference: https://www.researchgate.net/publication/228742295_Ten_years_of_industrial_experience_with_the_SST_turbulence_model

This implementation is correct is the same as in Nalu-Wind and is correct.

However, what is in AMR-Wind is based on production limiting for SST - which is not correct.

psakievich (approved):

sbidadi9 (approved):

PR #1182, opened 2023-05-03
Example usage of FieldManager in an NGP Kernel

psakievich:
This PR demonstrates how a Kernel can be converted to use a FieldManager.

The main goal is to move the field registration away from higher level abstractions like realm and equationsystem, to lower ones (algs and kernels).

This will essentially eliminate the need for `RegisterNodalFields` type calls, and will eliminate a range of bugs that can come in from not registering fields in the right place.  Having kernels and algs register the fields they need will also reduce the amount of fixturing needed for unit tests since constructing the object will register the fields for you.  You can then finalize the mesh and populate the fields with initial values. (Not demonstrated in this PR).

As a side bonus we don't need to store class members for field ordinals AND field pointers on kernels any more either.


src/node_kernels/TKESSTNodeKernel.C
```diff
@@ -19,29 +19,31 @@
...
+  const stk::mesh::MetaData& meta,
+  const FieldManager& manager,
+  stk::mesh::Part* part)
+  : NGPNodeKernel<TKESSTNodeKernel>(), nDim_(meta.spatial_dimension())
 {
+  manager.register_field("turbulent_ke", part);
+  manager.register_field("specific_dissipation_rate", part);
+  manager.register_field("density", part);
+  manager.register_field("turbulent_viscosity", part);
+  manager.register_field("dudx", part);
```
alanw0, line 32: I wonder if we should have some central location where field-names are associated with an enum or something. Because having string-literals scattered around in various kernels could be an opportunity for errors.

src/node_kernels/TKESSTNodeKernel.C
```diff
@@ -19,29 +19,31 @@
...
-    sdrID_(get_field_ordinal(meta, "specific_dissipation_rate")),
-    densityID_(get_field_ordinal(meta, "density")),
-    tviscID_(get_field_ordinal(meta, "turbulent_viscosity")),
-    dudxID_(get_field_ordinal(meta, "dudx")),
-    dualNodalVolumeID_(get_field_ordinal(meta, "dual_nodal_volume")),
-    nDim_(meta.spatial_dimension())
+TKESSTNodeKernel::TKESSTNodeKernel(
+  const stk::mesh::MetaData& meta,
+  const FieldManager& manager,
+  stk::mesh::Part* part)
```
psakievich, line 25: As @overfelt pointed out, if this is in the constructor it needs to take a partvec so all the parts are registered. Or we need a secondary registration process.  Planning to pursue the former with #1183

psakievich:
Making some progress on this.  I really want to get this swarmable in the next few weeks and work as a team to convert the code to use the fieldManager and SmartFields as much as possible. It would be a good way for a lot of newer people to see a lot of nalu-wind's structure.

src/node_kernels/TKESSTNodeKernel.C
psakievich: Actually looking at this closer, I'm not so sure we need a part vec. I think the field registry process will just add more parts each time it is called after a field is registered since multiple identical registrations are a no-op in stk. It would be good to test it though.

psakievich:
I don't plan to pick this back up.

Issue #1183, opened 2023-05-04
Registration of interior algs needs to use a partvec instead of a part


psakievich:
The current infrastructure is looping over parts where a partvec could be assembled instead. It will require a lot of code changes but it will improve things.

https://github.com/Exawind/nalu-wind/blob/472317ff83d8b20f0775a0ea1ea66030d5bb13f9/src/EquationSystems.C#L342-L359

marchdf:
I think this is fixed, looking at the history. Closing. Please reopen if not.

psakievich:
It's partially complete but not needed unless someone plans to pursue the SmartFields concept further. I don't see myself having time to devote to that in the near future.

PR #1185, opened 2023-05-09
Replace stk::mesh::Part* with stk::mesh::PartVector&.  Issue #1183

overfelt:
PartVectors allow for the specification of a field on a list of parts instead of one part at a time. For some classes it was only possible to pass a single part pointer for the field specification where a part vector would be more flexible.

Started with register_nodal_fields:
from: register_nodal_fields(stk::mesh::Part* part)
to:   register_nodal_fields(const stk::mesh::PartVector &part_vec)

The stk::mesh::put_field_on_mesh function does not take a PartVector so create a Selector form a PartVector:
  stk::mesh::Selector selector = stk::mesh::selectUnion(part_vec);
and use that instead. There were a couple of places where the construction of a PartVector from a single Part pointer:
   PartVector(1,Part*)
was useful.

There were a lot of files changed, but the changes are trivial.

### DIFF
--- a/include/utils/StkHelpers.h
+++ b/include/utils/StkHelpers.h
@@ -102,14 +102,14 @@ get_node_field(
 void register_scalar_nodal_field_on_part(
   stk::mesh::MetaData& meta,
   std::string name,
-  const stk::mesh::Part& selector,
+  const stk::mesh::Selector& selector,
   int num_states,
   double ic = 0);
 
 void register_vector_nodal_field_on_part(
   stk::mesh::MetaData& meta,
   std::string name,
-  const stk::mesh::Part& selector,
+  const stk::mesh::Selector& selector,
   int num_states,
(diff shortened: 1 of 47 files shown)


include/utils/StkHelpers.h
```diff
@@ -102,14 +102,14 @@ get_node_field(
 void register_scalar_nodal_field_on_part(
```
psakievich, line 102: This looks redundant for what the `FieldManager` does. These functions look like candidates for the chopping block in the future.

overfelt: Yes, I saw that register_scalar_nodal_field_on_part was only used in a couple of places but was not sure if it was the future or the past. I'll remove it on the next merge request.

Yes, This is just the first of a few merge requests to replace Part* with PartVector. Just wanted to get this in and work on other things for a few days.

PR #1198, opened 2023-07-16
add a string-function temperature IC/dirichlet bc

rcknaus:
Adds the ability to use a string function to specify a temperature, to handle a more arbitrary temperature profile as requested by Ken.  The string function runs on device but the auxfunction still doesn't.

Also gets rid of a bunch of miscellaneous warnings in the code too.


psakievich:
😍 let's make these work for bc's and ic's too!

psakievich (approved):

psakievich:
Yeah lets do that for now. We haven't blessed the latest trilinos on the gpu platforms yet. The CI will be using the default trilinos that most dev's are building against.

PR #1201, opened 2023-07-31
Always dump input file to the logfile

psakievich:
This makes it so the input file is echoed into the log file so we always know what parameters were used to run the case.

What do people think?  It will make the logfiles larger, but it will also decrease uncertainty on what was used for a given run.

### DIFF
--- a/nalu.C
+++ b/nalu.C
@@ -172,124 +172,132 @@ main(int argc, char** argv)
 
     // proceed with reading input file "document" from YAML
     YAML::Node doc = YAML::LoadFile(inputFileName.c_str());
+    if (!naluEnv.parallel_rank()) {
+      std::cout << std::string(20, '#') << " INPUT FILE START "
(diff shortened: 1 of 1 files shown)


lawrenceccheung (approved):
I think this is a great idea.  It'll help in debugging and removing uncertainty about what was actually used in the input file.

Lawrence

ashesh2512 (approved):

ashesh2512 (approved):
What does the log file look like? Is a line of #  obvious enough at first glance to separate the input file dump from the rest of the log file?

alanw0 (approved):

rcknaus:
This is a good change. It's also possible to push the input file into the output exodus files. I've found that useful before for documentation, since the output files often live around longer than the log files.

psakievich:
We should make sure this behavior gets ported to the driver too.

PR #1203, opened 2023-08-21
VOF consistent Mass-Momentum + advanced sharpening

wjhorne:
These changes introduce mass-momentum consistency for VOF calculations while also improving sharpening.


wjhorne:
I'll format the text and re-bless the droplet case, which has changed to reflect sharper interfaces, as for some final work here.

src/edge_kernels/VOFAdvectionEdgeAlg.C
```diff
@@ -39,27 +42,51 @@ VOFAdvectionEdgeAlg::VOFAdvectionEdgeAlg(
...
+  massForcedFlowRate_ =
+    get_field_ordinal(meta, "mass_forced_flow_rate", stk::topology::EDGE_RANK);
+  density_ = get_field_ordinal(meta, "density", stk::mesh::StateNP1);
+  velocity_n_ = get_field_ordinal(meta, "velocity", stk::mesh::StateN);
+
+  std::map<PropertyIdentifier, MaterialPropertyData*>::iterator itf =
+    realm_.materialPropertys_.propertyDataMap_.find(DENSITY_ID);
+
+  if (itf == realm_.materialPropertys_.propertyDataMap_.end()) {
+    density_liquid_ = 1000.0;
```
psakievich, line 54: What is the use case for this?  Don't we want a missing property map entry to throw as well?

src/edge_kernels/VOFAdvectionEdgeAlg.C
```diff
@@ -84,108 +144,112 @@ VOFAdvectionEdgeAlg::execute()
...
-      alhsfac = 0.5 * (vdot - stk::math::abs(vdot)) * alphaUpw;
+      alhsfac = 0.5 * vdot;
       smdata.lhs(1, 1) -= alhsfac / relaxFac;
       smdata.lhs(0, 1) += alhsfac;
 
-      // Compression term
-      const DblType velocity_scale = stk::math::abs(
-        vdot / stk::math::sqrt(av[0] * av[0] + av[1] * av[1] + av[2] * av[2]));
+      // Compression + Diffusion term
+      const DblType velocity_scale = global_max_velocity * 5.0;
```
psakievich, line 172: Will you add a comment/reference for the rational of the hard coded `5.0`?

src/edge_kernels/VOFAdvectionEdgeAlg.C
```diff
@@ -39,27 +42,51 @@ VOFAdvectionEdgeAlg::VOFAdvectionEdgeAlg(
...
+                               "calculations. Use a constant or "
+                               "VOF property for density.");
+      break;
+    }
+    }
+  }
 }
 
 void
 VOFAdvectionEdgeAlg::execute()
```
psakievich, line 81: It would be good to add some comments to acknowledge the paper where this advection scheme came from and to note and differences during the implementation.  Since we were one deep on the development I'd like to be a little more pedantic about it for this particular case.

psakievich (changes requested):
Overall looks good to me. Just some small book keeping/comments requests, and I would like to test the change in moving predicted states inside the nonlinear loop on a current production model like the FSI turbine.

src/edge_kernels/VOFAdvectionEdgeAlg.C
wjhorne: This is for the unit testing of VOFAdvectionEdgeAlg which would just throw unless all the piping for property maps were introduced. I could also make it something that is passed into the routine directly instead.

src/edge_kernels/VOFAdvectionEdgeAlg.C
wjhorne: I'll add a comment here. It comes from a formula in Jain, 2022

wjhorne:
Moved this over to the multiphase_dev branch

PR #1204, opened 2023-08-21
Add coordinate syncs for device compatibility

psakievich:
Add syncs for fields.  I have not confirmed if these fix the FSI case on frontier yet, but I wanted to open a PR.  I'm going to look for some more

### DIFF
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -2561,6 +2561,10 @@ Realm::set_current_coordinates(stk::mesh::Part* targetPart)
   VectorFieldType* displacement = meta_data().get_field<VectorFieldType>(
     stk::topology::NODE_RANK, "mesh_displacement");
 
+  modelCoords->sync_to_host();
+  currentCoords->sync_to_host();
+  displacement->sync_to_host();
+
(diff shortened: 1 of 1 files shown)


ashesh2512 (approved):

src/Realm.C
```diff
@@ -2561,6 +2561,10 @@ Realm::set_current_coordinates(stk::mesh::Part* targetPart)
   VectorFieldType* displacement = meta_data().get_field<VectorFieldType>(
     stk::topology::NODE_RANK, "mesh_displacement");
 
+  modelCoords->sync_to_host();
+  currentCoords->sync_to_host();
```
alanw0, line 2565: I think this is correct but overly aggressive.
Since currentCoords is going to be over-written, a sync-to-host seems unnecessary.
But the modify_on_host for it down below definitely seems necessary. Because that will signal any subsequent sync_to_device calls that the data does indeed need to be sync'd.

alanw0 (approved):

src/Realm.C
psakievich: We just had a successful run with this commit so I am going to merge as is.   We can get less aggressive later. :)

PR #1193, opened 2023-06-14
Fixes for FSI with ROCM

psakievich:
The original code failed to compile with ROCM. Evidently the types must be more explicit for ROCM than Cuda in this instance.

PR #1205, opened 2023-08-29
Crusher fix

psakievich:
Fixes solver convergence issues in run on crusher ROCM run. Not sure I understand the subtly of why yet. Below is the output of the first timestep in Nalu-Wind. This ran with no issues for the entire 30 min of wall clock time I allocated in the queue.

``` console
*******************************************************
Time Step Count: 1 Current Time: 0.00344353
 dtN: 0.00344353 dtNm1: 0.00344353 gammas: 1 -1 0
 DualNodalVolume min: 8.74301e-12 max: 0.737549 total: 119470
 Max Courant: 4073.99 Max Reynolds: 278166 (realm_1)
realm_1::advance_time_step() 
NLI    Name           Linear Iter      Linear Res     NLinear Res    Scaled NLR
---    ----           -----------      ----------     -----------    ----------
Aero models - Execute
1/4    Equation System Iteration
 1/1      myLowMach
        MomentumEQS_X          6        0.00219027       0.339946             1
        MomentumEQS_Y          6        0.00122359       0.208249             1
        MomentumEQS_Z          5       0.000648655      0.0273747             1
        MomentumEQS           17        0.00259137         0.3996             1
        ContinuityEQS         21          0.177085         7.0387             1
 1/1ShearStressTransportWrap
        TurbKineticEnergyEQS   5        0.00675613        0.70956             1
        SpecDissRateEQS        6           7257.94    1.08703e+06             1
2/4    Equation System Iteration
 1/1      myLowMach
        MomentumEQS_X          6          0.277873        14.9757       44.0533
        MomentumEQS_Y          6           0.12425        6.02642       28.9385
        MomentumEQS_Z          6          0.109949        3.21674       117.508
        MomentumEQS           18          0.323636        16.4602       41.1917
        ContinuityEQS         13           17.0319        584.207       82.9992
 1/1ShearStressTransportWrap
        TurbKineticEnergyEQS  98        0.00140091      0.0383274     0.0540157
        SpecDissRateEQS        9           21002.5    1.00507e+06      0.924603
3/4    Equation System Iteration
 1/1      myLowMach
        MomentumEQS_X         41          0.219607        6.31151       18.5662
        MomentumEQS_Y         37          0.112566        3.25961       15.6524
        MomentumEQS_Z         48         0.0387412        1.29877       47.4441
        MomentumEQS          126          0.249799        7.22129       18.0713
        ContinuityEQS         19           10.2296        319.018       45.3234
 1/1ShearStressTransportWrap
        TurbKineticEnergyEQS  13        0.00435818       0.122828      0.173104
        SpecDissRateEQS        1        1.1994e+06     8.6938e+10       79977.7
4/4    Equation System Iteration
 1/1      myLowMach
        MomentumEQS_X         10          0.100901        3.03511       8.92823
        MomentumEQS_Y         11         0.0316671        1.98288       9.52169
        MomentumEQS_Z          9         0.0137491       0.567972       20.7481
        MomentumEQS           30          0.106643        3.66965       9.18331
        ContinuityEQS         23           5.35502        145.578       20.6825
 1/1ShearStressTransportWrap
        TurbKineticEnergyEQS  11        0.00190954      0.0766999      0.108095
        SpecDissRateEQS        1            955315    3.46896e+11        319124
Aero models - Predict model time step
```


src/gcl/MeshVelocityAlg.C
```diff
@@ -79,8 +79,7 @@ MeshVelocityAlg<AlgTraits>::MeshVelocityAlg(Realm& realm, stk::mesh::Part* part)
     NUM_IP, isoParCoords_, isoCoordsShapeFcnHostView_.data());
   Kokkos::deep_copy(isoCoordsShapeFcnDeviceView_, isoCoordsShapeFcnHostView_);
 
-  auto scsFaceNodeMapHostView =
-    Kokkos::create_mirror(scsFaceNodeMapDeviceView_);
+  Kokkos::View<int[12][4], Kokkos::HostSpace>scsFaceNodeMapHostView;
```
psakievich, line 82:
```suggestion
  Kokkos::View<int[12][4], Kokkos::HostSpace> scsFaceNodeMapHostView;
```

src/gcl/MeshVelocityEdgeAlg.C
```diff
@@ -71,8 +71,7 @@ MeshVelocityEdgeAlg<AlgTraits>::MeshVelocityEdgeAlg(
     19, &isoParCoords_[0], isoCoordsShapeFcnHostView_.data());
   Kokkos::deep_copy(isoCoordsShapeFcnDeviceView_, isoCoordsShapeFcnHostView_);
 
-  auto scsFaceNodeMapHostView =
-    Kokkos::create_mirror(scsFaceNodeMapDeviceView_);
+  Kokkos::View<int[12][4], Kokkos::HostSpace>scsFaceNodeMapHostView;
```
psakievich, line 74:
```suggestion
  Kokkos::View<int[12][4], Kokkos::HostSpace> scsFaceNodeMapHostView;
```

src/gcl/MeshVelocityAlg.C
```diff
@@ -79,8 +79,7 @@ MeshVelocityAlg<AlgTraits>::MeshVelocityAlg(Realm& realm, stk::mesh::Part* part)
     NUM_IP, isoParCoords_, isoCoordsShapeFcnHostView_.data());
   Kokkos::deep_copy(isoCoordsShapeFcnDeviceView_, isoCoordsShapeFcnHostView_);
 
-  auto scsFaceNodeMapHostView =
-    Kokkos::create_mirror(scsFaceNodeMapDeviceView_);
+  Kokkos::View<int[12][4], Kokkos::HostSpace> scsFaceNodeMapHostView;
```
overfelt, line 82: Doesn't the change from create_mirror to a direct definition reverse a previous commit?  I wonder if int[12][4] is unstable across platforms and should be int** instead. But if this gets Frontier up and running, great catch.

overfelt (approved):

src/gcl/MeshVelocityAlg.C
alanw0: Yes, I agree with these questions. It would seem like `[12][4]` is specific to a single element/face topology. Is this algorithm only going to run on a single kind of elements/faces? Generally `create_mirror` is considered safer because it automatically creates the same type/shape as the device view.
But like James says, if this makes it work then it's hard to argue against.

psakievich: Looking at the CI failure indicates something is amiss here.

alanw0: Do we have any unit tests for this?

overfelt: If this is causing a problem in running the Challenge Problem, how about moving the const int scsFaceNodeMap_[12][4] array from the header and into the one place it is used inside the execute() function, replacing Kokkos::View scsFaceNodeMap with this const int array.  Just avoid copying it over to device at all since it is const data anyway.

PaulMullowney: What's really odd is that this only seems to be a problem when using FSI.

PaulMullowney: There should be a way to query the size and then loop over the dims correctly. This is host cost after all.

ashesh2512: This algorithm gets used only when FSI is activated.

psakievich: Yes the test that is failing CI.

psakievich:
For reference #1193 was the PR where we made these changes.

src/gcl/MeshVelocityAlg.C
alanw0: If James' solution of making it into a const int array is feasible, that sounds good.
But I agree with Paul, it should be easy to query the sizes and then that should guarantee correct looping. You can get those sizes with `.extent(0)` and `.extent(1)`, I think. Would need to verify exact syntax...

ashesh2512: @psakievich Might be worthwhile to query the sizes just as a sanity check, even though this is hard coded for one element type. Might reveal our underlying issue?

psakievich: okay. the gcl unit test doesn't run on device. need to find out why. We probably need to spin up a parallel test suite of this on summit to make sure we don't break cuda too

src/gcl/MeshVelocityEdgeAlg.C
```diff
@@ -71,8 +71,7 @@ MeshVelocityEdgeAlg<AlgTraits>::MeshVelocityEdgeAlg(
     19, &isoParCoords_[0], isoCoordsShapeFcnHostView_.data());
   Kokkos::deep_copy(isoCoordsShapeFcnDeviceView_, isoCoordsShapeFcnHostView_);
 
-  auto scsFaceNodeMapHostView =
-    Kokkos::create_mirror(scsFaceNodeMapDeviceView_);
+  Kokkos::View<int[12][4], Kokkos::HostSpace> scsFaceNodeMapHostView;
```
crtrott, line 74: this does not initialize the view, the pointer will be a null ptr. Needs a string as argument to the constructor.

ashesh2512: so a `Kokkos::View<int[12][4], Kokkos::HostSpace> scsFaceNodeMapHostView (<some string>)` ?

crtrott: yeah, but that will not allow for the deep_copy below because device and host view will have different layouts.

psakievich:
This is incorrect behavior. Summary from slack conversation
> So I have a hypothesis for what is going on in our code.  Currently my hypothesis is the deep copy doesn't happen but doesn't fail and the scsFaceNodeMapDeviceView_ is initialized to all zeros. This would make our source term zero which it is at the start of the simulation so everything would look fine.
However, when we create_mirror we populate  it with the correct values, but it likely exposes another bug.

This hypothesis was verified by `printf` statements inside the device kernels.

PaulMullowney:
https://github.com/Exawind/nalu-wind/blob/master/src/gcl/MeshVelocityAlg.C#L84

Shouldn't these hardcoded loop dimensions be fixed?

PR #1206, opened 2023-08-29
remove Kokkos directive

ashesh2512:
Make GCL and FSI-related united tests run with GPUs enabled


alanw0 (approved):
Removal of these ifdefs looks correct, the question is what happens when these tests run? Do they work or crash?

ashesh2512:
They crash. We expect them to be able to make them run on GPUs. I think the directives are from before we ported the FSI pathway to GPUs

PR #1210, opened 2023-09-15
Option to disable FSI ramping parameters

psakievich:
Add 3 booleans to disable the fsi ramping for split meshes with gaps between the blades. In those cases we should no longer need to ramp spatially. We should keep temporal ramping on always except for targeted tests, but this gives a clear way to turn it off.

- `enable_temporal_ramping: false`
- `enable_span_ramping: false`
- `enable_theta_ramping: false`

All of these are defaulted to `true` and they are optional parameters i.e. no input decks are backwards compatible.

### DIFF
--- a/include/aero/aero_utils/DeflectionRamping.h
+++ b/include/aero/aero_utils/DeflectionRamping.h
@@ -15,10 +15,10 @@
 namespace fsi {
 
 double KOKKOS_FORCEINLINE_FUNCTION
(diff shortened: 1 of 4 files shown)


ashesh2512 (approved):

ashesh2512:
no input decks are backwards compatible? Or they are all backward compatible, since these are optional?

psakievich:
they are all backwards compatible.  all old input decks will still run without these parameters

Issue #936, opened 2022-02-09
Coupled overset runs are broken


ashesh2512:
## Detailed description of the bug 

Recently a few of us have noticed coupled overset runs failing in nalu. Not exactly sure where the issue lies. We have a regression test that does use coupled overset runs - overset cylinder. However - 1) The location of the hole does not change in that regression test which is the scenario where others have run into issues 2) The ghosting doesn't change significantly which is the case with turbine runs. Additionally, for a stationary mesh, coupled runs still appear to do fine.

### Details to reproduce bug/build error

SHA ID of the Nalu-Wind commit you are using -
This bug is present as of the latest commit. Not quite sure when it was introduced, but this capability was thoroughly tested and worked as of the following commits - 
   Nalu-Wind GIT Commit SHA: 50ef780f1afb7cc014f4e9df364f04ca68c57d34
   Trilinos Version: 12.17-gadef041e9e
 
Details of your simulation (input file, mesh description, etc.) -
[nrel5mw.txt](https://github.com/Exawind/nalu-wind/files/8035948/nrel5mw.txt)

Stack trace of the runtime error output -
[nrel5mw.log](https://github.com/Exawind/nalu-wind/files/8035849/nrel5mw.log)

ddement:
@ashesh2512 Do you know where we can get access to the nrel5mw_overset.exo mesh used in this case?  Could you perhaps put it in the proj-shared folder on Frontier's Orion filesystem?  Also, what recipe did you use for running this case (number of nodes, GPUs, machine used, etc)?

mbkuhn:
Due to the age of this issue, and that a merged PR claims to fix it, I'll consider this issue closed.

PR #1214, opened 2023-09-19
Fixes errors in ghosting update for some overset mesh problems.

ddement:
Instead of trying to modify the ghosting for overset problems with relative motion, we now just rebuild the ghosting from scratch every time update_ghosting is called. This appears to fix errors such as those reported in #936


ashesh2512:
@ddement So we are essentially recreating the ghosting object every time step. Is that accurate?

ashesh2512 (approved):

ddement:
@ashesh2512 Exactly.  I'm sure there's a more efficient way to incrementally update it like we were attempting to do previously, but this approach should always produce a correct ghosting.  If this turns out to be a significant cost, we can readdress optimizing it at a later point.

alanw0:
That's right. At each time step there is a list of elements that need to be ghosted. The list for the next time step includes a lot of the same elements, but some new elements get added, and some get dropped. We were trying to just incrementally update the list, rather than completely dropping the old list and adding the new list. (I worked on this years ago with Stefan.) Apparently the incremental update wasn't correct, so dropping all of the old ones and then ghosting the new list is easier to get correct. Hopefully the performance difference shouldn't be too great. Like David says, we can revisit this if there is a big performance difference.

alanw0 (approved):

ashesh2512 (approved):

ashesh2512:
@gantech Could you please try the changes in this PR, for your simulations?

alanw0:
We think this PR is ready to merge. I'm only hesitating because I don't want to breathe on anything until the challenge problem is finished.

ashesh2512:
> We think this PR is ready to merge. I'm only hesitating because I don't want to breathe on anything until the challenge problem is finished.

I think that is a good life choice considering recent events

psakievich (approved):
I've stalled merging everything until we get it done as well. There are several pending PR's that are ready to merge.

PR #1226, opened 2023-10-23
Add option to dump mesh on failed jacobian check

psakievich:


psakievich:
@alanw0 will this be parallel consistent?  Looking at it now, I think not.

psakievich:
I think I fixed it.

alanw0:
@psakievich I'm having trouble figuring out if it's parallel consistent. You want to call provide_output on every mpi rank, but it looks like impl_negative_jacobian_check only throws on ranks where it finds a negative volume. I'm having trouble seeing where there's an allreduce to make the result of impl_negative_jacobian_check global. If that's in there, then it's all good.

psakievich:
@alanw0 ah you are right. Still need the all-reduce

gantech (approved):

alanw0 (commented):
Looks good.

PR #1208, opened 2023-09-05
WIP FSI Sync

PaulMullowney:
Adding syncs and modifies to appropriates methods in FSI

PR #1227, opened 2023-10-24
Correctly set current_coordinates for restart

psakievich:
When I refactored this in #1208 I left the behavior of not updating current_coordinates at this point. Seems like we should though. Let's see if this fixes the restart issue @gantech


psakievich:
Looks like this is a redundant call that should just be deleted all together. @gantech is confirming now.

psakievich:
Closing due to #1228

PR #1235, opened 2024-02-01
Updating IBlank

psakievich:
I am not sure update shared is the right operation. It seems like if a cell is detected as fringe anywhere it should be fringe even if it is not the owner.  In this case `parallel_max` makes more sense to me. I have not tested this.  

This could explain the behavior https://github.com/Exawind/exawind-driver/issues/65

### DIFF
--- a/src/overset/TiogaSTKIface.C
+++ b/src/overset/TiogaSTKIface.C
@@ -178,7 +178,7 @@ TiogaSTKIface::post_connectivity_work(const bool isDecoupled)
   ScalarIntFieldType* ibf =
     meta_.get_field<ScalarIntFieldType>(stk::topology::NODE_RANK, "iblank");
   std::vector<const stk::mesh::FieldBase*> pvec{ibf};
-  stk::mesh::copy_owned_to_shared(bulk_, pvec);
+  stk::mesh::parallel_max(bulk_, {ibf});
 
   post_connectivity_sync();
 


alanw0 (commented):
For entities that are shared among multiple MPI ranks, copy_owned_to_shared over-writes the non-owners' values with the owners' values.
parallel_max will over-write all ranks' values with the max of all ranks' values.

mbkuhn:
After getting into the issues with Ilker and Wyatt, this change does get at the problem. However, to fix the iblank array, it need to be "parallel_min" instead of "parallel_max", because the value of -1 needs to trump the other values. Though this fixes the iblank array for nalu-wind, the data pathways still don't exist for the nodes whose iblank values have been corrected. This leaves them uninitialized, preventing nalu-wind from going forward (the code crashes). 

We concluded that the underlying issue isn't this part of nalu-wind, but it is in TIOGA itself (it should get the same answer for a node that is shared by two different chunks). Because of that, it makes the most sense to address this bug through TIOGA directly, and Ilker has some ideas for doing that.

itopcuoglu:
As Michael mentioned, the parallel_min operation corrects the iblank field, and the artifacts near the outer boundary disappear in the iblank plot. Concurrently, this causes the hybrid solver to crash. This is likely because while the modification fixes the iblank field, no field value is interpolated to that cell in Tioga since it was classified as a field point (iblank=1) and not a receptor/fringe point (iblank=-1). So the fix has to happen inside Tioga, and it has to ensure that duplicate nodes at STK mesh partition boundaries have the same iblank values.

PR #1238, opened 2024-02-14
Enforcing consistency of iblanks at shared nodes

itopcuoglu:


psakievich (approved):
This seems like a reasonable and quick fix. Longer term I would like to see if there is a more efficient way to do this than copying back and forth between two identically sized fields via loops. But let's save that until we know what our next moves are with tioga.

itopcuoglu:
Ideally, we should just use the STK data structures to do anything related to iblank fields.

PR #1254, opened 2024-04-05
Default value of SDRWallFactor: 1 -> 10

BumseokLee:
The SDRWallFactor is a constant for the wall boundary condition of the omega in the SST turbulence model.

However, there are some uncertainties about this in Nalu-Wind. The original SST model by Menter uses 10 for the SDRWallFactor. Currently, in Nalu-Wind, users can specify the SDRWallFactor in the input yaml file, and the default value is 1 if not explicitly specified. Many users run RANS simulations with the default value. Some reg test cases in the repository used 0.625. Marc orignally performed the validation of the NASA’s 2D Zero Pressure Gradient (ZPG) Flat Plate Verification Case with SDRWallFactor of 10. 

The wall boundary condition plays a crucial role in transition simulations because the transition model triggers laminar-turbulent transition based on flow quantities close to the middle of the boundary layer, which is very near the wall.  

For this reason, I tested different values of SDRWallFactor for the NASA TMR’s 2D ZPG flat plate and NACA 0012 airfoil cases and adjusted the default value to 10 based on the test results, which I believe is the correct value.

 
I've attached some results for the ZPG flat pate cases below

Please refer to the attached file for more detailed analysis. Any discussion about the is welcomed!
 
[SDR_Wall_factor.pdf](https://github.com/Exawind/nalu-wind/files/14886922/SDR_Wall_factor.pdf)

### DIFF
--- a/src/SolutionOptions.C
+++ b/src/SolutionOptions.C
@@ -653,7 +653,7 @@ SolutionOptions::initialize_turbulence_constants()
   turbModelConstantMap_[TM_cmuCs] = 0.17;
   turbModelConstantMap_[TM_Cw] = 0.325;
   turbModelConstantMap_[TM_CbTwo] = 0.35;
-  turbModelConstantMap_[TM_SDRWallFactor] = 1.0;
+  turbModelConstantMap_[TM_SDRWallFactor] = 10.0;
   turbModelConstantMap_[TM_zCV] = 0.5;
   turbModelConstantMap_[TM_ci] = 0.9;
   turbModelConstantMap_[TM_elog] = 9.8;


gantech (approved):

psakievich (approved):

marchdf:
Fine by me. I think the only reason we didn't do this way back in the day is that it would change everyone's cases/runs and we seemed averse to doing that. Seems like going to 10 helps drags but hurts lift (based on comparing to NASA codes).

marchdf (approved):

psakievich:
What value are the NASA codes using?

BumseokLee:
@marchdf Did you mean the lift coefficient of NACA0012 case that I included in the file? The lift coefficients from 0.625 and 10 are almost identical. There are only minor differences in the drag coefficient at AoA=15deg, although the trends of the omega residual are significantly different.  

@psakievich NASA's codes use 10. I've attached the SST model page from NASA Langley's Turbulence Modeling Resource Website here:  https://turbmodels.larc.nasa.gov/sst.html .

PR #1255, opened 2024-04-12
Updates to use latest Trilinos develop branch

jrood-nrel:


src/Realm.C
```diff
@@ -3708,7 +3708,7 @@ Realm::check_job(bool get_node_count)
   // set number of nodes, check job run size
   if (get_node_count) {
     size_t localNodeCount =
-      ioBroker_->get_input_io_region()->get_property("node_count").get_int();
```
alanw0, line 3711: Yep, sorry about that change. (That was an artifact of our conversion from teuchos::rcp to std::shared_ptr.)

src/Realm.C
```diff
@@ -3792,10 +3792,10 @@ Realm::check_job(bool get_node_count)
     unsigned nfields = fields.size();
     for (unsigned ifld = 0; ifld < nfields; ++ifld) {
       stk::mesh::FieldBase* field = fields[ifld];
-      unsigned fszNode = field->max_size(stk::topology::NODE_RANK);
-      unsigned fszEdge = field->max_size(stk::topology::EDGE_RANK);
-      unsigned fszFace = field->max_size(stk::topology::FACE_RANK);
-      unsigned fszElem = field->max_size(stk::topology::ELEM_RANK);
```
alanw0, line 3798: Yep, that argument was deprecated because fields can only be on 1 'rank', so the argument doesn't really make sense.

alanw0 (approved):
Looks good, a bunch of search/replace and add prefix...
The other API changes you encountered went through a deprecation cycle where they would have been a warning for at least 6 weeks before they were actually changed in a way that causes an error. Sometimes we debate whether we should lengthen the deprecation window to be more gentle on users that don't update as often...

psakievich (approved):

psakievich:
Just a reminder that this will break all of our default builds so we will need to add versions and update dependencies in spack right away.

jrood-nrel:
Yeah I will get other stuff in place before merging this. I'm mostly curious if it will help me with the segfaults on Aurora.

PR #1265, opened 2024-06-10
Multiphase stabilization and mass-momentum consistency updates

wjhorne:
This PR contains all of the multiphase work that was used to successfully run hybrid cases of waves passing over solid bodies. The main changes include 

- mass-momentum consistency via modifying the mass flux when vof is used. This was found to be a substantial improvement for moving interfaces over long runs.
- upwinding stabilization with switching in the momentum equation at interfaces. This was a critical addition to ensure stability for high CFLs and harsh flow events that are encountered when waves pass over bodies.


psakievich:
I will try to take a look but I'm not sure how timely I can be.  @BumseokLee would you also be willing to take a look? @mbkuhn your input would also be helpful.  Also @rcknaus if you have some spare time your input is always appreciated.  We have a P/T you can charge for your time if needed.

BumseokLee:
Hi Phil, 
I am on a business trip until July 6. Also, I don’t have experiences with the multiphase stabilization or mass-momentum consistency in Nalu-Wind. I think it would be better to assign someone else to the work. 
Bumseok

psakievich:
@wjhorne will you add some regression tests that utilize the user functions you've added?

include/edge_kernels/VOFAdvectionEdgeAlg.h
```diff
@@ -39,7 +39,11 @@ class VOFAdvectionEdgeAlg : public AssembleEdgeSolverAlgorithm
   unsigned dqdx_{stk::mesh::InvalidOrdinal};
   unsigned edgeAreaVec_{stk::mesh::InvalidOrdinal};
   unsigned massFlowRate_{stk::mesh::InvalidOrdinal};
+  unsigned massVofBalancedFlowRate_{stk::mesh::InvalidOrdinal};
   unsigned density_{stk::mesh::InvalidOrdinal};
+
+  double density_liquid_;
```
marchdf, line 45: Liquid and gas makes me think this is a 2-phase flow and not arbitrary "multi" phase? Am I thinking about this right?

include/user_functions/DropletVelocityAuxFunction.h
```diff
@@ -0,0 +1,42 @@
...
+
+class DropletVelocityAuxFunction : public AuxFunction
+{
+public:
+  DropletVelocityAuxFunction(const unsigned beginPos, const unsigned endPos);
+
+  virtual ~DropletVelocityAuxFunction() {}
+
+  using AuxFunction::do_evaluate;
+  virtual void do_evaluate(
```
marchdf, line 28: Kind of unfortunate we call this with 2 verbs ;) Probably not relevant to this PR.

src/LowMachEquationSystem.C
```diff
@@ -165,6 +165,9 @@
 #include <user_functions/KovasznayVelocityAuxFunction.h>
 #include <user_functions/KovasznayPressureAuxFunction.h>
 
+#include <user_functions/DropletVelocityAuxFunction.h>
```
marchdf, line 168: Any reason for the ordering here? Want to group the multiphase aux functions together?

src/VolumeOfFluidEquationSystem.C
```diff
@@ -270,12 +281,12 @@ VolumeOfFluidEquationSystem::register_inflow_bc(
 
   stk::mesh::MetaData& meta_data = realm_.meta_data();
 
-  // register boundary data; gamma_bc
+  // register boundary data; vof_bc
```
marchdf, line 284: Should we just remove these comments (and the following)? The code is pretty explicit I think.

src/VolumeOfFluidEquationSystem.C
```diff
@@ -487,14 +495,22 @@ VolumeOfFluidEquationSystem::register_initial_condition_fcn(
                                  TurbulenceModel::SST_AMS)
                                   ? true
                                   : false;
+        VectorFieldType* velocity_ = realm_.meta_data().get_field<double>(
+          stk::topology::NODE_RANK, "velocity");
         ScalarFieldType* density_ = realm_.meta_data().get_field<double>(
           stk::topology::NODE_RANK, "density");
-        std::vector<double> userSpec(1);
+        std::vector<double> userSpec(3);
```
marchdf, line 502: we could initialize this with an initializer list instead of like this.

src/edge_kernels/MomentumEdgeSolverAlg.C
```diff
@@ -130,8 +144,8 @@ MomentumEdgeSolverAlg::execute()
         }
       }
 
-      const DblType pecfac = pecletFactor.get(edge, 0);
-      const DblType om_pecfac = 1.0 - pecfac;
+      DblType pecfac = pecletFactor.get(edge, 0);
```
marchdf, line 147: could we move this down a bit? Closer to your upwinding switch comment since that's closer to where it is used/needed?

src/edge_kernels/MomentumEdgeSolverAlg.C
```diff
@@ -146,12 +160,36 @@ MomentumEdgeSolverAlg::execute()
         }
       }
 
+      // Upwinding switch for multiphase cases.
+      // Factors determined by ensuring full upwinding
+      // at interfaces with interface
+      // widths that are ~2 cells thick
+      DblType density_upwinding_factor = 1.0;
+      DblType alphaUpw_w_vof = alphaUpw;
+      DblType om_alphaUpw_w_vof = 1.0 - alphaUpw_w_vof;
```
marchdf, line 169: We get so many questions about upwinding in nalu-wind and all the settings. It feels like we should document these extra modifications?

src/edge_kernels/MomentumEdgeSolverAlg.C
```diff
@@ -196,9 +234,9 @@ MomentumEdgeSolverAlg::execute()
         const DblType uiIp = 0.5 * (vel.get(nodeR, i) + vel.get(nodeL, i));
 
         // Upwind contribution
-        const DblType uiUpw = (mdot > 0.0)
-                                ? (alphaUpw * uIpL[i] + om_alphaUpw * uiIp)
-                                : (alphaUpw * uIpR[i] + om_alphaUpw * uiIp);
+        const DblType uiUpw =
+          (mdot > 0.0) ? (alphaUpw_w_vof * uIpL[i] + om_alphaUpw_w_vof * uiIp)
```
marchdf, line 238: the suffix `_vof` is confusing to me. Can we keep it as it is and just modify the value if vof is on? My thinking is that if I read this code I am thinking that it is vof specific.

src/user_functions/FlatDensityAuxFunction.C
```diff
@@ -0,0 +1,57 @@
...
+  const unsigned /*endPos*/) const
+{
+  for (unsigned p = 0; p < numPoints; ++p) {
+
+    const double x = coords[0];
+    const double y = coords[1];
+    const double z = coords[2];
+    const double interface_thickness = 0.015;
+
+    fieldPtr[0] = 0.0;
```
marchdf, line 45: probably don't need this one, Just make the line after than a plain `=`?

src/user_functions/FlatDensityAuxFunction.C
```diff
@@ -0,0 +1,57 @@
...
+    fieldPtr[0] += -0.5 * (std::erf(y / interface_thickness) + 1.0) + 1.0;
+
+    // air-water
+    fieldPtr[0] = 1000.0 * fieldPtr[0] + (1.0 - fieldPtr[0]);
```
marchdf, line 49: this is confusing? Maybe make the first assignement a variable and then use it here?

src/user_functions/SloshingTankPressureAuxFunction.C
```diff
@@ -0,0 +1,99 @@
...
+  const unsigned /*endPos*/) const
+{
+  for (unsigned p = 0; p < numPoints; ++p) {
+
+    const double x = coords[0];
+    const double y = coords[1];
+    const double z0 = coords[2];
+    const double z1 = domain_height_;
+
+    // These need to come from elsewhere
```
marchdf, line 63: yeah it should. Before we get into a mess like amr-wind and density.

marchdf (commented):
Great work! A bunch of minor comments.

include/edge_kernels/VOFAdvectionEdgeAlg.h
psakievich: I think liquid and gas are the only phases we will be equipped to deal with in a suite of physics nalu can handle

src/edge_kernels/MomentumEdgeSolverAlg.C
psakievich: Yeah I agree. It seems like the vof changes are to modifying the general upwinding?  If there is not a second set of behavior that requires these variables for vof then I think it makes sense to just remove these variables and modify the original as needed when VOF is active.

src/edge_kernels/MomentumEdgeSolverAlg.C
psakievich: Along those lines. I would be in favor of adding a warning or throwing an error if the upwinding is put in a state that you have discovered is unstable. I'm assuming pure central will not work with VOF? Or is you modification below sufficient to ensure that the local upwinding is sufficient?  Not knowing is also an acceptable answer, but I would like a clear guide line for the best practice, preferably  in the code that is checked at runtime.

src/user_functions/SloshingTankPressureAuxFunction.C
psakievich: Constants we should stash during construction.  There is more freedom on the API there and we can glob them into a type specific struct.

include/user_functions/DropletVelocityAuxFunction.h
wjhorne: It is strange for sure. There are a few places where patterns like this show up. There is also stuff like EquationSystem.C vs EquationSystems.C which can be confusing when working in the code. Likely worth a pass through the code to fix them down the road.

src/VolumeOfFluidEquationSystem.C
wjhorne: I purged a bunch of these comments. We should probably do a pass through the code base to get more of them since this is quite common throughout the code.

src/VolumeOfFluidEquationSystem.C
wjhorne: I agree and changed this over

src/edge_kernels/MomentumEdgeSolverAlg.C
wjhorne: The documentation should definitely reflect the changes here. Our general finding is that pure central at the multiphase interface does not work well for VOF when you have high density ratios and strong flow dynamics. 

The modifications here respect the users settings except explicitly at the multiphase interface when VOF is turned on. I think, at best, an appropriate warning would be to inform the user that by enabling VOF you have implicitly turned on upwinding at the interface.

src/edge_kernels/MomentumEdgeSolverAlg.C
wjhorne: It ends up being identical to alphaUpw and om_alphaUpw if a non-VOF case is run, or you are not at the interface with VOF. I'll change over the naming.

src/user_functions/SloshingTankPressureAuxFunction.C
wjhorne: This user function was added during some of the testing of the pressure behavior of some fully AMR-wind/Nalu-Wind coupled cases. Specifying the pressure is not necessary for this case. I will remove it completely

wjhorne:
> @wjhorne will you add some regression tests that utilize the user functions you've added?

We have fully-coupled tests that utilize the user functions. The only exception is the droplet case, which is also covered by VOFDroplet in our regression tests.

include/user_functions/FlatDensityAuxFunction.h
```diff
@@ -0,0 +1,42 @@
+// Copyright 2017 National Technology & Engineering Solutions of Sandia, LLC
```
mbkuhn, line 1: I'm pretty sure this was created when we were toying with using a perturbational pressure approach that needed a reference density field. None of the cases (that I'm adding as reg tests to exawind-driver) use this user_function.

include/user_functions/WaterLevelDensityAuxFunction.h
mbkuhn, line 1: same goes for this one too.

mbkuhn:
I'm getting a weird error with this PR. I'm using it to run [these reg tests](https://github.com/Exawind/exawind-driver/pull/73/files). When tpetra is used for the velocity solver, the reg tests run fine. But with hypre instead, I get the error "terminating due to uncaught exception of type std::runtime_error: HypreLinearSystem::sumInto not yet implemented for (NON) overset constaint algorithms. Exiting." What makes this strange is the other (single-phase) reg tests use hypre for velocity with no issues, and these two-phase cases run fine with hypre for the volume_of_fluid equation.

mbkuhn:
> I'm getting a weird error with this PR...

This doesn't have to do with any of the changes in the PR. Running one of the cases with master (and turning off the unavailable user function sloshing_tank) still leads to the error. It seems to be a problem with the velocity + volume_of_fluid equations together, not separately (because the zalesak reg test runs fine).

psakievich:
@mbkuhndo those single phase tests use overset? Here is where the error comes from: https://github.com/Exawind/nalu-wind/blob/53f493954891937db39876452af46dbf3ce3811c/src/HypreLinearSystem.C#L2399-L2403

 looking at the calls of sumInto it seems it would hit with the computing overset pressure call for this case here: https://github.com/Exawind/nalu-wind/blob/53f493954891937db39876452af46dbf3ce3811c/src/overset/AssembleOversetPressureAlgorithm.C#L184-L185

Also worth checking if the other cases are using a segregated velocity solve. It may use a different path. Looks like the HypreUVW solver has this as a null function. https://github.com/Exawind/nalu-wind/blob/53f493954891937db39876452af46dbf3ce3811c/include/HypreUVWLinearSystem.h#L171-L193

I don't have a great mental model of how all these go together though.

mbkuhn:
Update: the problem is not the VOF algorithm or even overset. It's the buoyancy term.

None of the other overset reg tests contain the buoyancy term, and typical runs with nalu-wind tend to employ the buoyancy_boussinesq term, which has no issue with hypre solvers. To check, I ran the problem exawind-driver PR reg tests with the buoyancy term turned off (no error). I looked into the nalu-wind reg tests to find tests with the plain "buoyancy" source term, I replaced the tpetra solver with hypre for velocity, and these tests generated the same sumInto error. These were the VOFDroplet and periodic3dEdge reg tests. All of these checks were performed with the master branch of nalu-wind, not this PR.

Should I make an issue for this?

wjhorne:
@psakievich I believe with the latest push it should be good to go for a merge. Let me know what you think

psakievich (approved):

reg_tests/test_files/VOFDroplet/VOFDroplet.norm.gold
mbkuhn, line 1: The big difference in gold norms is not because of changing the VOFDroplet setup (i.e., using input arguments instead of hard-coded parameters). It's because of using a much smaller interface thickness, which Wyatt chose -- I'm assuming with the intent of stressing the new stabilization. With the new stabilization but the old thickness, the norms are very close to the originals: 1.768e-5, 6.628e-6, 3.367e-6.

mbkuhn:
> We have fully-coupled tests that utilize the user functions. The only exception is the droplet case, which is also covered by VOFDroplet in our regression tests.

SloshingTank is/will be covered by reg tests for exawind-driver; I added a VOFInertialDroplet case so the droplet velocity user function would be covered.

src/edge_kernels/MomentumEdgeSolverAlg.C
mbkuhn: Warning has been added.

src/edge_kernels/MomentumEdgeSolverAlg.C
mbkuhn: done

reg_tests/test_files/VOFDroplet/VOFDroplet.norm.gold
psakievich: Thanks for the update. This is good to know.

For better or worse, we don't really track/commit the norms any more due to the wide differences we see between platforms/compilers/architectures.  The span is too big for comfort for these general integrated norms.  I suppose we could do it with the CI container now as a constant platform for evaluation, but the infrastructure for that is not currently in place.

mbkuhn: Good to know! Thanks, I wasn't sure.

PR #1266, opened 2024-06-13
ADM Turbine Level Search

gyalla:
Hi Phil, 

The question is regarding the line 
`auto points = pointCentroid_.template view<ActuatorFixedMemSpace>();`
in ActuatorBulk.C and why it does not work when replacing pointCentroid_ with hubLocations_ as in
`auto points = hubLocations_.template view<ActuatorFixedMemSpace>();`

Best,
Gopal


src/aero/actuator/ActuatorBulkFAST.C
```diff
@@ -242,6 +245,63 @@ ActuatorBulkFAST::init_epsilon(const ActuatorMetaFAST& actMeta)
   searchRadius_.sync_host();
 }
 
+void
+ActuatorBulkFAST::stk_turbine_search(
+  const ActuatorMeta& actMeta, stk::mesh::BulkData& stkBulk, bool onlyFine /* = false */)
+{
+  //It seems point_centroids includes all actuator points across all turbines. TODO: Is that right?
```
psakievich, line 252: Yes that is correct

src/aero/actuator/ActuatorBulkFAST.C
```diff
@@ -242,6 +245,63 @@ ActuatorBulkFAST::init_epsilon(const ActuatorMetaFAST& actMeta)
...
+  //If so, we need to build a similar list here that will be of size numberofActuators
+  //that includes all hublocations. The search radius also needs to be of the same size 
+  
+  // create bounding sphere and element boxes based on turbine location and search radius
+  auto points = hubLocations_.template view<ActuatorFixedMemSpace>();
```
psakievich, line 257: This templated view function is only for dual views where you can get a view on host or device.  The type for ActFixScalarDouble is already a view: https://github.com/Exawind/nalu-wind/blob/83c6597c0b7afd042cb510615af286e9fae30b29/include/aero/actuator/ActuatorTypes.h#L70-L71

src/aero/actuator/ActuatorExecutorsFASTNgp.C
```diff
@@ -25,33 +25,46 @@ ActuatorLineFastNGP::ActuatorLineFastNGP(
...
+  actBulk_.zero_source_terms(stkBulk_); 
 
-  // set range policy to only operating over points owned by local fast turbine
+  //Range for Kokkos parallel-for -- set range policy to only operating over points owned by local fast turbine
   auto fastRangePolicy = actBulk_.local_range_policy();
 
+  //Interpolate velocity to actuator points. This happens before fine search? Is this from previous timestep
   RunInterpActuatorVel(actBulk_, stkBulk_);
 
+  // Add FLLC correction to velocity field. Is this term treated explicitly? Also seems like it’s from previous step
```
psakievich, line 37: Yes it is explicit and there is a lag.  The value has a heavy relaxation factor.

include/aero/actuator/ActuatorBulk.h
```diff
@@ -67,6 +68,16 @@ struct ActuatorBulk
...
   void stk_search_act_pnts(
     const ActuatorMeta& actMeta, stk::mesh::BulkData& stkBulk);
+
+  //Generic stk search
+  void stk_search(
+    const ActuatorMeta& actMeta, stk::mesh::BulkData& stkBulk, bool onlyFine = false);
+
+  //do I need this to be virtual function so the generic stk_search can call it, even though 
+  //the turbine search only makes sense in the ActuatorBulkFast context as implemented? 
+  virtual void stk_turbine_search(
```
psakievich, line 78: No I wouldn't make this a virtual function and I wouldn't define it here. I think it would make more sense to have `stk_search` be virtual and then implement the conditionals and turbine search in the child classes that are actually turbines.

include/aero/actuator/ActuatorGenericTurbineSearchFunctor.h
```diff
@@ -0,0 +1,147 @@
...
+      }
+  }
+  }
+
+  ActuatorBulk& actBulk_;
+  stk::mesh::BulkData& stkBulk_;
+  VectorFieldType* coordinates_;
+  VectorFieldType* actuatorSource_;
+  ScalarFieldType* dualNodalVolume_;
+  functor innerLoopFunctor_;
```
psakievich, line 140: `const size_t innerLoopExtent_;`
Add this here for populating on construction

include/aero/actuator/ActuatorGenericTurbineSearchFunctor.h
```diff
@@ -0,0 +1,147 @@
...
+        stk::topology::NODE_RANK, "coordinates")),
+      actuatorSource_(stkBulk_.mesh_meta_data().template get_field<double>(
+        stk::topology::NODE_RANK, "actuator_source")),
+      dualNodalVolume_(stkBulk_.mesh_meta_data().template get_field<double>(
+        stk::topology::NODE_RANK, "dual_nodal_volume")),
+      innerLoopFunctor_(innerLoopFunctor)
+  {
+    actBulk_.coarseSearchElemIds_.sync_host();
+    actBulk_.coarseSearchPointIds_.sync_host();
+    innerLoopFunctor_.preloop();
```
psakievich, line 69: `innerLoopExtent = actBulk.singlePointCoarseSearch ? 1 : /*actuatorBulk points extent*/;`
Then add in the inner loop inside the turbine

include/aero/actuator/ActuatorFunctorsFAST.h
```diff
@@ -164,6 +164,14 @@ using ActFastSpreadForceWhProjection = GenericLoopOverCoarseSearchResults<
   ActuatorBulkFAST,
   ActFastSpreadForceWhProjInnerLoop>;
 
+using ActFastComputeThrustTurbineSearch = GenericLoopOverCoarseTurbineSearchResults<
+  ActuatorBulkFAST,
+  ActFastComputeThrustInnerLoop>;
+
+using ActFastSpreadForceWhProjectionTurbineSearch = GenericLoopOverCoarseTurbineSearchResults<
```
psakievich, line 171: In this function in amr-wind there is a distance check that is done prior to computing the exponentials.  In nalu-wind we used the search to filter this, but we probably want to add in the distance check here since we will now have the entire box surrounding the turbine.

include/aero/actuator/ActuatorGenericSearchFunctor.h
```diff
@@ -124,9 +126,19 @@ struct GenericLoopOverCoarseSearchResults
...
+      //    innerLoopFunctor_(actPtInd, nodeCoords, sourceTerm, dual_vol, scvIp[nIp]);
+      //  }
+      //
+      if (actBulk_.singlePointCoarseSearch_) { 
+        innerLoopFunctor_(pointId, nodeCoords, sourceTerm, dual_vol, scvIp[nIp]);
+      } else {
+        for (int actPtInd = 0; actPtInd < actBulk_.pointCentroid_.extent(0); actPtInd ++){
+          innerLoopFunctor_(actPtInd, nodeCoords, sourceTerm, dual_vol, scvIp[nIp]);
+        }
+      }
```
gyalla, line 140: I'm not sure the innerLoopExtent_ idea will work because if singlePointCoarseSearch_=true, the inner functor requires the pointId associated with the index passed as an argument to the generic loop functor. However, if singlePointCoarseSearch_= false, then we need to loop over all actuator points. Is this simple if-else statement appropriate?

src/aero/actuator/ActuatorFunctors.C
```diff
@@ -91,15 +91,21 @@ SpreadForceInnerLoop::operator()(
...
-    actuator_utils::Gaussian_projection(3, &distance[0], epsilon.data());
-
-  for (int j = 0; j < 3; j++) {
-    projectedForce[j] = gauss * pointForce(j);
-  }
+  //Check distance between actuator point and element centroid. Only needed if singlePointCoarseSearch_==False
+  //auto epsilonRadius = 
+  //  Kokkos::subview(actBulk_.searchRadius_.view_host(), pointId, Kokkos::ALL);
+  auto epsilonRadius = actBulk_.searchRadius_.h_view(pointId);
+  if (std::sqrt(distance[0]*distance[0] + distance[1]*distance[1] + distance[2]*distance[2]) < epsilonRadius) {
```
gyalla, line 98: I've added a check based on the distance from the actuator point to the element centroid before evaluating the exponential, which uses the predefined values in searchRadius_. I think this check is only needed if singlePointCoarseSearch_=false, right? If so, I can add another conditional to the if statement

include/aero/actuator/ActuatorGenericTurbineSearchFunctor.h
gyalla: See my comment below

include/aero/actuator/ActuatorFunctorsFAST.h
gyalla: See my comment below

include/aero/actuator/ActuatorGenericSearchFunctor.h
psakievich: Yeah I think this is an appropriate use case for if-else.

src/aero/actuator/ActuatorFunctors.C
psakievich: It shouldn't make a difference. The search radius is epsilon for the old case. I think profiling is the only way to know if this would slow the old case down.

marchdf:
Hi @gyalla, do you think this is close to being able to merge? It looks like you addressed all the outstanding comments.

gyalla:
> Hi @gyalla, do you think this is close to being able to merge? It looks like you addressed all the outstanding comments.

Hi @marchdf , thanks for following up on this. I still have a few to-do's on this PR so I can close it for now and re-open it once I check those off. I do need the NREL5MW actuator line/disk regression tests working before I can confidently validate this code. I see you have a PR related to those tests -- does that mean they are close to working again?

marchdf:
Oh if you still have things going on, feel free to keep it open. 

Re: the tests. I do have those working! But I need to move the pinned dev openfast commit we have in exawind-manager. Then I will merge my PR. Maybe I can do that today.

marchdf:
@gyalla got everything merged in! So those tests should pass now.

PR #1267, opened 2024-06-13
Removing the set_resolutions option to resolve mandatory receptor issues

itopcuoglu:


psakievich:
Are you sure this is never needed?

itopcuoglu:
None of the past regression tests use the `set_resolutions: no` option. We could also leave it in place, and make the code display a warning to say something like "WARNING: Nodal resolutions calculated by Tioga are not communicated at shared nodes. Use the default option "set_resolutions: yes" to ensure consistency at shared nodes".

psakievich:
Seems reasonable to me then.

src/overset/TiogaOptions.C
```diff
@@ -41,11 +38,6 @@ TiogaOptions::load(const YAML::Node& node)
   if (node["node_resolution_multiplier"]) {
     nodeResMult_ = node["node_resolution_multiplier"].as<double>();
   }
-
-  if (node["adjust_mandatory_fringe_resolutions"]) {
```
psakievich, line 45: Should we delete the functions associated with this option as well?

itopcuoglu: Yes, that would be more proper. Let me get that sorted out.

itopcuoglu: I removed the boolean flag adjustResolutionsForFringes_, the function that checks it, and the flag checks themselves.

psakievich (approved):

PR #1280, opened 2024-08-06
Balanced buoyancy

wjhorne:
@mbkuhn These are the balanced buoyancy changes. I still need to evaluate more solid body cases to ensure that it has fixed everything we saw, but so far it has been good in my testing.

### DIFF
--- a/include/LowMachEquationSystem.h
+++ b/include/LowMachEquationSystem.h
@@ -18,6 +18,7 @@
 #include "AMSAlgDriver.h"
 
 #include "ngp_algorithms/NodalGradAlgDriver.h"
+#include "ngp_algorithms/NodalBuoyancyAlgDriver.h"
 #include "ngp_algorithms/WallFricVelAlgDriver.h"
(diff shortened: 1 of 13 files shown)


mbkuhn:
@wjhorne I got it to pass CI, but we should still make sure that the reg tests pass properly and also should turn on the new feature in one of the reg tests, probably VOFDroplet

mbkuhn:
Reg tests look good. I added the balanced buoyancy option to the VOFDroplet reg test and confirmed that it gets a different result. I will check this with the hybrid cases, and it should be good to go, though we did want another nalu-wind developer to sign off on it as well.

wjhorne:
@mbkuhn I will discuss this a bit further, but I introduced a new weighting that should give you much more improvement. Feel free to give it a run if you would like to see the difference. I will also introduce a mask later today that will remove the fixed wall issue I noted yesterday.

mbkuhn (approved):

wjhorne:
@psakievich Do you know anyone that could review this to get it in? Our testing looks good for it

PR #1293, opened 2024-09-09
Move VOF sharpening and diffusion coefficients to be user modifiable

wjhorne:
We have found that it is desirable to be able to tune some of the coefficients of VOF, mostly for testing + verification purposes. This PR makes the two most influential coefficients modifiable in a user's input file. The values are defaulted to safe values that have been used in all of our showcases so far. @mbkuhn @psakievich


mbkuhn (approved):
straightforward changes

Issue #1289, opened 2024-08-26
Many of Nalu-Wind reg tests are not actively being run


mbkuhn:
The dashboard (and general usage of Nalu-Wind) uses hypre solvers instead of trilinos solvers. However, the majority of the Nalu-Wind reg tests still employ trilinos solvers, and these are not being run during the nightly tests.

Let's make an effort to convert these tests to using hypre and thus include them. After converting the trilinos-based input arguments to use hypre, these cases should be checked by either comparing the norms and ensuring that the diffs are reasonably small or by looking at the actual flow physics to ensure things are running properly. Something to be aware of: not all source terms are hypre-compatible, so some coding may be needed to add updated kernels, like in this merged PR: https://github.com/Exawind/nalu-wind/pull/1274

Below are all the reg tests with tpetra included (grep -linr -m 1 "tpetra" reg_tests/test_files/*, then the ones with "Trilinos" in the test name removed):

- [ ] ActLineSimpleFLLC/ActLineSimpleFLLC.yaml
- [ ] ActLineSimpleNGP/ActLineSimpleNGP.yaml
- [x] BLTFlatPlateT3A/BLTFlatPlateT3A.yaml
- [ ] BoussinesqNonIso/BoussinesqNonIso_R0.yaml
- [ ] BoussinesqNonIso/BoussinesqNonIso_R1.yaml
- [ ] BoussinesqNonIso/BoussinesqNonIso_R2.yaml
- [ ] IDDESPeriodicHillEdge/IDDESPeriodicHillEdge.yaml
- [ ] KEChannelEdge/KEChannelEdge.yaml
- [ ] KOChannelEdge/KOChannelEdge.yaml
- [ ] MeshMotionInterior/MeshMotionInterior.yaml
- [ ] SSTAMSChannelEdge/SSTAMSChannelEdge_rst.yaml
- [ ] SSTAMSChannelEdge/SSTAMSChannelEdge.yaml
- [ ] SSTAMSOversetRotCylinder/SSTAMSOversetRotCylinder_rst.yaml
- [ ] SSTAMSOversetRotCylinder/SSTAMSOversetRotCylinder.yaml
- [ ] SSTChannelEdge/SSTChannelEdge.yaml
- [ ] SSTLRChannelEdge/SSTLRChannelEdge.yaml
- [ ] SSTPeriodicHillEdge/SSTPeriodicHillEdge.yaml
[...]

PR #1298, opened 2024-09-23
Use hypre for convTaylorVortex case

marchdf:
This should fix one of the dashboard issues.

### DIFF
--- a/reg_tests/CTestList.cmake
+++ b/reg_tests/CTestList.cmake
@@ -264,7 +264,7 @@ if(NOT ENABLE_CUDA AND NOT ENABLE_ROCM)
   #=============================================================================
   # Comparing solution norm tests
   #=============================================================================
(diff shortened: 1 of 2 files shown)


marchdf:
Linked to #1289

jrood-nrel (commented):
I think you need these https://github.com/Exawind/exawind-driver/blob/1f6c2aa931b9c01412da9c686d4224f0ce2a2614/test/test_files/flat-surface/flat-surface-nalu.yaml#L15-L16

marchdf:
Probably. I grabbed the section from ablNeutralNGPHypre.yaml. That file doesn't have these. Should I add them there too?

marchdf:
nvm the abl thing I refered to above has: `hypre_cfg_file: ../../hypre_settings/hypre_blade_resolved.yaml`, which contains those settings.

jrood-nrel (approved):

PR #1301, opened 2024-09-23
Small fix to wall distance unit test on GPU

marchdf:


marchdf:
This doesn't totally fix the issues we are seeing but it should be merged anyway.

PR #1305, opened 2024-09-23
Reg test updates: remove old transition model test, update VOFZalDisk

mbkuhn:
BLT test was set up to test a model that was never merged. A new PR with a new transition model (and accompanying reg test) is in draft.

Updating the solver to hypre in the VOFZalDisk enabled it to pass with no diffs, so that should be an easy thing to change. I think the arguments that I included allow it to be GPU and CPU compatible.


mbkuhn:
haven't tested on GPUs. but given the general lack of coverage, I'd be fine with pushing and then finding out if changes need to be made.

marchdf (approved):

PR #1307, opened 2024-09-24
Move docs to gh-pages

jrood-nrel:


marchdf (approved):

.github/workflows/docs.yml
```diff
@@ -0,0 +1,57 @@
+name: Nalu-Wind-Docs
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
```
sseraj, line 7: Should these be `master`?

jrood-nrel: Yes, thanks!

PR #1315, opened 2024-10-03
update README docs links now that website has changed

mbkuhn:
removing references to readthedocs, now pointing to exawind.github.io


marchdf:
There are a bunch of other places in the docs too I think? Could you check that real fast while you are at it?

jrood-nrel (approved):

PR #1328, opened 2024-11-15
Remove mesh velocity from VOF scaling

wjhorne:
This PR uses a compute face velocity rather than the volume flow rate to calculate the velocity scaling of all VOF sharpening and diffusion at liquid/gas interfaces. This removes erroneous dependence of such interfaces on the mesh_velocity, which appears in the volume flow rate.

### DIFF
--- a/src/edge_kernels/VOFAdvectionEdgeAlg.C
+++ b/src/edge_kernels/VOFAdvectionEdgeAlg.C
@@ -112,7 +112,7 @@ VOFAdvectionEdgeAlg::execute()
   const auto density_liquid = density_liquid_;
   const auto density_gas = density_gas_;
   const auto velocity = fieldMgr.get_field<double>(
-    get_field_ordinal(realm.meta_data(), "velocity", stk::mesh::StateNP1));
+    get_field_ordinal(realm_.meta_data(), "velocity", stk::mesh::StateNP1));
 
   run_algorithm(
     realm_.bulk_data(),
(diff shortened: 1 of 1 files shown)


wjhorne:
@mbkuhn I have to rebuild my spack environment to test this due to some recent changes to dependencies in spack. The changes here should remove mesh_velocity from the thickness of the interface in Nalu-Wind. I am in the process of fixing my spack and will test it out on a spinning patch

mbkuhn:
I'll try to test this today!

mbkuhn:
do you expect this to be equivalent in the static case? Because I'm not seeing that

mbkuhn:
The two are similar after 500 steps of the flat interface problem, but I was expecting them to be closer.
PR vs master (vof)
Screenshot 2024-11-20 at 3 57 37 PM
Screenshot 2024-11-20 at 3 58 25 PM

mbkuhn:
PR vs master (velocity_magnitude)
Screenshot 2024-11-20 at 4 02 33 PM

Screenshot 2024-11-20 at 4 02 44 PM

wjhorne:
They will not be completely the same for static cases since the volume flow rate includes pressure gradient terms in Nalu-Wind. This is due to Nalu-Wind being collocated and requiring pressure terms in the volume flow rate. Using straight velocity, as is done here, removes that pressure gradient.

mbkuhn (approved):

PR #1329, opened 2024-11-25
Oscillating mesh motion

mbkuhn:


mbkuhn:
needs a reg test.

This should be useful for testing the overset solver, but more specifically, it's set up for the 2.5D OC7 cases. Pretty straightforward stuff. For thoroughness, here is what I am copying from (in the OC7 definition document).

image

wjhorne:
Looking good so far. I agree that it needs a regression test. How will the input work for this? I am not seeing links between it and our typical yaml input

mbkuhn:
@wjhorne what kind of reg test would make sense? I am thinking it would be nalu-nalu overset, but is there a non-overset case that would make sense?

wjhorne:
I could be wrong, but I believe that there is not an option to do mesh motion via mesh distortion rather than overset unfortunately. This would mean we need an overset-overset case to get this through

wjhorne (approved):
Looks good to go !

wjhorne:
@mbkuhn  Not sure why this isn't merging given you set it to auto and I approved it. Is it due to a necessary rebase of the branch? I think we are good to go for a merge

PR #1330, opened 2024-12-02
Sync anisgauss reg test to the other act line case

marchdf:
Increase epsilon a bit to not be equal to the grid size. This is an attempt to fix the dashboard for this case.

### DIFF
--- a/reg_tests/test_files/nrel5MWactuatorLineAnisoGauss/nrel5MWactuatorLineAnisoGauss.yaml
+++ b/reg_tests/test_files/nrel5MWactuatorLineAnisoGauss/nrel5MWactuatorLineAnisoGauss.yaml
@@ -9,8 +9,8 @@ linear_solvers:
     type: hypre
     method: hypre_gmres
     preconditioner: boomerAMG
-    tolerance: 1e-5
-    max_iterations: 50
+    tolerance: 1e-12
+    max_iterations: 200
     kspace: 5
     output_level: 0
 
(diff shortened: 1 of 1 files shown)


reg_tests/test_files/nrel5MWactuatorLineAnisoGauss/nrel5MWactuatorLineAnisoGauss.yaml
```diff
@@ -131,9 +131,10 @@ realms:
         nacelle_cd: 1.0
         nacelle_area: 8.0
         air_density: 1.225
-        epsilon: [ 5.0, 2.0, 2.0 ]
+        epsilon: [ 5.0, 4.0, 4.0 ]
         epsilon_tower: 5.0
         turbine_base_pos: [ 0.0, 0.0, 0.0 ]
+        turbine_hub_pos: [ 0.0, 0.0, 90.0 ]
```
tasmith4, line 137: is this addition on purpose?  was the default not good, or are we just trying to match the other case to see what happens?

tasmith4 (approved):

mbkuhn (approved):

reg_tests/test_files/nrel5MWactuatorLineAnisoGauss/nrel5MWactuatorLineAnisoGauss.yaml
marchdf: I think both your comments are going to the same thing. Yes I did need to make this change in #1320 to make things work well. I believe I had to do that so it matched the value in the openfast files? I can't remember...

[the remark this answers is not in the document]
marchdf: see above.

marchdf:
Just to follow up. Looks like I fixed this case with this PR: https://my.cdash.org/tests/236645207

PR #1332, opened 2024-12-12
Return to vdot method for calculating velocity scale for vof sharpening while considering mesh velocity

wjhorne:
This PR fixes an issue found in Nalu-Wind's VOF formulation around solid boundaries. Essentially the compression and diffusion of the interface should be zero along interfaces or strange behavior occurs. The PR addresses this issue, while avoiding issues with mesh motion, by taking the local velocity scale to always be the minimum of the local velocity and the volume flow rate.

@mbkuhn the flag approach proved to be complex due to the flag being tied to balanced buoyancy. I found this as a simpler way to tackle things and to possibly get the best of both worlds. To recreate the flag approach, we can also just check if vdot ~ 0.0 and then zero out the local velocity scale rather than always taking the minimum.

### DIFF
--- a/src/edge_kernels/VOFAdvectionEdgeAlg.C
+++ b/src/edge_kernels/VOFAdvectionEdgeAlg.C
@@ -113,6 +113,10 @@ VOFAdvectionEdgeAlg::execute()
   const auto density_gas = density_gas_;
   const auto velocity = fieldMgr.get_field<double>(
     get_field_ordinal(realm_.meta_data(), "velocity", stk::mesh::StateNP1));
+  const std::string velocity_rtm_name =
+    realm_.has_mesh_deformation() ? "velocity_rtm" : "velocity";
+  const auto velocity_rtm = fieldMgr.get_field<double>(
+    get_field_ordinal(realm_.meta_data(), velocity_rtm_name));
 
   run_algorithm(
     realm_.bulk_data(),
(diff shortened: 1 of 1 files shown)


mbkuhn:
We've already talked about this, but I'm putting it here for the record.

Though this implementation works pretty well, we've seen it be less robust than the original volume-flux-based approach. This is probably due to instances when the velocity (within the min expression) becomes preferred, and this appears to be less stable. We think the preferred approach is to incorporate the mesh velocity into a volume-flux-based approach so we can retain the volume flux base but also address mesh motion.

mbkuhn:
we may end up needing to have something along the lines of min(volume_flux, volume_flux-mesh_velocity)

mbkuhn:
hey @wjhorne, any progress on a new approach?

wjhorne:
@mbkuhn I have the changes together. Ran into a little hiccup testing things due to a system issue, but should get it through tonight or tomorrow

wjhorne:
I have run it for 100 steps so far and the stability is good. I'll let it run longer and visualize

mbkuhn:
made it past the other crashes for the sloshing tank - cylinder case! Will check it soon that it works as expected for cases with mesh motion. Then we'll merge it in.

mbkuhn:
Ok, here is the synopsis of what I found for the sloshing tank - cylinder case. I had been evaluating if the case made it to 200 steps, which is 1 s of simulation time. With the previous 2 methods that rely on using the flow velocity, not only the mass flux, to calculate the velocity scale, the case failed between 180 and 200 steps. With the current approach in this PR, the simulation makes it to 200 steps.

Just to be sure, I tried simulating it longer, and it failed only a few timesteps later. This made me go back to the original (before we started adjusting this part of the code, commit 006da378) and check (by rerunning) if it ever made it farther than that. It did not. So, instead of this approach being the thing that "fixes" the crashing case, it is more of a modest improvement to the other ones, only making it 10-20 more steps.

I've tried to look into where the crash comes from, and it's a little unclear. There is a possibility that AMR-Wind contributes to it because with the timestep size I've been using (0.005) the CFL goes too high. However, I think this is mainly from velocities within the nalu-wind domain and not ones that are actually being solved in AMR-Wind (so the CFL there is irrelevant to stability). I ran with half the timestep size (0.0025) and the CFL was fine, but the crash occurred at the same physical time (around 1 s), so that suggests it's not a timestep-based problem. My best guess is that it's boundary condition issues on the part of nalu-wind because there are unphysical things happening at the wall. I may try to set up a nalu-nalu version of this and see if the same thing happens.

Finally, I think this whole debacle doesn't rest on this particular part of the code, so I think the PR should go through. I will do a comparison on a moving mesh case to have that recorded here first.

mbkuhn:
Flat surface case with rotating Nalu-Wind mesh after 500 steps.
With implementation from PR #1328 (now in master).

image

With this PR:

image

Spurious currents (not shown) are about the same, but the interface is much more even and sharper with the new approach.

mbkuhn:
Looking at volume fraction in the nalu-wind domain:
Implementation in master:
image

Implementation in this PR:
image

So, even though it didn't fix what I thought it needed to, this seems to be a clearly better way of doing things.

mbkuhn (approved):

PR #1350, opened 2025-01-27
split out compile-time known information for master elements

rcknaus:
Splits out compile-time known information out of the master element, so you can use the isoparametric information directly in gpu kernels if you want.

This was done before in a different and I had a good experience with it. Having the compile-time known information available at compile time simplifies a lot of the data wrangling with the master elements and speeds up the code, at the expense of making the master element-style computations a little more cumbersome. 

Implemented for volume-rank SCS and SCV master elements but not face-rank ones. I just want volume for the GCL, so that's the origin of that. I can finish up the face stuff in a bit or as part of this if it's wanted.

### DIFF
--- a/include/master_element/CompileTimeElements.h
+++ b/include/master_element/CompileTimeElements.h
@@ -1,0 +1,184 @@
+// Copyright 2017 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS), National Renewable Energy Laboratory, University of Texas Austin,
+// Northwest Research Associates. Under the terms of Contract DE-NA0003525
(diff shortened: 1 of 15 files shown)


marchdf:
Nice work! I like the idea of doing them all in one go and keeping things consistent if that's not too much to ask.

jrood-nrel:
Let us know if this is done or not. We can probably merge it.

rcknaus:
> Let us know if this is done or not. We can probably merge it.

Yeah, this is done. Can't merge it myself.

jrood-nrel (approved):

PR #1353, opened 2025-02-07
Update elastodyn input file for actuator reg test

marchdf:


marchdf:
@jrood-nrel I am a bit confused as to why the dashboard isn't failing... I wasn't able to run this test without these modifications.

jrood-nrel:
Are the same versions of openfast being used?

marchdf:
Hard to know. The dashboard doesn't say: https://my.cdash.org/tests/251988223. It uses `@develop` but I think we should be using `@4.0` now: https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/nalu-wind/package.py#L65. Not sure why that's not getting picked up right... 

But my run is clearly stating the version of OF to be 4.0: `OpenFAST-v4.0.0`

PR #1359, opened 2025-02-26
Clean out some sync behavior in nodal grad

rcknaus:
Just need to set on device that grad phi is zero before summing into the field


wjhorne (approved):
Looks good! Are there are any other spots that come to mind where we are erroneously filling out values on host like this?

marchdf (approved):

sseraj (approved):
I finally got around to profiling this on Kestrel GPUs. For the hackathon case on 2 GPUs, the total time for `sierra::nalu::NodalGradAlgDriver<stk::mesh::Field<double> >::pre_work()` calls went from 2.1% to 0.0% of the overall run time (about 90 times speedup).

PR #1366, opened 2025-03-28
Separate time step calc from pre advance, add functions to access/set dt

mbkuhn:


wjhorne:
@mbkuhn I am a bit slammed for today, but I will get to reviewing this soon

wjhorne:
Could we add a test that exercises this capability? Otherwise It looks good!

mbkuhn:
good idea! A unit test for these functions, right?

mbkuhn:
I do have a reg test that will use these, but it's at the exawind-driver level.

wjhorne (approved):

PR #1370, opened 2025-04-02
Weaken penalty bc for VOF

wjhorne:


wjhorne:
@mbkuhn

marchdf (approved):

mbkuhn:
For the record, this PR is addressing the fact that the current VOF BC makes the VOF equation struggle to converge in real cases. This PR weakens the condition currently in master while keeping it effective. For comparison, this is a screenshot using the old BC:
old_BC_oc5_coarse_mesh_hos_nwt

the new BC, currently in master, often hits the full no. of iterations for this case. with this PR, the same case looks like
weakened_new_BC_oc5_coarse_mesh_hos_nwt

so it does a much better job getting the VOF values correct at the solid interface.

mbkuhn (approved):

PR #1380, opened 2025-07-21
Initialize scratch view in unit test

tasmith4:
This view was used uninitialized at line 50.


overfelt:
Yes, I think better checking of `simdrhs` would have exposed this error long ago.  Maybe add another element to the `result` vector and check that `simdrhs` was correctly set by the `testKernel.execute` call.  I see only a few indexes are checked, maybe index 0 and index 1 would be enough since the values from index 1 until the end of the array should now be all zero now that you have fixed the array initialization.

overfelt (approved):

PR #1381, opened 2025-07-25
Fix unit test kernel that accesses uninitialized memory

tasmith4:
Any kernel that tried to accumulate values into the shared memory lhs/rhs views provided by run_algorithm would be working with uninitialized data.

EDIT 9/15/2025: Switched focus of this PR to changing the accumulation to assignment, since it doesn't make sense to use a lambda function to do a reduction inside a parallel_for, especially using scratch memory to accumulate values. Also taking out the checks of `smdata.simdrhs` that were originally part of this PR, since it doesn't make sense to check scratch memory values during a parallel_for (we should be checking things that are set after the parallel_for concludes).

### DIFF
--- a/unit_tests/UnitTestScratchViews.C
+++ b/unit_tests/UnitTestScratchViews.C
@@ -47,7 +47,7 @@ public:
     auto& v_vel = scratchViews.get_scratch_view_2D(velocityOrdinal);
     auto& v_pres = scratchViews.get_scratch_view_1D(pressureOrdinal);
 
-    rhs(0) += v_vel(0, 0) + v_pres(0);
+    rhs(0) = v_vel(0, 0) + v_pres(0);
   }
 
 private:


include/AssembleElemSolverAlgorithm.h
```diff
@@ -96,6 +96,16 @@ class AssembleElemSolverAlgorithm : public SolverAlgorithm
 
         SharedMemData<DeviceTeamHandleType, DeviceShmem> smdata(
           team, nDim, dataNeededNGP, nodesPerEntity, rhsSize);
+        for (int i = 0; i < rhsSize; i++) {
```
marchdf, line 99: Would it be better to use something like `set_vals(smdata.rhs, 0.0);` ? We do that in `AssembleEdgeSolverAlgorithm.h`

tasmith4: Good point, will fix

marchdf (approved):

PR #1385, opened 2025-08-16
Super Ellipse Body

gantech:
Added unit test, regression test and documentation!

Creates a capability to stop the flow inside a super ellipse defined around a point in space $\vec{X}_{loc} = (x_{loc},y_{loc},z_{loc})$ with dimensions $X_c = (x_c, y_c, z_c)$ and orientation with respect to the inertial reference frame defined by the transformation matrix $\mathbf{R}$. The transformation matrix $\mathbf{R}$ will be prescribed using a Wiener-Milenkovic parameter instead of the full matrix. The point in the coordinate system aligned with the super ellipse will be

$$
(\vec{X} - \vec{X}_{loc})' = \mathbf{R} \cdot (\vec{X} - \vec{X}_{loc}).
$$

A given point in space $\vec{X}$ will be inside the super ellipse if 

$$
\left (\frac{(x-x_{loc})'}{x_c} \right )^6 + \left (\frac{(y-y_{loc})'}{y_c} \right)^6 + \left (\frac{(z-z_{loc})'}{z_c} \right)^6 - 1 < 0.
$$

For all mesh nodes inside the super ellipse, a forcing term will be added to the momentum equation proportional to the negative of flow velocity scaled by the time step $dt$ to bring the velocity to zero as 

$$
f^{seb}_i = - 10.0 \rho \frac{u_i}{dt} \; \Delta V.
$$

Files Modified
==============

1. LowMachEquationSystem.C - MomentumEquationSystem - register_interior_algorithm

   Create an option to initiate a NodeKernel and an object of the SuperEllipseBodySrc class if the appropriate option is selected in the input deck.

   The object of SuperEllipseBodySrc is owned by MomentumEquationSystem class as a unique pointer and a reference of this is passed to the NodeKernel class so that the object deletion is safe.

   We need a node kernel as well as a separate SuperEllipseBodySrc class because the node kernel execute function is per mesh node and we want another class to be able to read a file containing the position and orientation of the super ellipse body every time step.

```
        } else if ((srcName == "coriolis") || (srcName == "EarthCoriolis")) {
          nodeAlg.add_kernel<MomentumCoriolisNodeKernel>(
            realm_.bulk_data(), *realm_.solutionOptions_);
        } else if ((srcName == "superellipsebody") || (srcName == "SuperEllipseBody")) {
          seb_ = std::make_unique<SuperEllipseBodySrc>(SOMETHING); //TODO: Get the yaml node here
          nodeAlg.add_kernel<MomentumSuperEllipseBodyNodeKernel>(
            realm_.bulk_data(), *realm_.solutionOptions_);
        } 
```

2. LowMachEquationSystem.h - MomentumEquationSystem

   Declare the object of `SuperEllipseBodySrc` class as a unique pointer. 

```
private:
  ScalarFieldType* Udiag_;
  std::unique_ptr<SuperEllipseBodySrc> seb_;
```

3. LowMachEquationSystem.C - MomentumEquation - assemble_and_solve

   Call execute function of SuperEllipseBodySrc - needs to read and update position and orientation from a file

4. SuperEllipseBodySrc - header file and implementation

   At initialization - reads yaml input file to figure which text file contains information on position and orientation. Also reads information on size of super ellipse.

   At every time step, read and update position and orientation.

5. SuperEllipseBodySrcNodeKernel - header file and implementation.

    Implements the body force term and the Jacobian term for the momentum equation.

    My understanding is that the setup call for each node kernel is called every time step from AssembleNGPNodeSolverAlgorithm.C::execute(). So I can stuff whatever updates I need to update the location and orientation of the Super Ellipse from the seb_ object.

```
void
AssembleNGPNodeSolverAlgorithm::execute()
{
  using ShmemDataType = SharedMemData_Node<DeviceTeamHandleType, DeviceShmem>;

  const size_t numKernels = nodeKernels_.size();
  if (numKernels < 1)
    return;

  for (auto& kern : nodeKernels_)
    kern->setup(realm_);
```        
    
    Looks like vs::Vector objects will be copied over into the kernels!


include/node_kernels/MomentumSuperEllipseBodyNodeKernel.h
```diff
@@ -0,0 +1,78 @@
...
+  unsigned dualNodalVolumeID_{stk::mesh::InvalidOrdinal};
+  unsigned densityNp1ID_{stk::mesh::InvalidOrdinal};
+  unsigned velocityNp1ID_{stk::mesh::InvalidOrdinal};
+  unsigned coordinatesID_{stk::mesh::InvalidOrdinal};
+
+  // Reference to the SuperEllipseBodySrc object
+  const SuperEllipseBodySrc& seb_;
+
+  // Location of the Super Ellipse center
+  vs::Vector seb_loc_;
```
marchdf, line 65: This is a silly style thing but we seem to use snakeCase_ for member vars.

src/SolutionOptions.C
```diff
@@ -417,6 +417,10 @@ SolutionOptions::load(const YAML::Node& y_node)
             referenceVelocity_);
           get_if_present(
             y_user_constants, "rans_below_ks", RANSBelowKs_, RANSBelowKs_);
+          get_if_present(
+            y_user_constants, "super_ellipse_body_file", superEllipseBodyFile_,
+            superEllipseBodyFile_);
+          std::cerr << "super ellipse body file = " << superEllipseBodyFile_ << std::endl;
```
marchdf, line 423: why send to `cerr`?

PR #1411, opened 2026-04-13
Prescribed motion

wjhorne:
## Summary

This PR adds a prescribed motion mesh_motion type where a user can specify angular and translational positions or velocities and it will move a mesh accordingly. 

## Pull request type

- [ ] Bugfix
- [x ] Feature
- [ ] Build related changes
- [ ] Other (please describe):

## Checklist

The following is included:

- [ ] new unit-test(s)
- [ x] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x ] on CPU 
- the regression tests
  - [ ] on GPU 
  - [x ] on CPU 

## Additional background

I still need to re-bless the gold on the regression test. I am having some issues with my nalu-wind install on my development box with regards to new tests that I need to sort out. I did actually run it by hand and it looked reasonable.

Issue Number:


wjhorne:
I believe everything is now in order here. I was able to re-bless the relevant test without issue

wjhorne:
Sorry for pinging you @mbkuhn on the review. I am not sure who I could contact to do so and you were on the top of the list. I'll poke Lawrence for other options.

mbkuhn:
> Sorry for pinging you @mbkuhn on the review. I am not sure who I could contact to do so and you were on the top of the list. I'll poke Lawrence for other options.

All good! I'm probably the right person. I did add a different class here recently. Nalu-Wind developers are pretty bare these days, we just have to make do.

mbkuhn:
I'm seeing a few instances of "mismatched" indices, but that's just a consequence of relevant_motion having a time recorded at the 0th index, while the information is being put into arrays that don't have that offset. Is that right?

Also, could you explain the logic with min_index?

wjhorne:
The indices are certainly confusing since I included time as 0 and have things stored in long rows in a matrix which then gets pulled into 4x4 matrices and 3 vectors for the mesh_motion API. I had to go this route since this is all on device and nice readable classes aren't generally simple to work with. I'll double check that they are all correct, but my initial tests looked like things are good on the indexing. 

min_index is poorly named, but used to find the time data that is the closest to the current simulation time. It ultimately finds two time states when possible so that it can interpolate states at the current simulation time. The prescribed motion currently assumes constant velocity between time states (straight line between displacements). It calculates this velocity using displacements if they are given, it otherwise defers to specified velocities.

mbkuhn:
Thanks for the explanation! Makes sense. Could you add a handful of comments to annotate these less obvious parts of the code?

mbkuhn (approved):

wjhorne:
I added a few comments describing how the indices work. I'll merge in once everything clears CI again.

PR #1420, opened 2026-05-23
Add GCL correction for general topologies

rcknaus:
Adds an optional continuity residual correction to kynema-ugf. This will add a non-kinetic energy conserving residual correction if the flow, for whatever reason, has a discrete mass change including a GCL error. For the GCL conforming topologies  this has no impact in constant density flow. In non-constant density flow, this will always have an impact and will ensure discrete mass conservation on the first nonlinear iteration, where the predictor is insufficient for maintaining discrete conservation.

This capability will throw an error in a few, likely uncommon, use cases, such as nonconforming meshes and inflow from the FFT-based ABL top boundary condition capability. If those cases are wanted, let me know.

Also adds a hybridMeshDeformation test demonstrating free-stream preservation on a fully hybrid mesh containing all 3D topologies. The mesh itself will need to be added to the mesh repository in a separate PR for that sub-repo. https://github.com/kynema/kynema-ugf-meshes/pull/15


rcknaus:
Test failure for CPU-Trilinos build seems unrelated and pre-existing given https://github.com/kynema/kynema-ugf/pull/1422 . Interestingly it's failing at a different spot in each, so probably need an asan build

gantech:
Hey @rcknaus . Thanks for implementing this. Here are my comments. I need some more help to finish reviewing this.

1. This algorithm is different from the one we had earlier for hex-only meshes. Could you give a single page description as a comment or a pdf file of this algorithm and how it works to satisfy gcl? 

2. We already had a test MeshMotionInterior. Does this still work with this algorithm to the same tolerance in continuity?

3. Have we tried any of the exawind benchmarks with FSI with this algorithm? Specifically this one? https://github.com/kynema/kynema-benchmarks/tree/main/exawind/Pazy_Wing

rcknaus:
Hey Ganesh, in lieu of a document, basically the idea is that you subtract out the continuity residual from flow when the `include_continuity_residual: yes` option is on. This subtracts out (d_t rho + div mdot) q where q is the dof. That term is exactly the discrete free-stream preservation error (say q is constant) and so ensures free-stream preservation that way. Generally it'll impact any case where there is a discrete loss of mass conservation for whatever reason (so variable density, non-divergence free flows too given we don't fully ensure mass conservation there until nonlinear convergence).

For the new test vs MeshMotionInterior, the new term has no impact whether it's on or not for the MeshMotionInterior test (the continuity residual is ~1e-15 for that case).  Anyway, for the new test, running the case with a static mesh has a small free-stream preservation error that I think is due to a lack of volume/area conservation in the pyramid formulation itself. I didn't really track it down--sorry. The mesh motion increases the error ~100 times the static case, but there the mesh almost inverts there and is dramatically better than not having it the continuity residual on.

For testing, I haven't run the Pazy wing test. The expectation there is that this has no impact whether it's switched on or off for that case.  There was a less general version of this I ran on a hybrid case Nate worked on a while ago and kept the case from diverging. But more robust testing on more user cases is certainly warranted. 

There are definitely some things that could be wrapped up more here, but this should allow you to run deforming, hex-pyramid-tet meshes.

EDIT: also doesn't account for any continuity sources. We don't have any in kynema-ugf but if you ever add them, the term will also need to get adjusted.

gantech:
Hey Robert. Thanks. 

1. The new approach to satisfying GCL is still unclear to me.

2. I don't think the regression test I mentioned earlier will work without the GCL treatment we had earlier. 

I believe we need to turn off the existing mechanism to satisfy GCL to test the new implementation. Is this correct?

If this is correct, have we turned off the existing mechanism to satisfy GCL as a part of this pull request? I believe the existing mechanism turns on automatically if mesh deformation is on. https://github.com/kynema/kynema-ugf/blob/main/src/edge_kernels/ContinuityEdgeSolverAlg.C#L92 

3. Is the new approach proven through implementation in literature? I believe the previous approach was copied verbatim from Mavriplis et al. and used in a lot of work. If the new approach is not from the literature, then I think we must at least run the exawind benchmarks to prove the new approach on real cases before using it.

rcknaus:
@gantech updated with the small changes I mentioned in our meeting and a documentation blurb on the residual correction.  For the regression test, you'll need to merge https://github.com/kynema/kynema-ugf-meshes/pull/15 first and potentially will need to have @jrood-nrel rebless with gold file with nrel systems.

jrood-nrel (approved):
I'm fine with this. The CI failures are unrelated. Pragma once isn't consistent with the rest of the code, but I don't have a preference for it or not.

ndevelder:
FYI all, I got the NREL 5MW rotor with FSI (no tower or ABL) case running with this PR. It has run around 300 time steps and seems stable.  

Here's what the mesh looks like:
Screenshot 2026-06-23 at 11 49 01 AM

PR #1423, opened 2026-05-27
Bug fix in the IDDES 

BumseokLee:
## Summary

This pull request includes 1) a bug fix for the Pk limiter in IDDES, and 2) the implementation of the simplified IDDES model, which was reported and requested by Michael Stoellinger at the University of Wyoming 

Regarding the bug fix for the Pk limiter, the existing unit tests were unaffected because the limiter is only activated under extreme conditions. I also tested the NACA0021 airfoil cases, which are part of the Kynema benchmark cases, and observed only minor changes in the results. The overall trends remain the same.

The simplified IDDES model has also been implemented. The only difference between the original IDDES ( `turbulence_model: sst_iddes` ) and the simplified IDDES ( `turbulence_model: sst_iddes_simple` ) is that the original IDDES uses a non-zero elevation function, f_e ≠ 0, whereas the simplified IDDES sets  f_e = 0. Therefore, I reused the TKE node kernel implementation from IDDES and modified the simplified IDDES node kernel such that f_e = 0. I also created a new unit test for the simplified IDDES model; however, it is effectively identical to the existing IDDES unit test.

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] Other (please describe):

## Checklist

The following is included:

- [x] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [x] on CPU 
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

## Additional background

This pull request corresponds to the issue: https://github.com/kynema/kynema-ugf/issues/1397


jrood-nrel:
I was on the wrong PR adding reviewers, so copilot is probably going to review this even though I tried to cancel it.

mbkuhn (approved):

PR #1487, opened 2026-09-04
Add Beam mapping methods to KynemaFMBBase

gantech:
- Introduced `CalcLoadsAssembled` class for calculating assembled loads in the FSI context.
- Enhanced `KynemaFMBBase` with methods for mapping beam displacements and loads.
- Added unit tests for `KynemaFMBBase` to validate point and beam load mappings.

## Summary

FSI mapping for beams

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Other (please describe):

## Checklist

The following is included:

- [X] new unit-test(s)
- [ ] new regression test(s)

- the unit tests
  - [ ] on GPU 
  - [X] on CPU Mac + clang
- the regression tests
  - [ ] on GPU 
  - [ ] on CPU 

No change made to existing capabilities. So no need to run regression tests. Even for unit tests, only the new ones matter because again no change made to existing capabilities. 

## Additional background

FSI mapping for beams

Issue Number:

### DIFF
--- a/include/aero/fmb/KynemaFMBBeamUtils.h
+++ b/include/aero/fmb/KynemaFMBBeamUtils.h
@@ -1,3 +1,5 @@
+ #include <Kokkos_Core.hpp>
+
 namespace sierra {
 namespace kynema_ugf {
 
@@ -464,6 +466,11 @@ FindClosestPointOnBlade(
       const double fprime_new = BladePointFPrimeAtXi(
         xi_new, query_point, node_xi, node_positions, num_nodes, bary_weights,
         scratch_weights, scratch_dweights);
+      if (std::abs(fprime_new) < 1.e-14) {
+        a = xi_new;
+        b = xi_new;
+        break;
+      }
       // Narrow the bracket.
       if (fprime_new * fprime_a < 0.0) {
         b = xi_new;
--- a/include/aero/fmb/KynemaFMBBeamUtils.h
+++ b/include/aero/fmb/KynemaFMBBeamUtils.h
@@ -1,0 +1,485 @@
+namespace sierra {
+namespace kynema_ugf {
+
+
+
+    
+/**
+ * @brief Rotates provided vector by provided *unit* quaternion and returns the result
+ */
+template <typename Quaternion, typename View1, typename View2>
+KOKKOS_INLINE_FUNCTION void RotateVectorByQuaternion(
+    const Quaternion& q, const View1& v, View2& v_rot
+) {
+    v_rot[0] = (q[0] * q[0] + q[1] * q[1] - q[2] * q[2] - q[3] * q[3]) * v[0] +
+               2. * (q[1] * q[2] - q[0] * q[3]) * v[1] + 2. * (q[1] * q[3] + q[0] * q[2])
+ * v[2];
+    v_rot[1] = 2. * (q[1] * q[2] + q[0] * q[3]) * v[0] +
+               (q[0] * q[0] - q[1] * q[1] + q[2] * q[2] - q[3] * q[3]) * v[1] +
+               2. * (q[2] * q[3] - q[0] * q[1]) * v[2];
+    v_rot[2] = 2. * (q[1] * q[3] - q[0] * q[2]) * v[0] + 2. * (q[2] * q[3] + q[0] * q[1])
+ * v[1] +
+               (q[0] * q[0] - q[1] * q[1] - q[2] * q[2] + q[3] * q[3]) * v[2];
+}
+
+/**
+ * @brief Rotates provided vector by inverse of provided *unit* quaternion and returns the result
+ */
+template <typename Quaternion, typename View1, typename View2>
+KOKKOS_INLINE_FUNCTION void RotateVectorByQuaternionInv(
+    const Quaternion& q, const View1& v, View2& v_rot
+) {
+    v_rot[0] = (q[0] * q[0] + q[1] * q[1] - q[2] * q[2] - q[3] * q[3]) * v[0] +
+               2. * (q[1] * q[2] + q[0] * q[3]) * v[1] + 2. * (q[1] * q[3] - q[0] * q[2])
+ * v[2];
+    v_rot[1] = 2. * (q[1] * q[2] - q[0] * q[3]) * v[0] +
+               (q[0] * q[0] - q[1] * q[1] + q[2] * q[2] - q[3] * q[3]) * v[1] +
+               2. * (q[2] * q[3] + q[0] * q[1]) * v[2];
+    v_rot[2] = 2. * (q[1] * q[3] + q[0] * q[2]) * v[0] + 2. * (q[2] * q[3] - q[0] * q[1])
+ * v[1] +
+               (q[0] * q[0] - q[1] * q[1] - q[2] * q[2] + q[3] * q[3]) * v[2];
+}
+
+/**
+ * @brief Computes 3D cross product of two vectors and stores the result.
+ */
+template <typename View1, typename View2, typename View3>
+KOKKOS_INLINE_FUNCTION void CrossProduct3(
+    const View1& a, const View2& b, View3& c
+) {
+    c[0] = a[1] * b[2] - a[2] * b[1];
+    c[1] = a[2] * b[0] - a[0] * b[2];
+    c[2] = a[0] * b[1] - a[1] * b[0];
+}
+
+/**
+ * @brief Computes barycentric weights for interpolation nodes.
+ *
+ * Compute once for a given node set and reuse for many evaluations.
+ * This is O(n^2) and intended for setup, not the hot loop.
+ *
+ * @param xs Interpolation nodes
+ * @param n Number of interpolation nodes
+ * @param bary_weights Output barycentric weights (size n)
+ */
+KOKKOS_INLINE_FUNCTION
+inline void ComputeBarycentricWeights(
+    const double* KOKKOS_RESTRICT xs, int n, double* KOKKOS_RESTRICT bary_weights
+) {
+    for (int j = 0; j < n; ++j) {
+        double w = 1.0;
+        for (int m = 0; m < n; ++m) {
+            if (j != m) {
+                w /= (xs[j] - xs[m]);
+            }
+        }
+        bary_weights[j] = w;
+    }
+}
+
+/**
+ * @brief Computes Lagrange interpolation basis weights at x.
+ *
+ * Barycentric form is O(n) per evaluation and avoids dynamic allocations,
+ * making it suitable for use inside Kokkos kernels.
+ *
+ * @param x Evaluation point
+ * @param xs Interpolation nodes
+ * @param bary_weights Precomputed barycentric node weights
+ * @param n Number of interpolation nodes
+ * @param weights Output interpolation basis weights (size n)
+ */
+KOKKOS_INLINE_FUNCTION
+inline void LagrangePolynomialInterpWeights(
+    double x,
+    const double* KOKKOS_RESTRICT xs,
+    const double* KOKKOS_RESTRICT bary_weights,
+    int n,
+    double* KOKKOS_RESTRICT weights
+) {
+    constexpr double tol = 1.e-14;
+
+    // If x coincides with a node, return exact Kronecker-delta basis values.
+    for (int j = 0; j < n; ++j) {
+        if (std::abs(x - xs[j]) <= tol) {
+            for (int k = 0; k < n; ++k) {
+                weights[k] = 0.0;
+            }
+            weights[j] = 1.0;
+            return;
+        }
+    }
+
+    double denom = 0.0;
+    for (int j = 0; j < n; ++j) {
+        const double tmp = bary_weights[j] / (x - xs[j]);
+        weights[j] = tmp;
+        denom += tmp;
+    }
+
+    const double inv_denom = 1.0 / denom;
+    for (int j = 0; j < n; ++j) {
+        weights[j] *= inv_denom;
+    }
+}
+
+/**
+ * @brief Computes Lagrange basis values and their xi-derivatives in one pass.
+ *
+ * At interior points uses the barycentric derivative formula:
+ *   dl_j/dx = l_j * (S2/S - 1/(x - x_j))
+ * where S = sum_k w_k/(x-x_k) and S2 = sum_k w_k/(x-x_k)^2.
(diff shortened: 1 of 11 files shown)


unit_tests/aero/UnitTestKynemaFMBBase.C
```diff
@@ -0,0 +1,303 @@
...
+
+  for (const auto* bucket : bulk_->get_buckets(
+         stk::topology::NODE_RANK, meta_->locally_owned_part())) {
+    for (const auto node : *bucket) {
+      const double* model = stk::mesh::field_data(*coordinates, node);
+      const double* displacement = stk::mesh::field_data(*meshDisp_, node);
+      const double* velocity = stk::mesh::field_data(*meshVelocity_, node);
+      const double* current = stk::mesh::field_data(*currentCoords_, node);
+      EXPECT_NEAR(1.0 - model[0] - model[1], displacement[0], 1.e-12);
+      EXPECT_NEAR(2.0 + model[0] - model[1], displacement[1], 1.e-12);
```
PorterJustin, line 217: What is the logic? I am confused about why the x displacement is the difference between the x and y positions.

unit_tests/aero/UnitTestKynemaFMBBase.C
```diff
@@ -0,0 +1,303 @@
...
+      std::make_shared<sierra::kynema_ugf::CalcLoads>(point.forcing_surfaces);
+    point.calc_loads->setup(bulk_);
+    return point;
+  }
+
+  BeamBody make_beam_body()
+  {
+    BeamBody beam;
+    beam.bulk = bulk_;
+    beam.n_nodes = 2;
```
PorterJustin, line 133: The test appears to only cover linear elements not higher order elements.

include/aero/fmb/KynemaFMBBeamUtils.h
```diff
@@ -0,0 +1,501 @@
...
+  const double* KOKKOS_RESTRICT node_positions,
+  int num_nodes,
+  const double* KOKKOS_RESTRICT bary_weights,
+  double* KOKKOS_RESTRICT scratch_weights,
+  double* KOKKOS_RESTRICT scratch_dweights,
+  double& out_xi,
+  double* KOKKOS_RESTRICT out_position,
+  double& out_dist2)
+{
+  constexpr int kCoarseSamples = 10;
```
PorterJustin: I might be able to come up with an example that would break this by using higher order polynomials. Would it make sense to set this as `num_nodes` or `num_nodes+1`? Or something similar? 

I haven't thought through what exactly it would need to be to prevent breaking this function. On the other hand, any example that I can think might break this may break the ability of a plane section to remain plane. Also, since this is just at initialization, it seems pretty unlikely that the crazy function I could come up with would happen (versus that function may happen during the simulation but that doesn't matter here).

include/aero/fmb/KynemaFMBBeamUtils.h
```diff
@@ -0,0 +1,501 @@
...
+/**
+ * @brief Computes shape function matrix ϕg relating points ξb to ξg.
+ *
+ * Layout of shape_functions is row-major with [output][input], flattened as
+ * shape_functions[output_idx * num_input_points + input_idx].
+ *
+ * @param input_points Input points ξb in [-1, 1]
+ * @param num_input_points Number of input points
+ * @param output_points Output points ξg in [-1, 1]
+ * @param output_bary_weights Precomputed barycentric weights for output_points
```
PorterJustin, line 208: I'm not sure about this function and what you intend to use it for. When I have made matrices like this, it is to interpolate from beam nodes (inputs) to many more points for plotting (outputs). However, here you use the basis functions of the number of output points, which would be incorrect for my previous use cases. 

The only unit test on this function has inputs and outputs identical, so that is not informative as to the intention for the use of this function.

unit_tests/aero/UnitTestKynemaFMBBeamUtils.C
```diff
@@ -0,0 +1,187 @@
...
+  std::array<double, 2> barycentricWeights{};
+  std::array<double, 2> scratchWeights{};
+  std::array<double, 2> scratchDerivatives{};
+  std::array<double, 3> closestPosition{};
+  double closestXi = 0.0;
+  double distanceSquared = 0.0;
+
+  ComputeBarycentricWeights(nodes.data(), 2, barycentricWeights.data());
+  const std::array<double, 3> interiorQuery = {0.25, 2.0, -1.0};
+  FindClosestPointOnBlade(
```
PorterJustin, line 169: Do you want to test this using something higher order than a linear element?

unit_tests/aero/UnitTestKynemaFMBBase.C
```diff
@@ -0,0 +1,303 @@
...
+TEST_F(KynemaFMBBaseTest, MapsPointLoadsWithCalcLoads)
+{
+  auto point = make_point_mass();
+  point.p_data.pos = {0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0};
+
+  fmb_.map_loads_point(point);
+
+  EXPECT_NEAR(48.0, point.p_data.loads[0], 1.e-12);
+  EXPECT_NEAR(0.0, point.p_data.loads[1], 1.e-12);
+  EXPECT_NEAR(0.0, point.p_data.loads[2], 1.e-12);
```
PorterJustin, line 287: Should this check moments as well? Not looking too closely at the point tests since they don't appear to be for new functionality in this PR.

src/aero/fmb/KynemaFMBBase.C
```diff
@@ -140,5 +147,264 @@ KynemaFMBBase::map_loads_point(PointMass& point)
...
+      InterpolateFieldAtPoint(
+        xi, node_xi.data(), displacements.data(), n_nodes, 7,
+        bary_weights.data(), scratch_weights, disp);
+      InterpolateFieldAtPoint(
+        xi, node_xi.data(), velocities.data(), n_nodes, 6, bary_weights.data(),
+        scratch_weights, vel);
+
+      double rel_pos_g[3] = {qp[0] - pos[0], qp[1] - pos[1], qp[2] - pos[2]};
+      double rel_pos_l[3] = {0.0, 0.0, 0.0};
+      RotateVectorByQuaternionInv(pos + 3, rel_pos_g, rel_pos_l);
```
PorterJustin: I'm not completely sure about this logic.

src/aero/fmb/KynemaFMBBase.C
```diff
@@ -140,5 +147,264 @@ KynemaFMBBase::map_loads_point(PointMass& point)
...
+
+      double tmp_disp[3] = {0.0, 0.0, 0.0};
+      double rot_disp[3] = {0.0, 0.0, 0.0};
+      RotateVectorByQuaternion(disp + 3, rel_pos_l, tmp_disp);
+      RotateVectorByQuaternion(pos + 3, tmp_disp, rot_disp);
```
PorterJustin, line 289: I think we had briefly discussed this a while ago. Why can't you just do this?
```
RotateVectorByQuaternion(disp + 3, rel_pos_g, rot_disp);
```

Looking at FMB: [this call](https://github.com/kynema/kynema-fmb/blob/cb0c7a4de085a43543aab04c641fe19c0a221961/src/state/update_global_position.hpp#L38-L40) appears to combine the Quaternions as `RotateBy(Displacement, RotateBy(InitialPosition, vector))`. So to be consistent here, lines 288 and 289 would be swapped and then the two position rotations would cancel.

src/aero/fmb/KynemaFMBBase.C
```diff
@@ -140,5 +147,264 @@ KynemaFMBBase::map_loads_point(PointMass& point)
...
+      double xi = beam_xi.get(mi, 0);
+      const double force[3] = {tforce(mi, 0), tforce(mi, 1), tforce(mi, 2)};
+      double moment[3] = {0.0, 0.0, 0.0};
+
+      // Get the closest position on the beam to the query point and compute the
+      // relative position vector Relative position vector is expected to not
+      // change with beam deformation
+      InterpolateFieldAtPoint(
+        xi, node_xi.data(), positions.data(), n_nodes, 7, bary_weights.data(),
+        scratch_weights, closest_position);
```
PorterJustin, line 375: This assumption sounds incorrect. Say the beam rotates, then the direction of the vector changes significantly. FMB is all in global coordinates, so forces should be likewise. I think you should also interpolate displacements and include those. Since this is just the 3 translations, that should be easy enough.

unit_tests/aero/UnitTestKynemaFMBBase.C
```diff
@@ -0,0 +1,303 @@
...
+{
+  auto beam = make_beam_body();
+  fmb_.compute_mapping_beam(beam);
+
+  fmb_.map_loads_beam(beam);
+
+  double totalForceX = 0.0;
+  for (std::size_t node = 0; node < beam.n_nodes; ++node)
+    totalForceX += beam.beam_data_host.loads(node, 0);
+  EXPECT_NEAR(16.0, totalForceX, 1.e-12);
```
PorterJustin, line 300: I think the test should also check moments as well given that I am not sure about the lever arm calculation being used.

PorterJustin, line 300: The point load is 48 and the beam load is 16. Claude tracks this to a difference of loads at integration points versus at nodes. Should the integration scheme be consistent between the two formulations (points and beams)? And thus the total forces be the same in both cases?

PorterJustin (commented):
I have some concerns about the theory in KynemaFMBBase for beams detailed in comments. The KynemaFMBBeamUtils looks mostly ok, but a few questions there as well.

unit_tests/aero/UnitTestKynemaFMBBase.C
gantech: Addressed in 0f8085fd2e062fe3b630234ad7f0f2bb7c0639e1. Now covers 10th order element.

unit_tests/aero/UnitTestKynemaFMBBeamUtils.C
gantech: Addressed in 0f8085fd2e062fe3b630234ad7f0f2bb7c0639e1. Now covers 10th order element.

unit_tests/aero/UnitTestKynemaFMBBase.C
gantech: Addressed in 0f8085fd2e062fe3b630234ad7f0f2bb7c0639e1.

include/aero/fmb/KynemaFMBBeamUtils.h
gantech: This function was never used anywhere and was superseded by the one that computes both the shape functions and derivatives. So I removed this.

unit_tests/aero/UnitTestKynemaFMBBase.C
gantech: A `PointMass` selects the universal part as its moving mesh and uses:

- Center of mass: $(0, 0, 0)$.
- Position: translation $(1, 2, 3)$ and a unit quaternion for a $90$ degree
  rotation about $z$.
- Velocity: translational $(4, 5, 6)$ and angular $(0, 0, 2)$.

For every node at
$(x, y, z)$, it verifies:

$$
\text{mesh displacement} = (1 - x - y,\; 2 + x - y,\; 3)
$$

$$
\text{mesh velocity} = (4 - 2x,\; 5 - 2y,\; 6)
$$

unit_tests/aero/UnitTestKynemaFMBBase.C
gantech: That's a good point. Right now the force integration scheme is definitely not consistent between the point and the beam. The forces are technically computed on the sub-control surfaces of the boundary elements on the boundary side sets. Since all points/sub-control surfaces map to the same center of mass for points, the force assembly directly loops over the elements and sub-control surfaces and calculates the total force and moment about the center of mass. 

I created a slight add on to this for beams in CalcLoadsAssembled.C where I assemble the forces from each element onto its corresponding nodes. This is not conservative in the moment as it's neglected within each CFD boundary element. This is an architectural choice that allows me to reuse the mapping between the CFD mesh nodes and the beam. Otherwise I'd have to recreate a mapping to the center of the sub-control surfaces (3-4 per element) and increase my memory storage cost. What do you recommend here?

That said.. this shouldn't make much of a difference in a real case. Unless the test element is ridiculously big! I'll check this.

gantech: That was a bug! Nice catch! Thanks! Addressed in 8ba2fbba3d98e85069c8f9431a1f7a9f89b1ff82

unit_tests/aero/UnitTestKynemaFMBBase.C
gantech: Addressed in 8ba2fbba3d98e85069c8f9431a1f7a9f89b1ff82

src/aero/fmb/KynemaFMBBase.C
gantech: Agreed. Applied suggested fix in c61e879d083d1a5862aacdf41bd5311b2a2bb87c

src/aero/fmb/KynemaFMBBase.C
gantech: So.. that would change the relative order of application of disp and pos (reference) rotations. My understanding is they're not commutative. So we can do it either the current way or the way you suggest. But both cannot be right. So that still leaves showing which way is the correct way. My understanding is that pos (reference) stores the rotation one needs to apply to the inertial frame to bring it to the local frame. And then the disp stores the extra rotational displacement. Now for a point in the local frame the order of application would be reversed and the only thing constant between the reference configuration and the deformed configuration is that the relative position in the local frame is the same. So I think we would start by applying the rotational displacement to bring it back to the reference configuration and then again the reference orientation to bring it back to the inertial frame. That's why I think this order is correct.

The FMB call is doing it for the nodes on the beam itself for the sake of book keeping right? As in it's just storing a composed quaternion as the absolute position instead of a reference orientation and a displacement on top of it. it's not actually applying that rotation to any point. Probly a relevant comparison would be how you calculate the absolute position of a node on the beam if the hub had a rotational displacement and the beam was rigid?

Let's discuss this one-on-one.

